PR #8, opened 2017-06-29
Task/add post types components

alvarezmelissa87:
Add all post types to `shared-components`
- DashboardImagePost
- DashboardMultipleImagesPost
- DashboardLinkPost
- DashboardTextPost
- DashboardVideoPost


alvarezmelissa87:
Hm, noticing that the `linkTo` doesn't seem to be working anymore in the stories. Going to take a look - perhaps we need to pull it in separately instead of from `@storybook/react`  🤔

PR #9, opened 2017-06-30
Task/add post list

alvarezmelissa87:
- Add `DashboardPostList` to `shared-components`
- Ensure `onShareNowClick` is passed through props for all post types


alvarezmelissa87:
I'm not sure if the travis build will run for this 🤔 I'll merge on Monday after double checking the build still passes.

packages/shared-components/DashboardPostList/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+    case 'video':
+      return (<DashboardVideoPost {...postWithEventHandlers} />);
+    default:
+      return (<DashboardTextPost {...postWithEventHandlers} />);
+  }
+};
+
+/* eslint-enable react/prop-types */
+
+const DashboardPostList = ({
```
hharnisc: What do you think about renaming this component to `Posts`?

packages/shared-components/DashboardPostList/index.jsx
```diff
@@ -0,0 +1,112 @@
...
+  <div>
+    <div style={listHeaderStyle}>
+      <Text
+        color={'black'}
+      >
+        {listHeader}
+      </Text>
+    </div>
+    <List
+      items={posts.map(post =>
```
hharnisc: Was thinking about when we have `Drafts` (for Collab Tool) and `Posts` and how they share some similarities.

What if this component rendered a list of children rather than take the post data and translate it

Would look something like this when using it:

```js
<Posts> // or DashboardPostList
  <Post />
  <Post />
  <Post />
</Posts>
```

The implementation would look similar to what @emily-plummer is doing with Tabs https://github.com/bufferapp/buffer-publish/pull/10/files#diff-00df55b2c7a3bed11459be554ef2a0a5R12

alvarezmelissa87: Really like this idea, @hharnisc! 🙌  Plus it feels great to keep it consistent with lists of things like 'Tabs' 🤔 

I'll go ahead and update it! 😄

packages/shared-components/DashboardPostList/index.jsx
alvarezmelissa87: Sounds good! 👍  With the additional suggested changes, sounds like it's a perfect name for a shared component! 💯 Thanks, @hharnisc!

PR #10, opened 2017-06-30
adds tab and tabs components 🎉 

emily-plummer:
WIP branch for the new Tab & Tabs components! Leaving a few comments in the code :)


packages/shared-components/Tab/index.jsx
```diff
@@ -0,0 +1,48 @@
...
+import PropTypes from 'prop-types';
+import {
+  curiousBlue,
+} from '@bufferapp/components/style/color';
+import {
+  Button,
+} from '@bufferapp/components';
+import HoverableText from '../HoverableText';
+
+const tabStyle = (active) => {
```
emily-plummer: I haven't seen this style anywhere in the codebase - curious if this syntax is alright with everyone? Lets us pass in some parameters to use for the styles!

packages/shared-components/Tabs/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+const tabShape = PropTypes.shape({
+  title: React.PropTypes.string,
+  onClick: PropTypes.func,
+  active: PropTypes.bool,
+});
+
+Tabs.propTypes = {
+  tabs: PropTypes.arrayOf(tabShape).isRequired,
+};
```
emily-plummer: @hharnisc What do you think about passing in a custom shape to arrayOf here? I was getting an error that `arrayOf(PropTypes.obj)` was not allowed so decided to put in something more specific.

packages/shared-components/Tabs/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+const Tabs = ({
+  tabs,
+}) =>
+  <div style={tabsStyle}>
+    {tabs.map(tab => <Tab key={uuid()} onClick={tab.onClick}>{tab.title}</Tab>)}
+  </div>;
+
+const tabShape = PropTypes.shape({
+  title: React.PropTypes.string,
+  onClick: PropTypes.func,
```
emily-plummer: Was thinking that we could pass in the onClick event here from the parent, and the parent would handle showing whatever view/component tab we want to show when clicking on the tab. Does that leave too much work to the parent or would it be preferred to do some of that work in this component? cc @hharnisc

packages/shared-components/Tabs/index.jsx
```diff
@@ -0,0 +1,34 @@
...
+  active: PropTypes.bool,
```
emily-plummer: Same as the above with the `active` parameter, wasn't sure if we should handle the active state here to leave that up to the parent!

packages/shared-components/Tabs/index.jsx
hharnisc: Passing the shape is great here :smile: I imagine someone new to the code base looking at this and then knowing exactly what they need to pass into `Tabs`

packages/shared-components/Tabs/index.jsx
hharnisc: Thinking about the API/Props for Tabs (for this example)

```js
({
  tabs, // array of tabs
  activeTab, // the currently active tab id (or undefined if nothing active)
  onTabClick, // returns tab id that was clicked
}) => {/* render tabs */};
```

hharnisc: Then we could store the currently selected tab in Redux and let the reducer keep track of which tab was selected :smile:

packages/shared-components/Tab/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+import PropTypes from 'prop-types';
+import {
+  curiousBlue,
+} from '@bufferapp/components/style/color';
+import {
+  Button,
+} from '@bufferapp/components';
+import HoverableText from '../HoverableText';
+
+const tabStyle = (active) => { //eslint-disable-line
```
hharnisc: Absolutely love you're thinking about styling this way with props :heart: 

We've got a helper function in Buffer Components that captures this workflow here: 

https://github.com/bufferapp/buffer-publish/blob/master/packages/components/lib/utils.js#L17-L29

An example of how to use it:

https://github.com/bufferapp/buffer-publish/blob/master/packages/components/LinkStateless/index.jsx#L29-L45

hharnisc: Importing that function is a little strange right now -- but I think I'd like to break that out into its own package :smile: 

For now though

```js
import { calculateStyles } from '@bufferapp/components/lib/utils';
```

packages/shared-components/Tabs/story.jsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import {
+  action,
+  storiesOf,
+} from '@storybook/react';
+import { checkA11y } from 'storybook-addon-a11y';
+import Tabs from './index';
+
+const tabs = [
+  { title: 'Queue (1)', onClick: action('tab-click'), active: false },
```
hharnisc: The approach to pass in tabs as data definitely works and has some advantages around keeping tabs looking consistent. I think we might get more usage out of `Tabs` and `Tab` if the API on Tabs was adjusted just a little:

```js
<Tabs
  activeTab={activeTab}
  onTabClick={onTabclick}
>
  <Tab>Queue (1)</Tab>
  <Tab>Sent Posts</Tab>
  <Tab>Settings</Tab>
</Tabs>
```

The difference being to pass each `Tab` as children. Then set `active` and pass `onTabClick` from to `Tabs` to `Tab`

Would look something like this

```
const Tabs = ({ children, activeTabId, onTabClick }) => (
  <div>
    {React.Children.map(children, tab => (
      React.cloneElement(tab, {
        // pass props from `Tabs` to `Tab`
        active: tab.tabId === activeTabId,
        onClick: () => onTabClick({ tabId: tab.tabId }),
      })
    ))}
  </div>
);
```

emily-plummer: Oooh I like this much better, feels great! 💯

emily-plummer: One question! How should we create & pass tab ids? In this example, `tab` is a node right? Is this example just accessing the React elements props when it does `tab.tabId`?

alvarezmelissa87:
Heya @emily-plummer - this should be working now! You might need to do `./dev restart` once you pull 😄

emily-plummer:
You are the best @alvarezmelissa87!!!! 😻 ✨ 🙌

KatieLo:
yay all checks passing ✨ great persistence @emily-plummer!

PR #11, opened 2017-06-30
Updated dev and new production Dockerfiles. Add deployment pipeline

djfarrelly:
### Purpose
Let's run this thing in production! Why not update to Node 8 in the meantime 👍 

### Notes
The `Dockerfile.development` has also been updated when updating things for prod. To rebuild this:
```
./dev rebuild publish
```


package.json
```diff
@@ -8,7 +8,8 @@
     "test": "if [ -n \"$CI\" ]; then npm run init; fi && lerna run test",
     "init": "npm install && lerna bootstrap",
     "start": "cd ./packages/server/ && npm start",
-    "publish": "echo \"will eventually run: lerna publish\""
+    "publish": "echo \"will eventually run: lerna publish\"",
+    "build": "webpack  --config ./packages/server/webpack.config.js --output-path=`pwd`"
```
djfarrelly: Need to add `-p` once the `RPCClient` module has been transpiled to es5 and published 😉

.dockerignore
```diff
@@ -0,0 +1,6 @@
+.git
+node_modules
+packages/**/node_modules
+packages/**/__mocks__
+packages/**/__snapshots__
+packages/**/.storybook
```
djfarrelly: This greatly reduces the files sent to the Docker context when building the image and makes builds way faster 🏃

packages/server/index.js
```diff
@@ -1,20 +1,31 @@
...
-  return (req, res) => mw(req, res, () => next(req, res));
-};
+const middleware = [];
 
-const hotMiddleware = (next) => {
-  const mw = webpackHotMiddleware(compiler);
-  return (req, res) => mw(req, res, () => next(req, res));
-};
+if (process.env.NODE_ENV === 'development') {
+  const webpack = require('webpack');
```
djfarrelly: eslint doesn't like these nested require's, maybe we should fork this server entry point into `index.dev.js` and `index.prod.js`? 🤔

packages/server/package.json
```diff
@@ -29,7 +29,10 @@
   "dependencies": {
     "@bufferapp/micro-rpc": "0.1.0",
     "micro": "7.3.3",
-    "microrouter": "2.1.1"
+    "micro-rpc-client": "0.0.1",
+    "microrouter": "2.1.1",
+    "request": "2.81.0",
+    "request-promise": "4.2.1"
```
djfarrelly: Not sure how this was working before without these explicitly added here?

packages/server/service.js
```diff
@@ -16,4 +16,5 @@ module.exports = router(
   get('/', index),
   post('/rpc', rpc),
   get('/*', notFound),
-  post('/*', notFound));
+  post('/*', notFound),
```
djfarrelly: node 8 allows trailing commas in functions

pre-build.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
djfarrelly: This will run on Jenkins before the image is built!

packages/server/service.js
```diff
@@ -8,12 +9,21 @@ const {
...
+let bundleScript = '/static/bundle.js';
+if (process.env.NODE_ENV !== 'development') {
+  const staticAssetFile = join(__dirname, 'staticAssets.json');
+  const staticAssets = JSON.parse(fs.readFileSync(staticAssetFile, 'utf8'));
+  bundleScript = staticAssets['bundle.js'];
+}
 const html = fs.readFileSync(`${__dirname}/index.html`, 'utf8');
-const index = (req, res) => send(res, 200, html);
+const index = (req, res) => {
+  send(res, 200, html.replace('{{{bundle}}}', bundleScript));
```
hharnisc: I think it would make more sense to do the `html.replace` out of the request handler so it's cached in memory.

production.yaml
```diff
@@ -0,0 +1,72 @@
...
+        image: bufferapp/buffer-publish:test-1
+        ports:
+        - containerPort: 80
+        # env:
+        # - name: SESSION_COOKIE_SECRET
+        #   valueFrom:
+        #     secretKeyRef:
+        #       key: session-cookie-secret
+        #       name: publish-app
+        # livenessProbe:
```
hharnisc: Nice!

packages/server/index.js
```diff
@@ -1,20 +1,33 @@
...
+  const wpHotMiddleware = (next) => {
+    const mw = webpackHotMiddleware(compiler);
+    return (req, res) => mw(req, res, () => next(req, res));
+  };
+
+  middleware.push(wpHotMiddleware, wpMiddleware);
+}
+
+
+const server = micro(middleware.reduce((p, c) => c(p), service));
```
hharnisc: Was curious if starting `micro` was easier than installing micro into the docker image? Or maybe was this to make the `npm start` command be `node .`

package.json
hharnisc: I'll add that to the list of things to get done this week :smile:

packages/server/service.js
```diff
@@ -8,12 +9,35 @@ const {
...
+    isStaticAssetsLoaded = true;
+    resolve(staticAssets);
+  });
+});
+
 const html = fs.readFileSync(`${__dirname}/index.html`, 'utf8');
-const index = (req, res) => send(res, 200, html);
+const index = async (req, res) => {
+  const assets = await getStaticAssets();
+  send(res, 200, html.replace('{{{bundle}}}', assets['bundle.js']));
```
hharnisc: Would be simpler/faster to get the static assets outside the request handler I think and do a sync load of the json file at startup. Also only do the `html.replace` at load time so the contents of `html` are served from memory:

```js
const staticAssets = fs.readFileSync(/*static assets path*/, 'utf8');
const html = fs.readFileSync(`${__dirname}/index.html`, 'utf8').replace('{{{bundle}}}', assets['bundle.js']);
const index = (req, res) => send(res, 200, html);
```

packages/server/package.json
```diff
@@ -5,17 +5,17 @@
...
   },
   "jest": {
     "verbose": true,
     "coverageThreshold": {
       "global": {
-        "branches": 100,
-        "functions": 100,
-        "lines": 100,
-        "statements": 100
+        "branches": 20,
```
hharnisc: :sob:

packages/server/service.js
djfarrelly: I tried this, but I couldn't get the jest mock to work before the file was loaded itself. That was the key blocker and the only reason I refactored it get the tests to pass: dfd2387fe05e3ec0b75b9c227d2c9a5c0afa90c8

packages/server/index.js
djfarrelly: We're installing micro as a dependency already to use so it felt redundant to install as a global package in the docker file. Also I think the clarity of using our file as the entry point and not abstract some of the logic using their binary to start our app. With node 8 we shouldn't need their checks that they do also: https://github.com/zeit/micro/blob/3b6be39fa24da1bb19b69e687fe59db30fb3127b/bin/micro.js#L70

packages/server/package.json
djfarrelly: I put this low just during this phase. Refactoring a file can quickly drop percentage a lot and it feels more of a hinderance to always hit 100% at this stage :)

PR #15, opened 2017-07-03
Add/linkified text color config and divider

alvarezmelissa87:
- Add `Divider` component to `components` package
- Add `color` config to`LinkifiedText` component in `components` package


alvarezmelissa87:
Heya @hharnisc, @emily-plummer  😄 Would love your 👀 s on this one. Wanted to see how it feels to use the `hr` element here. 

Also, when this gets merged, curious if the `components` package will need version updating or if we're not doing that yet? 

Thanks!

emily-plummer:
Heya @alvarezmelissa87 this looks great! I think using `<hr>` is the closest semantic element we could use here!

alvarezmelissa87:
Great one @emily-plummer! 😄  Looks like for IE < 8 there are some style issues to change. Curious if we're supporting IE < 8? 

cc @hharnisc

emily-plummer:
@alvarezmelissa87 I don't believe we are! Only 10 upwards 🙌

packages/components/Divider/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+}) =>
+  (<div>
+    {children && <Text size={size} color={color}>{children}</Text>}
+    <hr style={dividerStyle} />
+  </div>
+  );
+
+Divider.propTypes = {
+  children: PropTypes.string,
+  color: PropTypes.oneOf([
```
djfarrelly: I wonder if this is something that we can create a "common buffer proptypes" module to make this code DRYer? I see that we're using this in a few places:
https://github.com/bufferapp/buffer-publish/blob/4b7a49611d8ea58b6dc4e3e0566aa91cb8e95cb8/packages/components/Text/index.jsx#L62-L87
https://github.com/bufferapp/buffer-publish/blob/4b7a49611d8ea58b6dc4e3e0566aa91cb8e95cb8/packages/components/Icon/index.jsx#L55-L80
That might create a bit of a copy-and-paste challenge when we need to update it. I think also importing these common lists will save in the overall bundle size when it's done 😄 

```js
import CustomPropTypes from '../custom-prop-types'

Something.propTypes = {
  color: CustomPropTypes.colors,
  // etc...
```

alvarezmelissa87: Heya @djfarrelly 😄  Yeah great one! I was actually working on that next 😂  Seems like a great idea and we definitely use it in many places, and will probably continue to use it anytime we use the `Text` component. 

Feels great to have a "common proptypes" module! 🙌  😄

hamstu: @djfarrelly @alvarezmelissa87 - another option could be to import the colors like so:

```
import colors from '../style/color';
```

And then do:

```
Something.propTypes = {
  color: Object.keys(colors)
  // ...
```

PR #17, opened 2017-07-03
Task/reorganize shared component posts

alvarezmelissa87:
Update `shared-components`:
- rename all post types by removing 'Dashboard' from name
- add `PostFooterButtons` component to be used by `PostFooter` for additional flexibility 
- update test snapshot with renamed components/ new component


alvarezmelissa87:
Not sure what's going on with this build fail - looks like it's web related - which wasn't changed at all in this PR 🤔 Will dig into it more on Wednesday. 😄 
cc @hharnisc

alvarezmelissa87:
Looks like the `web` package test snapshots weren't updated with the newest `Welcome to Buffer Publish!` copy - which was causing the build fail. I've gone ahead and updated the test snapshot for that and should be good to go! 😄

PR #24, opened 2017-07-05
Update server package tests

hharnisc:
Gets us back to :100: test coverage without getting in the way of development. This ignorins one `if` block that reads from the file system.

More importantly we're serving `index.html` from memory again without doing a string replace on every request.


packages/server/service.test.js
```diff
@@ -16,13 +18,18 @@ describe('service', () => {
...
+      webpackConfig.output.publicPath,
+      webpackConfig.output.filename,
+    );
+    const html = fs.readFileSync(join(__dirname, 'index.html'), 'utf8').replace('{{{bundle}}}', webpackBundleLocation);
     expect(body)
       .toBe(html);
-  });*/
+  });
+
   it('should should return 404 - GET', async () => {
```
KatieLo: curious ceiling cat comment unrelated to this PR:
is this `should` `should return 404`, as in `should return` is a thing that is expected to happen? How does that work? Or is that  a typo with two `shoulds`? 
really cool to see how these tests work 👀

hharnisc: Very much a typo! I've read right over that 100 times now :smile:

hharnisc:
Merging this pending :green_apple:

djfarrelly:
Hey @hharnisc, thank you for writing a test, I appreciate that. This does bring me back to our discussion in Slack from Monday and it feels like something we should get on the same page about. 

Ideologically, I'm not sure that we should make 100% test coverage a requirement at this point. I think it's great to measure code coverage, but I think it would be better to use a tool like Codecov ([see PR here](https://github.com/bufferapp/buffer-publish/pull/27)) to highlight what code is tested/not rather than enforce a percentage, especially one as a high and strict as 100%. Honestly, I feel that 100% test coverage is a hinderance to making good steady progress on this project. I know you may argue that starting good habits early will set us up better long term, but I think that testing certain straightforward functions isn't necessary at the moment. 

For example, this test is in essence just repeating the code within the function for `fs.readFileSync` and `replace`. We're pretty much copying the code to ensure that these two methods work which feels like we're just doing that to hit 100% rather than truly testing some logic that we wrote. For example if there was dynamic code that was doing some more complex where something could go wrong, I'd say it might be worth testing.

There are some great thoughts in [this Martin Fowler blog post](https://martinfowler.com/bliki/TestCoverage.html) which I think is useful for thinking about coverage. I highlight this statement:

> Like most aspects of programming, testing requires thoughtfulness. TDD is a very useful, but certainly not sufficient, tool to help you get good tests. If you are testing thoughtfully and well, I would expect a coverage percentage in the upper 80s or 90s. I would be suspicious of anything like 100% - it would smell of someone writing tests to make the coverage numbers happy, but not thinking about what they are doing.

I think for people contributing to this project they should understand the difference between writing a quality test for code that should be covered vs. just writing to get to 100% line coverage. We're moving from a world where we had little to no tests on our `buffer-web` code so we are working on a shift in how we do things. I believe understanding why code should be tested and writing fewer quality tests is more important to our team than aiming for a hard goal.

Overall, we're aiming to deliver value to users and I feel that we should be on the same page with these things. I'd be happy to talk about this next week on Monday if you'd like 😄

PR #26, opened 2017-07-06
async data fetching package

hharnisc:
- will also replace the login flow in the login package


hharnisc:
Gonna merge this so we can start using it! :green_apple:

PR #34, opened 2017-07-12
Update tabs package in web 

alvarezmelissa87:
- Adds `TabNavigation` to `ProfilePage`
- update click handler to correct name on `Tab component in `TabNavigation`


alvarezmelissa87:
Heya @emily-plummer - these are the super small changes we added yesterday. If it feels okay, maybe merging this in before moving forward on the routing stuff would be good to keep PR's small 😄

PR #35, opened 2017-07-12
Create settings package

alvarezmelissa87:
Create `settings` package


packages/settings/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@bufferapp/settings",
+  "version": "0.0.1",
+  "description": "Sent Posts package",
```
hamstu: Just a small one! 👍

alvarezmelissa87: Oh! Good catch @hamstu! 👍 Thank you! 😄

PR #39, opened 2017-07-13
Task/input autocomplete

emily-plummer:
Adds the new InputAutocomplete component! 🎉 This is a dummy component, most of the heavy lifting will be done by functions passed in through props :) Uses react-autocomplete, an autocomplete module built by the react team 😸

### DIFF
--- a/packages/components/InputAutocomplete/index.jsx
+++ b/packages/components/InputAutocomplete/index.jsx
@@ -1,0 +1,98 @@
+import React from 'react';
+import uuid from 'uuid';
+import Autocomplete from 'react-autocomplete';
+import PropTypes from 'prop-types';
+import Text from '../Text';
+import {
+  calculateStyles,
+} from '../lib/utils';
+import {
+  aquaHaze,
+} from '../style/color';
+
+const matchValueToItem = (item, value) =>
+  item.toLowerCase().indexOf(value.toLowerCase()) !== -1 ||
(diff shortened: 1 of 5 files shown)


packages/components/InputAutocomplete/index.jsx
```diff
@@ -0,0 +1,98 @@
...
+import {
+  calculateStyles,
+} from '../lib/utils';
+import {
+  aquaHaze,
+} from '../style/color';
+
+const matchValueToItem = (item, value) =>
+  item.toLowerCase().indexOf(value.toLowerCase()) !== -1 ||
+  item.toLowerCase().indexOf(value.toLowerCase()) !== -1;
```
alvarezmelissa87: heya @emily-plummer 😄  
Noticed both these lines (14,15) are checking the same thing.

emily-plummer: oooh whoops, copied this over from the example and forgot to fix this! Thanks! Great eye! 😂

packages/components/package.json
```diff
@@ -34,5 +34,8 @@
     "eslint": "3.19.0",
     "jest": "19.0.2",
     "storybook-addon-a11y": "3.0.0"
+  },
+  "dependencies": {
+    "react-autocomplete": "^1.5.9"
```
alvarezmelissa87: Heya @emily-plummer - I _think_ we want to use exact dependencies here 🤔   To be consistent with what we've been doing in publish.

PR #42, opened 2017-07-14
bring schedule setting components over from buffer-web-components

emily-plummer:
Brings in some of schedule setting components from buffer-web-components into shared-components :)


alvarezmelissa87:
Looks good pending build passing! 😄 Thanks @emily-plummer!

PR #60, opened 2017-07-20
Task/sent data transform

emily-plummer:
Transforms the sent post data and puts then in the sent tab! 🎉 

### Staging Deployment

### DIFF
--- a/packages/sent/reducer.js
+++ b/packages/sent/reducer.js
@@ -23,13 +23,13 @@ export default (state = initialState, action) => {
         ...state,
         loading: true,
       };
-    case `queuedPosts_${dataFetchActionTypes.FETCH_SUCCESS}`:
+    case `sentPosts_${dataFetchActionTypes.FETCH_SUCCESS}`:
       return {
         ...state,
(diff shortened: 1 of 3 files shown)


packages/sent/reducer.js
```diff
@@ -18,11 +18,19 @@ const initialState = {
...
-    //   return {
-    //     ...state,
-    //     loading: true,
-    //   };
+    case `sentPosts_${dataFetchActionTypes.FETCH_START}`:
+      return {
+        ...state,
+        loading: true,
+      };
+    case `queuedPosts_${dataFetchActionTypes.FETCH_SUCCESS}`:
```
alvarezmelissa87: oop! I think you meant `sentPosts` here 😄

alvarezmelissa87:
Heya @emily-plummer, since these are sent posts you will have to update this bit in `parsePosts`
https://github.com/bufferapp/buffer-publish/blob/master/packages/server/rpc/utils/postParser.js#L105

It should be `true` if `post.status === 'sent'`  - something like that.

PR #61, opened 2017-07-20
Fetch user data for `app-sidebar` 🎉 

hamstu:
This PR does user data fetching for the sidebar.

👉 **Important:** Currently returning a fake avatar image until we update the Buffer API to return avatars by default.

### Staging Deployment
👀  https://task-app-sidebar-fetch-data.publish.buffer.com


packages/server/rpc/utils/parseUser.js
```diff
@@ -0,0 +1,6 @@
+module.exports = userData => ({
+  id: userData.id,
+  email: userData.email,
+  avatar: 'https://pbs.twimg.com/profile_images/847849987841167360/WEVTxvUA_400x400.jpg', // userData.avatar,
+  /* TODO: Return the actual `avatar` once the API is doing that. */
```
hamstu: This is where we'll need to tweak once avatar's are working in the API.

alvarezmelissa87:
🎉

PR #64, opened 2017-07-24
Patch/inputtime

stevemdixon:
This PR patches InputTime to take a `displayTimeColon` prop so that we can achieve the design set out for our new Posting Schedule setting. 👍 

### Staging Deployment


KatieLo:
Hmm, I don't think I'm following why these changes need to be merged into the publish repo. @stevemdixon what did you have in mind here? I might be missing some context here. 

From my understanding, we're updating the inputtime component for the new schedule tab on current buffer-web, so I'm not following why this change is here in the new dashboard?

hamstu:
(_Also posted this in Slack_)
Hey @stevemdixon! Sorry you've been blocked on this. I've added a commit to your PR that I think might fix things. Those errors you posted are actually just warnings and don't fail the build (though we _do_ need to fix them, :sweat_smile:), the actual failure in your build happened because a test snapshot was not updated. By changing the `InputTime` you caused the snapshot for the `settings` package buffer-publish to be out-of-date, and that has to be updated manually.

stevemdixon:
Hey, @KatieLo & @hamstu! I'm super sorry for any confusion I've caused here. 😓  I saw that our buffer-components repo has been moved to a package of buffer-publish and thought that any changes we might need to make to our components might need to be in https://github.com/bufferapp/buffer-publish/tree/master/packages/components. I initially did all of my work in the buffer-components repo but ran into some permission issues with that and saw that that might be the case with that repo not having any updates in a while.

Thanks for helping fix that one up, @hamstu!

stevemdixon:
Hey, @KatieLo & @hamstu! I had a conversation with @hharnisc yesterday in Slack and it feels that buffer-publish is where we'll be making any component updates and also where we'll be adding any new components, too. We'll no longer be using the buffer-components repo to make any further changes to components. 👍 

The changes I needed to make here were part of InputTime, which is a component with buffer-components (or now buffer-publish/tree/master/packages/components), which is what we import and use in the ScheduleTableCell component in our buffer-web-components repo. For extra context and clarity, our Posting Schedule uses components from both of these repos, so that's why the changes were initially made in this repo versus the buffer-components repo.

I hope that clears a few things up! Meanwhile, I'll go ahead and close this PR as @hharnisc has hopped on to this in his PR here: https://github.com/bufferapp/buffer-publish/pull/75

PR #65, opened 2017-07-24
Keep selected profile in sync with browser history

hharnisc:
Selected profileId is using the URL as the source of truth

### Staging Deployment

_This repo is CI/CD enabled and staging deployment should be available at https://task-update-profile-route.publish.buffer.com :smile:_


hharnisc:
Looks like the push failed during the NPM install step, but then passed when running the PR - a timeout. Going to merge this since the tests passed

PR #66, opened 2017-07-24
[WIP] Transform profiles on the back end

hharnisc:
Transform profiles server side so the client doesn't need to do it in several different places!

### Staging Deployment

_This repo is CI/CD enabled and staging deployment should be available at https://task-transform-profiles-rpc.publish.buffer.com :smile:_


hharnisc:
Looks like there is some flakyness on Node 8 again. Tests look good on node 7 and on the PR again. Gonna merge this and try updating Lerna since they've got an official 2.0 :tada:

PR #67, opened 2017-07-24
Get tabId select state from URL

hharnisc:
This will match the way that profiles are selected. This means that the URL is the source of truth for the selection state both tabId and profileId
### Staging Deployment

_This repo is CI/CD enabled and staging deployment should be available at https://task-select-tab-from-url.publish.buffer.com :smile:_


hharnisc:
merging this one since branch build is :green_apple:

packages/tabs/index.js
```diff
@@ -1,16 +1,22 @@
...
 // default export = container
-export default withRouter(connect(
-  state => ({
-    profileId: state.tabs.profileId,
+export default connect(
+  (state, ownProps) => ({
+    selectedTabId: ownProps.tabId,
+  }),
+  (dispatch, ownProps) => ({
+    onTabClick: tabId => dispatch(push(generateProfilePageRoute({
```
emily-plummer: very cool!

PR #68, opened 2017-07-24
Task/post list metadata

alvarezmelissa87:
- export `PostLists` from `shared-components`
- Use `PostLists` in queue
- format posts to fit with `PostLists` component in queue

### Staging Deployment


alvarezmelissa87:
Heya @hamstu 😄 
I really like the idea of simplification but wasn't sure we'd want to do data manipulation in the reducer - but if it is in line with what a reducer is responsible for then it feels much more simple to do that! :raised_hands:

PR #71, opened 2017-07-25
Add profileTimezone and 24hr time config to settings state

alvarezmelissa87:
- Include `profileTimezone` in `settings` state for schedule purposes.
- Add `hasTwentyFourHourTimeFormat` from user data to `settings` state 
    -- display postingTimeForm in 24hr time if set
    -- display table time in 24hr time if set

- Update `settings` `ProfileSettings` story and props

### Staging Deployment


packages/settings/components/ProfileSettings/index.jsx
```diff
@@ -48,19 +50,24 @@ const ProfileSettings = ({
...
     return (<div>Loading...</div>);
   }
-  // TODO: set initial time to current time in profile timezones
-  // const now = moment().tz(timezone);
-  const initialValues = {
-    time: {
-      hours: 14, // now.hour()
-      minutes: 20, // now.minute()
-    },
-  };
```
alvarezmelissa87: Added this to allow storybook to pass for now - since the moment.now would create a new value each time.

emily-plummer: 👍

packages/settings/reducer.js
```diff
@@ -14,6 +15,8 @@ const initialState = {
   loading: false,
   days: [],
   items: timezones,
+  profileTimezone: 'Europe/London',
```
emily-plummer: Curious if we should leave the default as an empty string? Or maybe it doesn't matter too much since we should always have a profileTimezone coming in if there is a profile hooked up?

alvarezmelissa87: hey! 😄 Good one @emily-plummer This one should actually end up being unnecessary since the `due_at` and `sent_at` will be already set to the profile timezone - taking this out is in the TODO for this week 😄 Was just initially put in as a stub.

emily-plummer: Sounds good :)

PR #74, opened 2017-07-25
add bugsnag to client

hharnisc:
It might be worth bringing the bugsnag key into an env var to keep it in one place in a followup commit (since we add the same key in #73 )

### Staging Deployment

_This repo is CI/CD enabled and staging deployment should be available at https://add-bugsnag-to-client.publish.buffer.com :smile:_


hharnisc:
Merging this one!

PR #75, opened 2017-07-26
Add hovered arrows to InputTime noStyle config

hharnisc:
### Staging Deployment


hharnisc:
Had some tests time out - verifying that his works locally

packages/components/InputTime/index.jsx
```diff
@@ -61,7 +112,11 @@ const displayHour = (hour, select24Hours) => {
   return modHour === 0 ? 12 : modHour;
 };
 
-const InputTime = ({
+const displayTimeColonStyle = {
+  marginRight: '0.25rem',
```
stevemdixon: This is looking great! 👍  I have one tweak here, which would be to set the `font` key value to `fontSizeExtraSmall`. This'll keep the ":" at a size similar to the font size within the select elements. 😀

PR #76, opened 2017-07-26
Get profile timezone initial value to show

hamstu:
This PR ensures we show the profile's timezone value in the settings page. 

### Staging Deployment

https://profile-timezone-initial-value.publish.buffer.com


packages/settings/reducer.js
```diff
@@ -15,7 +15,7 @@ const initialState = {
   loading: false,
   days: [],
   items: timezones,
-  profileTimezone: 'Europe/London',
+  profileTimezone: '',
```
alvarezmelissa87: Heya @hamstu - this looks great! One thing is I do think we need a default timezone here. `PostingTimeForm` initial value calculations use timezone and it _might_ cause an error if we don't have one. I can double check this. 😄

hamstu: Hey @alvarezmelissa87 – ah cool! I didn't see any errors; but that doesn't mean there aren't any 😅  Would love to know if you think this has mixed things up at all. I removed it here since I didn't want it showing in the UI before the correct one loads. (Once we have proper loading states, we probably won't have that issue, though.)

PR #79, opened 2017-07-26
Fix tab style ✨

hamstu:
Just a quick tweak to the tab styling to keep it more inline with the Invision/mockup.


hamstu:
🌞 Anytime Emily!

PR #89, opened 2017-07-27
Task/infinite loading for queue

alvarezmelissa87:
### Purpose
- Implement infinite loading for `queue`

#### Staging Deployment


packages/queue/reducer.js
```diff
@@ -15,41 +12,47 @@ export const actionTypes = {
...
 
-  return postLists;
+const increasePageCount = (page) => {
+  page += 1;
+  return page;
 };
 
+const determineIfMoreToLoad = (action, currentPosts) =>
+  (action.result.total > (currentPosts.length + action.result.updates.length));
+
```
alvarezmelissa87: Don't keep fetching if we've got all the pending updates loaded.

We could listen for a push event eventually for when an update is created to reset `moreToLoad` to false so that the user can see the newest stuff.

packages/queue/index.js
```diff
@@ -3,12 +3,34 @@ import { connect } from 'react-redux';
 // load the presentational component
 import QueuedPosts from './components/QueuedPosts';
 
+
+const formatPostLists = (posts) => {
+  const postLists = [];
+  let day;
+  let newList;
+
+  for (let i = 0; i < posts.length; i++) { // eslint-disable-line
```
hharnisc: I think this could be done with a `forEach` (or possibly even a `reduce`). For loops written like this are blocking even loop and will introduce render jenk.

packages/queue/reducer.js
```diff
@@ -15,41 +12,47 @@ export const actionTypes = {
...
 export default (state = initialState, action) => {
   switch (action.type) {
     case `queuedPosts_${dataFetchActionTypes.FETCH_START}`:
       return {
         ...state,
-        loading: true,
+        loading: !action.args.isFetchingMore,
```
hharnisc: Excellent use of `action.args` here :raised_hands:

packages/web/components/ProfilePage/index.jsx
```diff
@@ -71,39 +73,81 @@ TabContent.defaultProps = {
...
+    }).isRequired,
+    handleScroll: PropTypes.func.isRequired,
+    loadingMore: PropTypes.bool.isRequired,
+    moreToLoad: PropTypes.bool.isRequired,
+    page: PropTypes.number.isRequired,
+  }
+
+  handleScroll ({ ev, profileId, page, tabId }) { // eslint-disable-line
+    const { loadingMore, moreToLoad } = this.props;
+    const element = ev.target;
```
hharnisc: Very cool to see we're able to grab the element from the event :raised_hands: This means we can keep the PropfilePage a functional component and drop the `class`.

Also we'll probably want to debounce the scroll handling function with something like this: https://davidwalsh.name/javascript-debounce-function

alvarezmelissa87:
Heya @hharnisc -thanks for taking a look! 😄 
I went ahead and updated with the requested changes. 
Currently implementing the debounce of the onScroll - happy to put that in a separate PR if this feels like it's getting too big 😄

alvarezmelissa87:
Tested on Chrome, Firefox - will be tested on Safari next

alvarezmelissa87:
Hm looks like the onScroll is never firing in Safari - investigating that now.

packages/web/components/ProfilePage/index.jsx
```diff
@@ -28,6 +30,7 @@ const contentStyle = {
   flexDirection: 'column',
   marginLeft: '1rem',
   marginRight: '1rem',
+  height: '100vh',
```
hamstu: Might be worth commenting here to explain why we had to add this. 👍

packages/web/components/ProfilePage/index.jsx
```diff
@@ -71,13 +74,55 @@ TabContent.defaultProps = {
   tabId: '',
 };
 
+// Returns a function - as long as it continues to be invoked, will not
+// be triggered. The function will be called after it stops being called for
+// N milliseconds. If `immediate` is passed, trigger the function on the
+// leading edge, instead of the trailing. Taken from underscore.js
+function debounce(func, wait, immediate) {
```
hamstu: Definitely a good one to put into a utils package I think. In another PR for sure though. Alternatively we could just pull in this method from an existing npm package, like this one: https://www.npmjs.com/package/lodash.debounce

PR #90, opened 2017-07-28
Add new temp login page

djfarrelly:
### Purpose
This login page will replace the console & rpc login method.

### Notes
The new login will set the jwt in the `session` cookie which will allow the front-end to have the cookie already when redirected to the dashboard after successful login.

We should probably move the `server/middleware.js` to `/server/lib/middleware.js` for the sake of organization.

### Review
Hey @hharnisc, this one is not fully cleaned up yet, but it's working. To follow this PR we should implement logout and remove the old login/logout rpc methods. Will remove some of the old login/logout RPC endpoints once we have full login/logout w/ TFA implemented.

#### Staging Deployment
https://task-add-login-page.publish.buffer.com :smile:_


packages/async-data-fetch/reducer.js
```diff
@@ -1,12 +1,22 @@
+import Cookie from 'js-cookie';
+
 export const actionTypes = {
   FETCH: 'FETCH',
   FETCH_START: 'FETCH_START',
   FETCH_SUCCESS: 'FETCH_SUCCESS',
   FETCH_FAIL: 'FETCH_FAIL',
 };
 
-// TODO: grab the token from the cookie at init
```
hharnisc: My bad here! The TODO here was completed by the `login_${actionTypes.FETCH_SUCCESS}` action handled below. I think this could be revered back to its original state - keeping the cookie/login logic in the Login package.

packages/login/middleware.js
```diff
@@ -18,18 +18,19 @@ export default ({ dispatch }) => next => (action) => {
   next(action);
   switch (action.type) {
     case 'APP_INIT': {
-      const session = Cookie.get('session', {
+      // TODO - Remove this duplicate code - this cookie is now
```
hharnisc: I think this is a good pattern to keep here - dispatching the successful `login` event so other packages can catch it.

packages/async-data-fetch/reducer.js
djfarrelly: @hharnisc, I was thinking about fully removing the login package once the login/logout pages were completed and then fully extracted into another service. I don't know what purpose the login package would really serve then? This is def some of the cleanup that I assumed we could do after this logout and the tfa login is fully complete. What do you think?

packages/login/middleware.js
djfarrelly: Hey @hharnisc, similar to my comment above, I think since login won't be handled by the client side itself, maybe we can remove this fully and the web package can handle the app init in a centralized place?

PR #95, opened 2017-08-01
Fix app-sidebar buttons in Firefox + Improve a11y

hamstu:
### Purpose

Fix the popover menus in Firefox.

### Notes

Also added keyboard support to popover menus using roving tabIndex! :)
https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_roving_tabindex


emily-plummer:
This is AMAZING @hamstu!!! I just tested it out on Voice Over and it's so smooth! Wow! I'm super impressed with the creative solution you came up with & all your research into roving tab indexes! It's really touching to see you take on an accessibility challenge with such motivation and effort! ❤️ Looked over the code and I think it looks great! 👍 🎉 ✨ 🌷 😸

PR #96, opened 2017-08-01
{WIP} Task/select inputs

emily-plummer:
### Purpose
This PR updates the select elements in the posting time form to use the shared select component with it's shiny styles :) I've also refactored the Select component to take in an options array full of objects which contain a `value` and `name` so that we can display something different than what the API needs to consume (ex. weekdays are displayed a "Monday, Tuesday, etc" but we need to store "mon, tue, etc")

#### Staging Deployment


emily-plummer:
Commenting so I remember: need to take another look at the weekdays options... we might need to refactor options to take an array of objects that looks like this { value: '', name: ''} so that we can use a "display" name and the name needed to communicate with the API

alvarezmelissa87:
Looks good to ship pending build passing 😄

PR #97, opened 2017-08-01
Task/infinite loading for sent

alvarezmelissa87:
### Purpose
Add infinite loading to sent package
### Notes
Heya @hharnisc - this seems to work but there's an issue where, when switching tabs (from queue to sent, for example) if the scroll bar is not at the top for the current tab, the scroll event is automatically triggered and so a call for more posts will occur as soon as tab is switched over.

Seems like the solution here is to unbind/rebind the scroll event and/or ensure we scroll to top when we're switching tabs. 
Curious to hear your thoughts on that 😄 

Adding task in basecamp for solutions for getting scrollbar back to top when switching tabs and unbinding scroll event in subsequent PR to keep this one smaller if that makes sense.
#### Staging Deployment


packages/web/components/ProfilePage/index.jsx
```diff
@@ -170,15 +167,19 @@ ProfilePage.propTypes = {
...
-    loading: state.queue.loading,
-    loadingMore: state.queue.loadingMore,
-    moreToLoad: state.queue.moreToLoad,
-    page: state.queue.page,
-    posts: state.queue.posts,
-    total: state.queue.total,
-    translations: state.i18n.translations.example, // all package translations
-  }),
+  (state, ownProps) => {
+    const splitPath = ownProps.history.location.pathname.split('/');
```
alvarezmelissa87: Using history here instead of match.path... since match has a slight lag where it is still set to the last tab before it switches over. `history.location.pathname` is always current.

packages/sent/components/SentPosts/index.jsx
```diff
@@ -28,13 +44,19 @@ const SentPosts = ({
       <PostLists
         postLists={postLists}
       />
```
alvarezmelissa87: Thinking of breaking the loading bit into a component so both queue/sent can share it 🤔

PR #103, opened 2017-08-03
Queue empty state

emily-plummer:
### Purpose
Adds a generic empty queue state :)
### Review
cc @hamstu 

#### Staging Deployment

### DIFF
--- a/packages/queue/components/QueuedPosts/index.jsx
+++ b/packages/queue/components/QueuedPosts/index.jsx
@@ -28,10 +28,9 @@ const loadMoreStyle = {
 };
 
 const QueuedPosts = ({
+  total,
   loading,
   loadingMore,
-  hasSentPosts,
-  total,
   postLists,
   onCancelConfirmClick,
   onDeleteClick,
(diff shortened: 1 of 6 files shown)


packages/queue/components/QueuedPosts/index.jsx
```diff
@@ -39,14 +41,20 @@ const QueuedPosts = ({
...
   return (
     <div>
       <div style={composerStyle}>
         <Input
           placeholder={'What would you like to share?'}
         />
       </div>
+      {total < 1 &&
+        <EmptyState
+          title={'It looks like you haven\'t got any posts in your queue!'}
```
hamstu: I wonder if we should use translations more? The same way @hharnisc did in `profile-sidebar`. Not a high priority one, since I have a lot of strings I've been meaning to move over too, but it would be cool to move towards that.

packages/shared-components/EmptyState/index.jsx
```diff
@@ -0,0 +1,61 @@
...
+    width: '100%',
+  };
+  return (
+    <div style={wrapperStyle}>
+      <div style={containerStyle}>
+        {heroImg &&
+          <img style={imageStyle} alt="" src={heroImg} />}
+        {title &&
+          <div style={headerStyle}>
+            <Text size={'large'} weight={'bold'}>
```
hamstu: What do you think about using `size="large"`, `weight="bold"`, etc. instead? I think it's a bit cleaner/easier to read, since you don't need the curly braces. I've been doing this in other places and I feel like it would be good to standardize, esp. since eslint doesn't care either way.

If it feels good to you, I'd say this would apply to the whole PR.

packages/shared-components/EmptyState/story.jsx
```diff
@@ -0,0 +1,15 @@
...
+import EmptyState from './index';
+
+
+storiesOf('EmptyState')
+  .addDecorator(checkA11y)
+  .add('default', () => (
+    <EmptyState
+      title={'It looks like you haven\'t got any posts in your queue!'}
+      subtitle={'Click the box above to add a post to your queue :)'}
+      heroImg={'https://s3.amazonaws.com/buffer-publish/images/fresh-queue.png'}
```
hamstu: So fresh! 🍓  ✨ 

One request; could you use this `@2x` version instead and then scale down to the normal size so it's not fuzzy on Retina?
https://www.dropbox.com/s/bwfvz2f49xpck55/fresh-queue%402x.png?dl=0

packages/queue/components/QueuedPosts/index.jsx
hamstu: Maybe something we can do in a separate PR, time permitting.

packages/shared-components/EmptyState/story.jsx
emily-plummer: yes! thank you!

packages/shared-components/EmptyState/index.jsx
emily-plummer: will do!

packages/queue/components/QueuedPosts/index.jsx
emily-plummer: Agreed!

PR #106, opened 2017-08-03
Task/loading animation

emily-plummer:
### Purpose
Adds the loading buffer logo animation:
- between login & the tabs rendering (on the default page)
- while the queue tab loads
- while the sent tab loads
- while the settings tab loads

### Review
@hamstu 

#### Staging Deployment


packages/components/Image/index.jsx
```diff
@@ -11,6 +11,8 @@ const Image = ({ src,
   height,
   maxHeight,
   maxWidth,
+  marginTop,
+  marginBottom,
```
emily-plummer: added this so I can use Image component in another spot (will replace in separate branch)

PR #111, opened 2017-08-04
📜 Scroll to top when changing tabs, scroll tweaks, etc.

hamstu:
### Purpose

* Move the scroll functionality into a separate component
* Fix style for 'loading more...'
* Style scrollbars in webkit

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -9,4 +9,4 @@ RUN npm install -g nodemon
 
 EXPOSE 80
 
-CMD nodemon --watch ./packages/server -e html,js,json packages/server/index.js
+CMD nodemon --ignore 'index.html' --watch ./packages/server -e html,js,json packages/server/index.js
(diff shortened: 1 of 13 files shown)


Dockerfile.development
```diff
@@ -9,4 +9,4 @@ RUN npm install -g nodemon
 
 EXPOSE 80
 
-CMD nodemon --watch ./packages/server -e html,js,json packages/server/index.js
+CMD nodemon --ignore 'index.html' --watch ./packages/server -e html,js,json packages/server/index.js
```
emily-plummer: oooh what's this for?? 👀

hamstu: Ah yes! So this (and the change in the server js) was so I could change the `index.html` file without having to wait for an unnecessary restart and rebuild of webpack! :)

packages/shared-components/ScrollableContainer/index.jsx
```diff
@@ -0,0 +1,72 @@
...
+      if (!immediate) func.apply(context, args);
+    };
+    const callNow = immediate && !timeout;
+    clearTimeout(timeout);
+    timeout = setTimeout(later, wait);
+    if (callNow) func.apply(context, args);
+  };
+};
+
+class ScrollableContainer extends Component {
```
alvarezmelissa87: Yeeeeaah! 🙌 🎉

PR #117, opened 2017-08-09
Add TFA to temp login

djfarrelly:
### Purpose
To allow people to login w/ TFA code. With the temp login setup this will allow Buffer admins to securely login and test this app.

### Notes
This stores if the user is in the process of entering a TFA code in their session. We check the code then log them in. The login and session code should be extracted into a separate service and package.

### Review
hey @hharnisc and @alvarezmelissa87, although this needs quite a bit of refactor and extraction into a separate service and package I wanted to get this in there for admins to test with their real accounts 😄 I think we'll need to write more up about how we are architecting sessions, logins, etc.

#### Staging Deployment

https://task-login-tfa.publish.buffer.com :smile:_


packages/server/lib/bufferApi.js
```diff
@@ -16,3 +16,18 @@ bufferApi.signin = ({ email, password }) => rp({
   },
   json: true,
 });
+
+// TODO - Add backup phone number usage
+// TODO - Add visitor_id linking when session is used on marketing
+bufferApi.tfa = ({ userId, code }) => rp({
+  uri: `${BUFFER_API_ADDR}/1/user/twostep.json`,
+  method: 'POST',
+  strictSSL: false, // TODO - In prod this should be secure
```
alvarezmelissa87: If NODE_ENV is available here, could we use something like this for strictSSL: `!(process.env.NODE_ENV === 'development')`

PR #120, opened 2017-08-14
Prefix package names with publish

hharnisc:
### Purpose

Helps us with package name collisions with `analyze-*`

### Notes

Not every package requires the prefix since, things like packages shared between applications don't have the prefix (i.e. components)

### Review

Did I miss anything?

#### Staging Deployment

https://fix-app-prefixed-package-names.publish.buffer.com :smile:_


Dockerfile.development
```diff
@@ -4,8 +4,7 @@ RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
 ENV NODE_ENV development
-
-RUN yarn install -g nodemon
+RUN npm install -g nodemon
```
alvarezmelissa87: Oop! good catch @hharnisc - thank you! 😄

PR #129, opened 2017-08-23
Fix regression: settings page error before render

hharnisc:
### Purpose

Fix regression causing settings page to error before rendering

### Notes

@hamstu I think we need to move the caching into it's own caching. The profile page has gotten sufficiently complex - at least we should move the `connect` logic into a container component and out of the component folder.

### Review

Does this work on your machine?

#### Staging Deployment

https://fix-settings-page-render.publish.buffer.com :smile:_


hharnisc:
Tests failing due to the A11Y checks timing out

hharnisc:
Going to merge this one since it allows us to see settings again

PR #130, opened 2017-08-23
Add composer package

pioul:
### Purpose

Add Composer package to buffer-publish

### Notes

Lots of moving pieces still:

- Mock data is being passed in, many things are hardcoded
- Composer is being tested inside the QueuedPosts component straight from a click handler
- Composer shows up sporadically

Next steps:

- [x] Bring in a few additional dependencies that MC shared with buffer-web into MC: (shared/appDispatcher, shared/libs/twitter.text, and the modal
     - Bring in `shared/appDispatcher`, we don't need a shared dispatcher anyway
     - `shared/libs/twitter.text`: import npm package instead
     - `shared/views/components/modal/...`: replace buffer-web's modal component+store with the Popover component from buffer-components
- [x] Pass actual data in (retrieving + formatting all data pieces from buffer-publish)
- [ ] Improve reliability, write small abstraction on buffer-publish to ease initialization
- [ ] Clean up:
     - [x] Bring composer-specific deps inside own package
     - [ ] Use internal Popover component instead of `__legacy-buffer-web-shared-components__/modal`
     - [ ] ExtractTextPlugin had to be disabled for now: is a JS+CSS mixed bundle more viable now than it was before? Otherwise bring ExtractTextPlugin back in this new setup to have separate JS and CSS bundles
- [x] Add normalize.css to buffer-publish? Are styles normalized at all in Publish? If not, can MC do without it to avoid adding a new dep? (see commented import statement in App.jsx)
- [ ] Re-enable Pusher, w/o it being exposed globally
- [x] Add remaining assets (e.g. fonts)
- [x] Make urls absolute (instead of relative urls meant for the buffer-web app environment, use absolute urls for all external assets and endpoints – based on local/prod environment)
- [ ] Make sure MC build config in buffer-publish doesn't leave anything behind (double-check package versions are exactly the same, that no plugins/transforms are left behind for both JS and CSS)
- [x] Convert normalize.css to individualized per-component/element rules, and create wrapper components where needed to reuse styles (won't be a quick win, but the alternative is just as time-consuming and not as clean: we'd need to scope normalize.css rules, and then also use that same scope and apply it to all CSS Modules inside MC to preserve precedence)
- [ ] Reset MC store when starting fresh Composer session (both for Storybook and buffer-publish; on mount / unmount / new props?)
- [ ] Add `<link href="//icons.buffer.com/0.16.4/buffer-icons.css" rel="stylesheet" type="text/css"/>` to buffer-publish's head

#### Staging Deployment


package.json
```diff
@@ -24,6 +24,12 @@
   },
   "homepage": "https://github.com/bufferapp/project-donut#readme",
   "devDependencies": {
+    "@bufferapp/buffer-js-api": "0.3.0",
```
hharnisc: Hey @pioul I think it would be cleaner to bring these new dependencies in to the `composer` package. They'll get hoisted up to the top level `node_modules` with `Lerna` and when we publish `composer` package all of the deps will be satisfied.

pioul: @hharnisc Great point, will do! 👍 I'm curious where you'd draw the line: I'm tempted to bring all of those new deps into the local package, not sure if you'd leave some here as common dependencies? (e.g. moment will probably end up being a common dep as more packages are added, are we keeping it local until then?)

alvarezmelissa87:
PR to ensure imagedimensions_key comes in correctly: https://github.com/bufferapp/buffer-web/pull/13857

djfarrelly:
@alvarezmelissa87 did we mean to check in the `yarn-error.log`? just noticed that locally, it may be a file to gitignore 😄

PR #131, opened 2017-08-23
Task/help and support not link

alvarezmelissa87:
### Purpose
`Help & Support` header should not show pointer on hover as it is not a link. 
This adds 'curser: default' style to header.
#### Staging Deployment


hamstu:
✨  Thanks Melissa!

PR #132, opened 2017-08-23
sidebar icon click causes reload if on active tab

alvarezmelissa87:
### Purpose
When clicking on navigation icon, if on page for current selected navigation icon - do not open new tab with same page - reload current page. 
### Review
Just some questions on this one @hamstu 😄 Thanks!
#### Staging Deployment


packages/app-sidebar/components/AppSidebar/index.jsx
```diff
@@ -40,7 +40,7 @@ const AppSidebar = ({
   <nav style={style} aria-label="sidebar" role="menubar">
     <BufferLogo />
 
-    <PopoverButton icon={<PublishIcon />} active={activeProduct === 'publish'} label="Publish" href="/" />
+    <PopoverButton icon={<PublishIcon />} active={activeProduct === 'publish'} label="Publish" reload href="/" />
```
alvarezmelissa87: Heya @hamstu - curious to hear your thoughts since you may have more context on this.

I was thinking we could set `reload` to something similar to `active` - so that it only sets to `true` if we are on that current product already.

I wasn't sure if `activeProduct` was _always_ going to be 'publish' or if potentially we'd have navigation items that would be within this repo as well. 🤔 If so, it would make sense to do something like 'if we are currently on the product for the icon we are clicking just do a reload' (changes in `handleClick` below).

alvarezmelissa87:
Hm, not sure what's going on with the build on this one 🤔 Passes fine locally but keeps failing on the accessibility audit bit for the components package because of a timeout.

PR #134, opened 2017-08-23
Task/delete post

emily-plummer:
### Purpose
This PR add the ability to delete posts from the queue! @alvarezmelissa87 also did some ground work to transform the posts into nested keyed objects so that we can easily get posts by profile ids :) 

### Review
would love a review @alvarezmelissa87!
#### Staging Deployment


packages/queue/middleware.js
```diff
@@ -1,5 +1,6 @@
 import { actionTypes } from '@bufferapp/publish-profile-sidebar';
 import { actions as dataFetchActions } from '@bufferapp/async-data-fetch';
+import { actionTypes as queueActionTypes } from './reducer';
```
alvarezmelissa87: It might make more sense to rename the `publish-profile-sidebar`'s  actionTypes to `profileSidebarActionTypes` and keep the queue's action types as `actionTypes` since we're in the queue package here. More of a nitpick but it might make more sense 😄

packages/queue/reducer.js
```diff
@@ -33,8 +33,59 @@ const increasePageCount = (page) => {
...
 
-const determineIfMoreToLoad = (action, currentPosts) =>
-  (action.result.total > (currentPosts.length + action.result.updates.length));
+const determineIfMoreToLoad = (action, currentPosts) => {
+  const currentPostCount = Object.keys(currentPosts).length;
+  const resultUpdatesCount = Object.keys(action.result.updates).length;
+  return (action.result.total > (currentPostCount + resultUpdatesCount));
+};
+
+const postReducer = (state, action) => {
```
alvarezmelissa87: Niiiiiice! 🙌

packages/queue/reducer.js
```diff
@@ -33,8 +33,59 @@ const increasePageCount = (page) => {
...
+  switch (action.type) {
+    case actionTypes.POST_CLICKED_DELETE:
+      state.isConfirmingDelete = true;
```
alvarezmelissa87: So on each of these where we're updating the state - like on line 45 - we want to maintain immutability by making a copy of the state -  that would be the more Redux-ey way to do it. 

Maybe something like this: `return { ...state, isConfirmingDelete: true };`
Here's a link for the why: http://redux.js.org/docs/recipes/reducers/ImmutableUpdatePatterns.html

packages/queue/middleware.js
emily-plummer: good point :) I'm going to go back and fix this!

packages/queue/reducer.js
emily-plummer: Thank you! I was wondering about that 🤔

packages/server/rpc/deletePost/index.js
```diff
@@ -0,0 +1,20 @@
...
+      uri: `${process.env.API_ADDR}/1/updates/${updateId}/destroy.json`,
+      method: 'POST',
+      strictSSL: !(process.env.NODE_ENV === 'development'),
+      qs: {
+        access_token: session.accessToken,
+      },
+    })
+    .then(() => 'OK')
+    .catch((err) => {
+      throw createError({ message: err.message });
```
alvarezmelissa87: 🤔 Not sure how we'd want to  handle an error here - perhaps we can log it instead of throwing an error? cc @hamstu  for thoughts 😄

PR #140, opened 2017-08-25
add lightbox

emily-plummer:
### Notes
This PR adds the image lightbox to image & multiple image posts! 🎉 
#### Staging Deployment


packages/shared-components/.yarnclean
```diff
@@ -0,0 +1,42 @@
+# test directories
```
hharnisc: Lets avoid running checking in these `yarn clean` files for now

packages/shared-components/LightboxWrapper/index.jsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
```
hharnisc: Looking at this wrapper its just passing through values, I think this could be removed and then use `Lightbox` directly.

packages/shared-components/ImagePost/index.jsx
```diff
@@ -6,6 +6,7 @@ import {
   LinkifiedText,
 } from '@bufferapp/components';
 import Post from '../Post';
+import LightboxWrapper from '../LightboxWrapper';
```
hharnisc: I think we can use `Lightbox` directly here and not have to use the wrapper.

packages/shared-components/ImagePost/index.jsx
```diff
@@ -75,6 +78,11 @@ const ImagePost = ({
           border={'rounded'}
           objectFit={'cover'}
         />
+        <LightboxWrapper
```
hharnisc: I'm not exactly sure why but I looks like the `Lightbox` on the image post is not opening for me locally or in the branch build `http://task-lightbox.publish.buffer.com`

packages/web/components/ProfilePage/index.jsx
```diff
@@ -46,6 +46,7 @@ const TabContent = ({ tabId, profileId }) => {
           onDeleteConfirmClick={() => { }}
           onEditClick={() => { }}
           onShareNowClick={() => { }}
+          onImageClick={() => { }}
```
hharnisc: Is this to silence an error?

emily-plummer: I actually don't know why these are here 🙈 I don't think they are doing anything? I tried taking them out and it doesn't seem to make a difference

packages/shared-components/ImagePost/index.jsx
emily-plummer: 👍 good call, I was meaning to take this out once I decided to go stateless :)

PR #146, opened 2017-08-31
Share Now 💫

emily-plummer:
### Purpose
This PR does the initial ground work for sharing posts through the share now button.
- Captures "Share Now" click handler in redux connect
- Dispatches event to async data fetch with sharePostNow
- Uses RPC endpoint to handle sharePostNow
- Uses the isWorking state on the Post to disable the post while sharing
- Handles pusher event to remove shared post from queue

#### Staging Deployment


packages/server/rpc/index.js
```diff
@@ -7,6 +7,7 @@ const queuedPostsMethod = require('./queuedPosts');
 const sentPostsMethod = require('./sentPosts');
 const userMethod = require('./user');
 const deletePostMethod = require('./deletePost');
+const sharePostNow = require('./sharePostNow');
```
alvarezmelissa87: Good stuff, @emily-plummer! 🙌 
Just a nitpick but we probably want to call this `sharePostNowMethod` to keep consistent with the others 😄

packages/server/rpc/sharePostNow/index.js
```diff
@@ -0,0 +1,17 @@
...
+    rp({
+      uri: `${process.env.API_ADDR}/1/updates/${updateId}/share.json`,
+      method: 'POST',
+      strictSSL: !(process.env.NODE_ENV === 'development'),
+      qs: {
+        access_token: session.accessToken,
+      },
+    })
+      .then(result => console.log(result)),
+);
```
alvarezmelissa87: Heya @emily-plummer -  are we relying on pusher to detect the share success and then update the store based on that?

packages/sent/reducer.js
```diff
@@ -89,6 +106,17 @@ export default (state = initialState, action) => {
...
+      profileId = getProfileId(action);
+      if (profileId) {
+        return {
+          byProfileId: {
+            ...state.byProfileId,
+            [profileId]: profileReducer(state.byProfileId[profileId], action),
+          },
+        };
+      }
+      return state;
```
alvarezmelissa87: After a rebase, should be able to remove this since the code is the same for each case and without a break/return it will continue to the end. 👍

packages/queue/reducer.js
```diff
@@ -137,7 +142,16 @@ const profileReducer = (state = profileInitialState, action) => {
...
+    case actionTypes.POST_SHARE_NOW:
+      return {
+        ...state,
+        posts: postsReducer(state.posts, action),
+      };
+    case actionTypes.POST_SENT:
+      return {
+        ...state,
+        posts: postsReducer(state.posts, action),
+      };
```
alvarezmelissa87: If both POST_SHARE_NOW and POST_SENT are before POST_ERROR - we should be able to remove the `return` code for each case as lack of return/break should go continue on to the next case. 

Since the logic is basically repeated for these, it can trickle down to POST_ERROR and be handled there.

packages/queue/reducer.js
```diff
@@ -91,6 +86,16 @@ const postsReducer = (state = {}, action) => {
...
+    case actionTypes.POST_SHARE_NOW:
+      return {
+        ...state,
+        [action.updateId]: postReducer(state[action.updateId], action),
+      };
+    case actionTypes.POST_SENT:
+      return {
+        ...state,
+        [action.post.id]: postReducer(state[action.post.id], action),
+      };
```
alvarezmelissa87: Same as comment below 😄

packages/server/rpc/sharePostNow/index.js
emily-plummer: yes! I can probably take this out, I just put it in for testing purposes to make sure that the share was successful before I got Pusher working :)

PR #147, opened 2017-08-31
Task/blue hovered input time

hharnisc:
### Purpose

Remove the arrows that show on hover and instead change the select text to blue on the `minimal` InputTime

### Notes

InVision: https://buffer.invisionapp.com/share/8YD8YKL5C#/screens/250875036

### Review

How does the UX feel?

#### Staging Deployment

https://task-blue-hovered-input-time.publish.buffer.com :smile:_


hharnisc:
Merging this and then publishing the package changes :smile:

PR #148, opened 2017-08-31
remove color contrast test from axe-core tests

emily-plummer:
### Purpose
Removes the color contrast tests from our axe-core tests, which were slowing things down a bit :) Should help solve the timeout issues we were seeing. For now, let's test color contrast with the axe extension! https://chrome.google.com/webstore/detail/axe/lhdoppojpmngadmnindnejefpokejbdd?hl=en-US

#### Staging Deployment

### DIFF
--- a/packages/components/lib/a11yTestHelper.js
+++ b/packages/components/lib/a11yTestHelper.js
@@ -42,7 +42,11 @@ const testComponentA11y = (app, config = []) => {
   const node = findDOMNode(wrapper.component);
 
   if (typeof config === 'function') {
-    config = {};
+    config = {
+      rules: {
+        'color-contrast': { enabled: false },
+      },
+    };
   }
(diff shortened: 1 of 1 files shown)


packages/components/lib/a11yTestHelper.js
```diff
@@ -42,7 +42,11 @@ const testComponentA11y = (app, config = []) => {
   const node = findDOMNode(wrapper.component);
 
   if (typeof config === 'function') {
-    config = {};
+    config = {
```
hamstu: Looks like this probably isn't having any effect since it's within the `if (typeof config === 'function')` block, and from what I can tell we never pass `config` into this function.

emily-plummer: 🙈 WHOOPS!

PR #149, opened 2017-08-31
minimize InputTime width

hharnisc:
### Purpose

Minimize the InputTime width so it works better in the schedule tab on Firefox

### Notes

This is to fix a bug in firefox that cause AM/PM to get cutoff in the schedule tab

### Review

How does this branch look?

#### Staging Deployment

https://task-minimize-input-time-width.publish.buffer.com :smile:_


hharnisc:
Merging this one :green_apple:

PR #151, opened 2017-09-01
Fix InputTime on IE 11

hharnisc:
### Purpose

InputTime component was using Object.keys to generate arrays of numbers. Using an IE compatible solution.

### Notes

Internet Exploader? :thinking: 

### Review

Does this work in IE, Firefox, Chrome and Safari for you?

#### Staging Deployment

https://fix-input-time-ie-11.publish.buffer.com :smile:_


hharnisc:
merging :green_apple:

Issue #143, opened 2017-08-30
Enforce https profile images


djfarrelly:
### Purpose
To push the page to be fully secure and use only https resources, we made a change a while back in buffer-web to change all profile urls to https:

<img width="511" alt="screen shot 2017-08-30 at 5 13 37 pm" src="https://user-images.githubusercontent.com/1509457/29895438-1a0efa14-8da7-11e7-848f-c42259aef953.png">

vs.

<img width="285" alt="screen shot 2017-08-30 at 5 17 39 pm" src="https://user-images.githubusercontent.com/1509457/29895450-27838dea-8da7-11e7-8504-c035bd5ee75a.png">

PR #154, opened 2017-09-05
Use https avatar over default

djfarrelly:
### Purpose
Fixes: #143

### Notes
This was originally completed in #145, but due to a rebase after #130 that PR which moved the file, the PR was automatically closed.

We may want to simplify this overall and only use a single avatar url instead of two. I believe `avatar_https` will default to an `http` version if the https version isn't available anyway. Thoughts @hamstu?


packages/utils/profileParser.js
```diff
@@ -1,6 +1,6 @@
 module.exports = profile => ({
   id: profile.id,
-  avatarUrl: profile.avatar,
+  avatarUrl: profile.avatar_https,
   avatar_https: profile.avatar_https,
```
djfarrelly: This was added in #130.

PR #155, opened 2017-09-06
Update queue reducer + get environment from server package into queue

alvarezmelissa87:
### Purpose
- create environment method in rpc to get environment into queue store 
- Pass in correct dynamic environment to ComposerWrapper.
- Ensure we are updating state correctly in queue reducer by passing actual state through to all reducers.
### Notes
Noticed some things to make follow-up tasks/PRs for:

- profile-sidebar count needs to update with post delete/create
- ensure `enabledApplicationModes` is in expected format in ComposerWrapper
- only have current profile set as 'selected' in MC

### Review
Heya @hamstu - would love your 👀 on this one when you get a chance! 😄 
cc @hharnisc for context 😄 
#### Staging Deployment


packages/queue/reducer.js
```diff
@@ -152,14 +186,101 @@ export default (state = initialState, action) => {
...
+            [profileId]: profileReducer(state.byProfileId[profileId], action),
+          },
+        };
+      }
+      return state;
     case actionTypes.POST_CANCELED_DELETE:
+      profileId = getProfileId(action);
+      if (profileId) {
+        return {
+          ...state,
```
emily-plummer: is this the big change here, just making sure to add byProfileId to the state instead of just returning byProfileId? Just want to make sure I'm understanding the shift!

alvarezmelissa87: Yeah, so basically the changes are making sure we are actually sending the state back to each reducer so that it's getting updated correctly instead of always getting overwritten by `profileInitialState` - which was happening at the `profileReducer` because we were never passing through the current state.

packages/queue/reducer.js
```diff
@@ -124,22 +145,41 @@ const profileReducer = (state = profileInitialState, action) => {
...
+        posts: postsReducer(state.posts, action),
+      };
     case actionTypes.POST_ERROR: {
-      let adjustTotal = 0;
-      if (action.type === actionTypes.POST_CREATED) {
-        adjustTotal = 1;
-      } else if (action.type === actionTypes.POST_DELETED) {
-        adjustTotal = -1;
-      }
       return {
```
alvarezmelissa87: Don't need this anymore as we're doing the total adjustment in POST_CREATED and POST_DELETED themselves

alvarezmelissa87:
Follow up task: ensure profile gets deselected once we switch profile tabs

packages/queue/middleware.js
```diff
@@ -8,6 +8,11 @@ import { actionTypes } from './reducer';
 export default ({ dispatch }) => next => (action) => { // eslint-disable-line no-unused-vars
   next(action);
   switch (action.type) {
+    case 'APP_INIT':
+      dispatch(dataFetchActions.fetch({
```
hamstu: I wonder if we can do this more quickly in the future; like by setting a variable when the code is bundled which would be different in production? Then we wouldn't need the latency of an API request to check for the environment. 

Could be a build step thing: https://webpack.js.org/guides/production/#specify-the-environment

hamstu: Not a big one right now though, just thinking of optimizations :)

packages/queue/reducer.js
hamstu: Curious why you made this change, as well as duplicating all the handler code for all the `actionTypes.*` above? The `case` statements can be stacked to do the same thing when you ommit `break;` (or `return;`) so this way just adds more and duplicated code for the same effect. 🤔

packages/queue/reducer.js
```diff
@@ -152,18 +192,118 @@ export default (state = initialState, action) => {
...
+      if (profileId) {
+        return {
+          ...state,
+          byProfileId: {
+            ...state.byProfileId,
+            [profileId]: profileReducer(state.byProfileId[profileId], action),
+          },
+        };
+      }
+      return state;
```
hamstu: Same comment as before; all this duplicated code can be removed in favor of the previous stacked `case` statements. 

Let me know if anything doesn't make sense, or if perhaps I'm missing some context. 😄

packages/queue/reducer.test.js
```diff
@@ -145,6 +161,9 @@ describe('reducer', () => {
           total: 1,
         },
       },
+      enabledApplicationModes: [],
+      showComposer: false,
+      environment: 'production',
```
hamstu: Keepin' them tests fresh! ✨

packages/queue/reducer.js
alvarezmelissa87: Heya @hamstu - you're totally right! I added these in for testing and never pushed the changes to remove them - totally my fault!
I'll go ahead and remove the duplicate code 👍

packages/queue/middleware.js
alvarezmelissa87: Heya @hamstu 😄  Great one! Yeah that might be a great way to optimize 🤔 I think this is a great one to make note of but feels like for the sake of moving forward now, I'll go ahead and keep this.

hamstu: @alvarezmelissa87 💯 Sounds great.

PR #162, opened 2017-09-11
reset composer store data on save success

alvarezmelissa87:
### Purpose
https://app.getflow.com/organizations/369191/teams/339227/tasks/25945410

Composer store should reset after successful save.
#### Staging Deployment


alvarezmelissa87:
closing because branch name is causing jenkins to error due to length.

PR #166, opened 2017-09-14
Profile sidebar loading ✨

hamstu:
### Purpose

Resolves [this task in Flow](https://app.getflow.com/organizations/369191/teams/339227/tasks/26146392) as well as adding some polish to the UI with loading animation. 💅 

newdash-profile-loading

#### Staging Deployment

### DIFF
--- a/packages/profile-sidebar/reducer.js
+++ b/packages/profile-sidebar/reducer.js
@@ -1,25 +1,30 @@
 import { actionTypes as dataFetchActionTypes } from '@bufferapp/async-data-fetch';
-import profiles from './mockData/profiles';
-import lockedProfiles from './mockData/lockedProfiles';
 
 export const actionTypes = {
   SELECT_PROFILE: 'SELECT_PROFILE',
 };
 
 const initialState = {
-  profiles,
-  lockedProfiles,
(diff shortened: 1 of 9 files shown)


packages/profile-sidebar/reducer.js
```diff
@@ -1,25 +1,30 @@
...
+        ...state,
+        loading: true,
+      };
     case `profiles_${dataFetchActionTypes.FETCH_SUCCESS}`: {
       return {
         ...state,
+        loading: false,
         profiles: action.result
-          .filter(profile => !profile.locked),
+          .filter(profile => !profile.disabled),
```
hamstu: 🔑 This ensures we sort profiles properly. (disabled is synonymous with locked and both come through the API but profileParser is only returning disabled.)

packages/server/index.html
```diff
@@ -30,6 +30,14 @@
       [data-hide-scrollbar]::-webkit-scrollbar {
         display: none;
       }
+      @keyframes placeHolderShimmer {
```
hamstu: I don't _love_ putting this on the global scope here, but it was the quickest way to effect the loading animation.

PR #168, opened 2017-09-15
Fix RPC credential and composer errors

hamstu:
### Purpose

This PR fixes the composer's RPC client so that it's API calls work. (Thanks for the help there @hharnisc!)

---

In addition I've added a few more things to this PR:

1. Errors from requests made via the `composerApiProxy` are correctly passed back to the composer, this ensures they display correctly in the UI. ✅ Fixes [this flow task.](https://app.getflow.com/organizations/369191/teams/339227/tasks/26247610)

2. Added `cache-loader` (only in dev) to increase the speed of webpack rebuilds when `nodemon` restarts the app (when changing files in packages/server/*) – I'm seeing a >50% speed increase from this change, which helped a lot with development and debugging.

3. Small tweak to add `--inspect=PORT` to the nodemon command in `Dockerfile.development`. This allows you to debug any server code in the Chrome DevTools! [Read here for more details](https://nodejs.org/en/docs/inspector/).  _Note: This won't work for you without exposing the right port in the `docker-compose.yml` file in `buffer-dev` — I will add that in to another PR/branch.._

4. Removed some unnecessary `package-lock.json` files.

#### Staging Deployment


Dockerfile.development
```diff
@@ -8,4 +8,4 @@ RUN npm install -g nodemon
 
 EXPOSE 80
 
-CMD nodemon --ignore 'index.html' --watch ./packages/server -e html,js,json packages/server/index.js
+CMD nodemon --inspect=0.0.0.0:5858 --ignore 'index.html' --watch ./packages/server -e html,js,json packages/server/index.js
```
djfarrelly: I wonder if it makes sense to put all of these options into a [`nodemon.json`](https://github.com/remy/nodemon#config-files) so devs can make more on the fly changes. If we keep them in the Dockerfile CMD it may be more work to override them. Just curious if opening the inspect port adds any overhead that someone may not want to do development with all the time 🤔  cc @hamstu @hharnisc

hamstu: Great idea @djfarrelly on using a config file. 
I'm not certain of overhead, but I can certainly see it being possible! 

One thing I _have_ noticed is the server will fail on the first nodemon restart since the inspect port is 'taken', which suggests it's not being released fast enough. 

Will investigate both these together. Thanks for the feedback! 🙌

PR #171, opened 2017-09-19
reset data before mounting

emily-plummer:
### Notes
This PR moves the resetting of the composer data to happen before a composer mounts instead of right before it unmounts. This is because `draft.isSaved` gets set asynchronously and sometimes can happen after the data gets reset, leading to the edit composer not being openable :)

### Review
would love a quick look @pioul since you were working on this last week!
#### Staging Deployment


pioul:
Awesome find @emily-plummer, and cool simple solution! Quick question though: have you noticed any change in speed when opening the composer? The reason this reset call was initially set in `componentWillUnmount` is that there's quite a bit going on in there, so we might as well do it at a time where speed is less critical. I personally haven't noticed any change in perf in Publish, just wanted to mention this so that we keep it in mind. If we wanted to speed things up, we could put the reset call back in `componentWillUnmount` and make sure that the actions you've [mentioned](https://app.getflow.com/organizations/369191/teams/339227/tasks/26247609#) run in the expected order :)

PR #174, opened 2017-09-19
1️⃣ ➡️ 2️⃣ Update post counts

hamstu:
### Purpose

Ensure the post counts are up-to-date throughout the app.

Also fixes handling of `POST_SENT` action in the sent reducer so posts show up there.

### Notes

This PR causes an action to be dispatched whenever the post counts change. Any part of the app that cares can subscribe and update accordingly.

While this does mean that the count information is stored in multiple places (the queue and sent reducers `total` property on their profiles, as well as the `profiles` arrays in the `profile-sidebar`) I think that it ends up being cleaner and easier than using react-redux to `connect()` components so they can reach deep into the state for another package.


hamstu:
Haha, no worries @emily-plummer! Thanks for the review! 😄

PR #180, opened 2017-09-21
Share now success notification ✅ 🗯

hamstu:
### Purpose

Show a success notification when your posts are successfully 'Share Now'd.

Also includes some minor tweaks to the design of the notifications; adjusted spacing and font size to [better match the mockups](https://buffer.wake.com/set/3/31)

cc @hharnisc Wake made the notification look bigger because of the @2x source image.

### DIFF
--- a/packages/components/Notification/index.jsx
+++ b/packages/components/Notification/index.jsx
@@ -45,7 +45,7 @@ const Notification = ({
     >
       <div
         style={{
-          margin: '1.5rem',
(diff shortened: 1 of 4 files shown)


hamstu:
That's a great question Emily! Might not work on Linux (emoji support there is spotty), but other than that it should appear fine. I'll have to take a look on my Windows machine at some point though, as their style might not jive with ours so much, and could feel off.

PR #186, opened 2017-09-26
Drag and Drop in the Queue 🐲 (Front-end)

hamstu:
### Purpose

PR for Drag and Drop – front end only. (Does not save to back-end or sync.)

### Notes

I've decided to go with the [`react-dnd`](http://react-dnd.github.io/react-dnd/) library from Dan Abramov of facebook / redux fame. _This isn't accessible right now, but we will work on this later._ 💯 

### Review

Can you drag and drop? 😄


KatieLo:
👀

hharnisc:
Hey @hamstu great choice on the library :+1: I'm confident we can move forward with this library and reuse it for any other DnD stuff we need to build!

A note on ES7 decorators (which are basically just wrappers), there is an ES6 example (hidden in a tab) next to the ES7 example.

hamstu:
@hharnisc Thanks for weighing in Harrison! Yep, I decided to use the ES6 style wrapping instead. 👍

emily-plummer:
Heya @hamstu! 

This is looking great! My only concern is (of course 👀 ) accessibility. It looks like the library doesn't have any accessible features or plans to implement accessibility in the future https://github.com/react-dnd/react-dnd/issues/382

I'm looking into ways we might be able to make it accessible ourselves, or perhaps submit a PR to react-dnd to make it accessible. 

That being said, I'm not sure how well maintained the library is. In the issue above they mention that the library is not actively developed anymore. It looks like there's been a few releases over the last 3 weeks & then one in July & one in February. Might be something that we'd want to consider! 

I've found another library that Atlassian has open sourced which builds off the ideas in react-dnd but is fully accessible! They mention that it's not as robust as react-dnd but it looks like it should cover all the use cases we'd need! :) 

https://github.com/atlassian/react-beautiful-dnd

hamstu:
@alvarezmelissa87 Would love your eyes on this PR! 🙏

packages/queue/components/QueuedPosts/index.jsx
```diff
@@ -1,20 +1,18 @@
...
   LoadingAnimation,
 } from '@bufferapp/components';
 import {
-  geyser,
-  fillColor,
-  curiousBlue,
-} from '@bufferapp/components/style/color';
-import {
-  PostLists,
+  // PostLists,
```
alvarezmelissa87: Maybe we can go ahead and remove this commented bit if we're not using PostLists? 😄

packages/queue/index.js
```diff
@@ -5,22 +5,16 @@ import QueuedPosts from './components/QueuedPosts';
...
-  const postIds = Object.keys(posts);
-
-  postIds.forEach((postId) => {
-    if (posts[postId].day !== day) {
-      day = posts[postId].day;
-      newList = { listHeader: day, posts: [posts[postId]] };
-      postLists.push(newList);
-    } else { // if same day add to posts array of current list
-      newList.posts.push(posts[postId]);
+  const orderedPosts = Object.values(posts).sort((a, b) => a.due_at - b.due_at);
```
alvarezmelissa87: nice! 🙌

packages/queue/reducer.js
```diff
@@ -56,6 +57,16 @@ const getPostUpdateId = (action) => {
...
 
+const movePost = (arr, from, to) => {
+  const clone = [...arr];
+  const fromIndex = from < to ? from : to;
+  const toIndex = to > from ? to : from;
+  Array.prototype.splice.call(clone, toIndex, 0,
+    Array.prototype.splice.call(clone, fromIndex, 1)[0],
+  );
+  return clone;
+};
```
alvarezmelissa87: `movePost` is a bit difficult to follow here - maybe adding a comment would be helpful? 🤔

packages/queue/reducer.js
```diff
@@ -179,6 +190,35 @@ const profileReducer = (state = profileInitialState, action) => {
...
+        (map, post) => {
+          map[post.id] = post; return map;
+        },
+        {},
+      );
+      return {
+        ...state,
+        posts: newPostsMap,
+      };
+    }
```
alvarezmelissa87: Might be worth pulling this logic into a utils file just so the reducer is a bit easier to read 😄

packages/queue/components/QueuedPosts/index.jsx
hamstu: Great catch! Will do 😄

packages/queue/reducer.js
hamstu: Great idea! Thanks for that insight. Will be helpful for anyone (myself included, haha) looking at this code down the road.

PR #192, opened 2017-10-02
Update to React 16!

emily-plummer:
### Purpose
Upgrading our codebase to React 16! Just needed to change a bunch of React.PropTypes calls to use the prop-types package & update a few of our dependencies :) 

### Notes
With the new enzyme 3.0, we'll need to run wrapper.update() to receive props changes!

### Review
Would love a review @alvarezmelissa87 @hharnisc !
#### Staging Deployment


emily-plummer:
Open issues to monitor:
Draft.js Plugins: https://github.com/draft-js-plugins/draft-js-plugins/issues/906
React Tooltip: https://github.com/wwayne/react-tooltip/issues/342

packages/composer/package.json
```diff
@@ -51,11 +52,12 @@
...
     "moment": "2.16.0",
     "moment-timezone": "0.5.12",
-    "react-day-picker": "3.1.1",
-    "react-dropzone": "3.7.3",
-    "react-simple-dropdown": "1.1.4",
-    "react-textarea-autosize": "4.0.5",
-    "react-tooltip": "3.2.0",
+    "react-day-picker": "6.1.1",
+    "react-test-renderer": "^16.0.0",
+    "react-dropzone": "4.1.3",
```
emily-plummer: Can we moved some of these shared dependencies that are in compose & in the root package.json out? Still wondering if there's anything separate between how the packages for the composer work versus how the packages for any other our packages work?

packages/components/Button/test.jsx
```diff
@@ -16,10 +19,12 @@ describe('Button', () => {
     const button = wrapper
       .find(ButtonStateless);
     button.simulate('mouseEnter');
-    expect(button.props().hovered)
+    wrapper.update();
+    expect(wrapper.find(ButtonStateless).props().hovered)
```
emily-plummer: Heya @hharnisc @alvarezmelissa87 @hamstu @msanroman heya! just wanted to give y'all a heads up that with the new enzyme 3.0, we'll need to run wrapper.update() to receive props changes. I also need to check the props on the ButtonStateless component directly instead of checking the props of the `button` instance we've created on line 19.

packages/composer/package.json
hamstu: Yep, you can keep these in the root `package.json`, especially if they're already there. When we run `yarn run bootstrap` it triggers a `lerna bootstrap --hoist` which pulls all the npm deps into the root `node_modules` so they are available to all packages. [The docs explain it really well here](https://github.com/lerna/lerna/blob/master/doc/hoist.md).

PR #196, opened 2017-10-11
Fixing webpack config error

kiriappeee:
### Purpose
Fixing up an error in webpack bundling that showed up yesterday at about 10 AM UTC. 

### Notes
Full background details are here -> https://paper.dropbox.com/doc/Plan-for-webpack-config-error-of-Buffer-Publish-gRibh8o1jnrPGMFMMJEGl#:uid=539565976936751064627124&h2=Background:

### Review
While this works and things seem to be working fine, I'm not well versed in the nuances of `webpack` bundling and would love to get some eyes on this to see if I have broken anything in the background. 

#### Staging Deployment

https://task-fix-webpack-config-error-publish.dev.buffer.com

### DIFF
--- a/packages/server/webpack.config.js
+++ b/packages/server/webpack.config.js
@@ -1,5 +1,3 @@
-
-var webpack=require('webpack');
 module.exports = {
   context: __dirname,
   entry: [
--- a/packages/server/webpack.config.js
+++ b/packages/server/webpack.config.js
@@ -10,11 +10,12 @@ module.exports = {
     filename: 'bundle.js',
     publicPath: '/static/',
   },
-  plugins: [
-     new webpack.IgnorePlugin(/\.\/locale$/)
-  ],
+  plugins: [],
   resolve: {
     extensions: ['.js', '.json', '.jsx'],
+    alias: {
+      moment$: 'moment/moment.js',
+    },
   },
   module: {
     rules: [
--- a/packages/server/webpack.config.js
+++ b/packages/server/webpack.config.js
@@ -1,3 +1,5 @@
+
+var webpack=require('webpack');
 module.exports = {
   context: __dirname,
   entry: [
@@ -8,7 +10,9 @@ module.exports = {
     filename: 'bundle.js',
     publicPath: '/static/',
   },
-  plugins: [],
+  plugins: [
+     new webpack.IgnorePlugin(/\.\/locale$/)
+  ],
   resolve: {
     extensions: ['.js', '.json', '.jsx'],
   },


emily-plummer:
Heya @kiriappeee ran this on local and it seems to work fine! :D Thanks for this quick fix!

packages/server/webpack.config.js
```diff
@@ -1,3 +1,5 @@
+
+var webpack=require('webpack');
 module.exports = {
```
alvarezmelissa87: Heya @kiriappeee - curious why we need to require webpack here?

packages/server/webpack.config.js
```diff
@@ -11,6 +13,9 @@ module.exports = {
   plugins: [],
   resolve: {
     extensions: ['.js', '.json', '.jsx'],
+    alias: {
+      moment$: 'moment/moment.js',
+    },
   },
```
alvarezmelissa87: So aliasing this should match the import moment from 'moment' exactly and then cause locale importing statements like to be ignored?

Looks like this should work but I'm concerned as to why we suddenly started seeing these issues. 🤔 
My hunch would be to move forward since it's blocking people but to make a note and investigate further as to why this started happening.

Looping in @hamstu - as he has some context with webpack. 😄

packages/server/webpack.config.js
bufferbot: @alvarezmelissa87 Oh that's a good one. That's from the first commit I made. It's not needed anymore. I had a line previously in plugins which stated:

`new webpack.IgnorePlugin(/\.\/locale$/)` and that required the webpack import.

kiriappeee: Oh that's a good one @alvarezmelissa87 . It's from the first commit where I had a line in plugins that looked like: 

`new webpack.IgnorePlugin(/\.\/locale$/)`

So the require statement can be removed now 👍 . Thanks so much for the spot

packages/server/webpack.config.js
kiriappeee: @alvarezmelissa87 yeps. This comment words what's happening much better than I can 😁 (And it's a pretty much 1:1 match with what you said)

https://github.com/ant-design/ant-design/issues/4491#issuecomment-270887624

>Looks like this should work but I'm concerned as to why we suddenly started seeing these issues. 🤔
>My hunch would be to move forward since it's blocking people but to make a note and investigate further as to why this started happening.

💯 % agree (if there's a percent emoji, I'm sorry. my emoji game isn't strong 🙈 😅 )

kiriappeee:
Since this seems to work, I'm merging it in :). Shall work with the team to figure out what happened (might not be able to help much here), and see if I can change the pipeline in any way to be more resilient to external changes.

packages/server/webpack.config.js
hamstu: This is a great fix! Thanks for digging in you @alvarezmelissa87 and @kiriappeee  🙏 

This appears to be an issue with the latest version of `moment`, you can see more reports in this GH issue: https://github.com/moment/moment/issues/2979. I'd say let's leave it for now, and hopefully someone will patch it in the next version and we can remove this fix.

PR #200, opened 2017-10-24
Task/failed posts

emily-plummer:
### Purpose
- Shows failure messages inline when a post fails in the queue
- Creates the new requeue RPC method to handle requeueing posts
- Allows users to Retry Now 💞 

### Notes
We've uncovered some weirdness with the requeue endpoint which Colin is fixing up for us here: https://github.com/bufferapp/buffer-web/pull/14169 🌟 

#### Staging Deployment


packages/server/rpc/requeuePost/index.js
```diff
@@ -0,0 +1,28 @@
...
+        },
+      });
+    } catch (err) {
+      if (err.error) {
+        const { message } = JSON.parse(err.error);
+        throw new Error(message);
+      }
+      throw err;
+    }
+    return 'OK';
```
emily-plummer: We don't need to do anything with the data that gets passed back here, since we need to trigger a call to /pending in order to get all the posts back and order them in the queue (sometimes when requeueing one post, other posts also need to shift to accommodate that post, so we need to refresh the posts each time 😄 )

PR #204, opened 2017-10-27
⏯ Pause Button

hamstu:
### Purpose

This PR adds pause button functionality to Publish that is similar to the UI in buffer-web, with some tweaks to be more publish-y.

**Message on the queue:**
image

**On the settings page:**
image

### Video

✨ 🎥 See a video here: https://cl.ly/0R2U3B141R1f

#### Staging Deployment

### DIFF
--- a/packages/profile-sidebar/reducer.js
+++ b/packages/profile-sidebar/reducer.js
@@ -3,6 +3,9 @@ import { actionTypes as dataFetchActionTypes } from '@bufferapp/async-data-fetch
 export const actionTypes = {
   SELECT_PROFILE: 'SELECT_PROFILE',
   POST_COUNT_UPDATED: 'POST_COUNT_UPDATED',
+  PROFILE_UNPAUSED: 'PROFILE_UNPAUSED',
+  PROFILE_PAUSED: 'PROFILE_PAUSED',
+  PUSHER_PROFILE_PAUSED_STATE: 'PUSHER_PROFILE_PAUSED_STATE',
 };
 
 const initialState = {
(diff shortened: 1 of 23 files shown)


packages/profile-sidebar/reducer.js
```diff
@@ -31,6 +34,23 @@ const profilesReducer = (state = [], action) => {
             ? action.counts.sent
             : profile.sentCount,
         }));
+    case actionTypes.PROFILE_PAUSED:
+    case actionTypes.PROFILE_UNPAUSED:
+      return state
+        .map(profile => ({
```
emily-plummer: Beautiful use of all the ES6 goodies! 🙌

packages/server/rpc/pauseQueue/index.js
```diff
@@ -0,0 +1,30 @@
...
+  async ({ profileId, paused }, { session }) => {
+    let result;
+    try {
+      result = await rp({
+        uri: `${process.env.API_ADDR}/1/profiles/${profileId}/paused.json`,
+        method: 'POST',
+        strictSSL: !(process.env.NODE_ENV === 'development'),
+        qs: {
+          access_token: session.accessToken,
+          paused: paused ? '1' : '0',
```
emily-plummer: Are we using 1/0 here instead of true/false because this gets evaluated to a string anyways?

packages/queue/components/QueuedPosts/index.jsx
```diff
@@ -68,6 +71,7 @@ const QueuedPosts = ({
           onFocus={onComposerPlaceholderClick}
         />
       </div>
+      {!!paused && <QueuePausedBar handleClickUnpause={onUnpauseClick} />}
```
emily-plummer: Curious why you opted to use a !! here? Is it possible for paused to be a non truthy/falsy value?

PR #211, opened 2017-11-15
Changed link from uservoice to typeform

suprasanna:
### Purpose
Changed link from Uservoice (which we no longer use) to Typeform for the Wishlist link.

### Notes
Though I changed the preferences link for now just to get it off of Uservoice, we'll change that eventually too to be actually going to Preferences. Left it for now.

#### Staging Deployment


hamstu:
This is great @suprasanna!

The build is failing on this branch since the update means our 'test snapshots' don't match. So before this can be merged & deployed you'll need to run this in the terminal.

```bash
# in the buffer-publish folder...
$ yarn run test-update
```

That should update the snapshots

Then you can commit and push again here and wait to see if the _Travis_ job is green when it completes:

(As you can see it's red right now 👇 )
image

If any of this gives you trouble I'm happy to step in! Just letting you know the steps so you can deploy all yourself in the future. 🚀

PR #220, opened 2017-11-30
Add unauthorized redirect

hharnisc:
### Purpose

When a user's session expires (or they accidentally delete cookies) the user is redirected when they try to do something that requires authorization.

### Notes

Authorization status is determined by `checkToken`: https://github.com/bufferapp/buffer-publish/blob/master/packages/server/rpc/checkToken.js

### Review

I've moved the environment middleware over to its own package, we'll also use this package in analyze. Does the approach make sense for other types of shared data (like users)?

#### Staging Deployment

https://task-add-unauthorized-redirect-publish.dev.buffer.com :smile:_


hharnisc:
looks like this is passing travis :D gonna merge it :green_apple:

hamstu:
I like this a lot @hharnisc! 

> Does the approach make sense for other types of shared data (like users)?

I think that's a great approach for shared data like this. 💯

packages/app-sidebar/index.js
```diff
@@ -7,7 +7,7 @@ export default connect(
   state => ({
     translations: state.i18n.translations.example, // all package translations
     user: state.appSidebar.user,
-    environment: state.appSidebar.environment,
+    environment: state.environment.environment,
```
emily-plummer: nioooce 👍

packages/server/index.js
```diff
@@ -68,18 +68,18 @@ app.use(sessionMiddleware.getSession({
...
-  requiredSessionKeys: ['publish.accessToken', 'global.userId'],
-}));
-
 app.post('/rpc', (req, res, next) => {
   rpc(req, res)
     // catch any unexpected errors
     .catch(err => next(err));
 });
 
+// make sure we have a valid session
```
emily-plummer: curious if there's a reasoning for moving this one down? Just getting started learning more about express 😸

PR #221, opened 2017-12-02
Fix redirect in production

hharnisc:
### Purpose

Redirect works in production environment now

### Notes

Only worked locally before!

#### Staging Deployment

https://task-fix-redirect-production-publish.dev.buffer.com :smile:_


hharnisc:
Verified the dev build is working as expected, merging this!

PR #223, opened 2017-12-11
✨ Better Dev Environment 🙏

hamstu:
### Purpose

We've had a number of issues with buffer-publish (most of which are related to running webpack inside the Docker container); specifically:

1. 👀  Webpack would frequently stop picking up file changes, requiring a **painfully slow** restart of the entire publish container to get things working again. The precise cause here is unknown but this is likely a Docker issue specific to macOS.

2. ♻️ Any changes to `packages/server/**` would cause `nodemon` to restart the server (which is great and expected) but would _also_ unnecessarily restart the webpack bundling (since that was a child process of the node server). This meant **waiting upwards of 30-45 seconds before you could view the page again.**

3. 🔥 HMR was not working correctly (except for a very limited set of components) nor were reducers hot reloaded. This meant more manual refreshes. 😢 

---

👉  **This PR fixes these issues by separating the webpack bundling from the express server.**

The only caveat is that you must now run (on your system; outside the container):

```bash
$ yarn run start
```

In `buffer-publish` to start the webpack watching.

You can then browse to https://publish.local.buffer.com as usual.

### Notes

* [📖  **Read the migration guide to test this on your system**](https://paper.dropbox.com/doc/Better-Dev-Environment-for-Buffer-Publish-Migration-Guide-ymaH6fZdn2wSVi4a1y2pi)

* Added a `console.error` to failed RPC errors so we can see what went wrong more easily. (Otherwise all errors—even JavaScript errors—were being swallowed by the RPC middleware).
* Also removed the node debugging switch on the server (`--inspect`), which was causing errors when server files changed and nodemon would fail to restart.

### Review

❓ Does this work for you locally?

### DIFF
--- a/packages/server/webpack.config.dev.js
+++ b/packages/server/webpack.config.dev.js
@@ -1,5 +1,6 @@
 const webpack = require('webpack');
 const config = require('./webpack.config');
+const fs = require('fs');
 
 // NOTE: Bugsnag will not notify in local setup with current weback configuration
 // https://docs.bugsnag.com/platforms/browsers/faq/#4-code-generated-with-eval-e-g-from-webpack
@@ -7,13 +8,25 @@ config.devtool = 'cheap-module-eval-source-map';
 
 config.entry.unshift(
   'react-hot-loader/patch',
-  'webpack-hot-middleware/client',
+  // 'webpack-hot-middleware/client',
 );
+
+config.devServer = {
+  hot: true,
+  publicPath: config.output.publicPath,
+  contentBase: false,
+  port: 8080,
+  host: 'local.buffer.com',
+  headers: { 'Access-Control-Allow-Origin': '*' },
+  https: {
+    key: fs.readFileSync('../reverseproxy/certs/local.buffer.com-wildcard.key'),
+    cert: fs.readFileSync('../reverseproxy/certs/local.buffer.com-wildcard.crt'),
+  },
+};
 
 config.plugins.unshift(
   new webpack.HotModuleReplacementPlugin(),
+  new webpack.NamedModulesPlugin(),
 );
-
-config.module.rules[1].loaders.unshift('cache-loader');
 
 module.exports = config;
--- a/packages/web/index.jsx
+++ b/packages/web/index.jsx
@@ -31,11 +31,40 @@ const renderApp = (AppComponent) => {
 
 renderApp(App);
 
+/**
+ * We're using HMR (Hot Module Reloading) here to ensure that the components update
+ * and you don't need to reload the page manually.
+ *
+ * Because of the way we have packages setup (suing lerna), we need to specifically
+ * include them in the call to `module.hot.accept()`.
+ */
 if (module.hot) {
-  console.log('module is HOT! 🔥');
-  module.hot.accept('./components/App', () => {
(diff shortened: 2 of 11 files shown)


hamstu:
cc @msanroman and @federicoweber – not sure if you've had many issues, but implementing something like this in analyze may be helpful! Not up to date on your setup over there, so let me know if I can help at all.

packages/server/webpack.config.dev.js
```diff
@@ -1,19 +1,32 @@
...
-  'webpack-hot-middleware/client',
+  // 'webpack-hot-middleware/client',
 );
 
+config.devServer = {
+  hot: true,
+  publicPath: config.output.publicPath,
+  contentBase: false,
+  port: 8080,
+  host: 'local.buffer.com',
```
msanroman: Curious about the choice of using `local.buffer.com`, is it just for being able to reuse the certs we have there? :)

packages/web/index.jsx
```diff
@@ -31,9 +31,40 @@ const renderApp = (AppComponent) => {
...
+ * We're using HMR (Hot Module Reloading) here to ensure that the components update
+ * and you don't need to reload the page manually.
+ *
+ * Because of the way we have packages setup (suing lerna), we need to specifically
+ * include them in the call to `module.hot.accept()`.
+ */
 if (module.hot) {
-  module.hot.accept('./components/App', () => {
+  module.hot.accept([
+    './components/App',
```
djfarrelly: Could we use something like `fs.readdirSync(pathToPackagesDir)` to grab all of the directories and just add `index.js` and `reducer.js` to all of them?

packages/server/webpack.config.dev.js
hamstu: Yeah that's it. I did a similar setup for the salary calculator. I think it helps having everything being served from the same domain, even if the ports are different (to avoid any cross-origin weirdness with webpack-dev-server). It _could_ work on localhost, so might be worth testing. Pretty sure I did this for more than just cosmetic reasons on the salary calculator :)

packages/web/index.jsx
hamstu: This code has to run in the web context; so unfortunately that wouldn't work. (Could be implemented with a bundle-time webpack plugin, maybe?)

But yeah, I don't love this, but it's the only way that got true HMR working as far as I could figure it out.

hamstu: Some further research suggests that the first argument to `module.hot.accept()` [has to be statically analyzable](https://github.com/webpack/webpack/issues/5668#issuecomment-329118902). Which I recall being true in some previous testing where I tried to generate this array from an array of strings.

hamstu: 🤔  Still researching here; I think there's a good chance we won't need all these paths. For example, I started working on a new package based on this new dev setup and **I did not have to add it to this list** for HMR to work correctly. I think the way the HMR changes are bubbling up and getting accepted is a little funky, but I don't know enough about the internals to know what's happening.

I'll keep an eye on things as we continue to build out our packages.

msanroman: Testing things on Analyze, it seems to be working well by just requiring `components/App`, `store/index` and `store/reducers`!

I think any component will be propagated through `App` anyway 🤔

PR #227, opened 2017-12-18
Connect social accounts click handler added

alvarezmelissa87:
### Purpose
Link 'Connect Social Account' button to https://buffer.com/app/profile/connect
Related Flow task: https://app.getflow.com/organizations/369191/teams/339227/tasks/28351363
### Notes
Looking into having it redirect to publish after connection is done. 
#### Staging Deployment


packages/profile-sidebar/middleware.js
```diff
@@ -13,6 +13,19 @@ import {
...
+  if (window.location.hostname === 'publish.local.buffer.com') {
+    return 'https://local.buffer.com/app/profile/connect';
+  }
+  return 'https://buffer.com/app/profile/connect';
+};
+
+// TODO: ensure this works in all browsers
+const openInNewTab = (url) => {
+  const win = window.open(url, '_blank');
+  win.focus();
```
hamstu: I don't _think_ you need `win.focus()`, since it will focus automatically. 

Regardless I don't think we actually want this in a new tab, because of the redirect that will happen once you connect, you'll just end up with a duplicate tab. So I'd go with just changing `window.location`.

alvarezmelissa87: Ah good one, @hamstu - part of the acceptance criteria was to have it in a new tab _but_ I think as you say with the redirect changing the current tab should be good 👍 
I'll add a note in Flow for @suprasanna

alvarezmelissa87:
Heya @hamstu - I've updated with the suggested changes! Would love to get your 👀 on this for those when you get a chance! 😄 Thank you!

packages/profile-sidebar/middleware.js
hamstu: Oh great one! Yeah, I think @suprasanna would agree with same tab as that's the functionality everywhere else. Thanks Melissa!

PR #228, opened 2017-12-19
Task/upgrade cta

alvarezmelissa87:
### Purpose
Adds top header `Upgrade` link to `buffer.com/pricing`

Related Flow task: https://app.getflow.com/organizations/369191/teams/339227/tasks/28351367
### Review
@hamstu, @emily-plummer - would love your thoughts on the update to `Tabs` to be able to handle children that can just be links. 
My thinking was that in the future we might need to add more to the top navigation that won't necessarily act exactly as the other tabs.
Maybe if this feels okay we can update `Tabs` to `Navigation` or something since the first implies children will be tabs. 🤔 
#### Staging Deployment


packages/tabs/components/TabNavigation/index.jsx
```diff
@@ -4,6 +4,14 @@ import {
   Tabs,
   Tab,
 } from '@bufferapp/publish-shared-components';
+import { Link } from '@bufferapp/components';
+
+const getClassicBufferPricingURL = () => {
```
hamstu: Might be time for a utils method to help with these URLs. I've had to use a method like this a few times too. 🤔

packages/shared-components/Tabs/index.jsx
```diff
@@ -10,10 +10,13 @@ const Tabs = ({
   onTabClick,
 }) => (
   <div>
-    {React.Children.map(children, tab => React.cloneElement(tab, {
-      selected: selectedTabId === tab.props.tabId,
-      onClick: onTabClick,
-    }))}
+    {React.Children.map(children, (tab) => {
+      if (!tab.props.tabId) return tab;
```
hamstu: Great escape hatch for non-tabs! I agree with your idea on generalizing the container to become `Navigation`, but for now I think this will work just fine. 👍

emily-plummer: Agree with you and Hamish here - renaming it to something more general like Navigation would feel nice :) Would be great if we could throw in some `<nav>` tags too to make it semantic & more accessible :)

PR #234, opened 2017-12-22
Task/add times to paused days

alvarezmelissa87:
### Purpose
- Updates addTime util to add time to both paused and unpaused days.
- Use `updatePausedSchedule` rpc method since it handles updates to both schedules (needed because when times are added to multiple days some of those days may be paused)

Related Flow card: https://app.getflow.com/organizations/369191/teams/339227/tasks/28510811
#### Staging Deployment


packages/settings/middleware.js
```diff
@@ -53,9 +53,10 @@ export default ({ dispatch, getState }) => next => (action) => {
       break;
     case actionTypes.ADD_SCHEDULE_TIME:
       dispatch(dataFetchActions.fetch({
-        name: 'updateSchedule',
+        name: 'updatePausedSchedules',
```
emily-plummer: Smart! This works because updating a paused schedule always updates the schedule with the latest right? 🙌

packages/settings/utils/scheduleUtils.js
```diff
@@ -153,22 +158,35 @@ const addTimeToScheduleForApi = (unformattedSchedule, action) => {
...
+  const updatedSchedules = {};
 
   daysToAdd.forEach((dayToAdd) => {
-    unformattedSchedule.forEach((daySchedule) => {
-      if (daySchedule.days.includes(dayToAdd)) {
-        daySchedule.times.push(newTime);
-        daySchedule.times.sort();
-      }
-    });
+    if (pausedDays.includes(dayToAdd)) {
```
emily-plummer: I think this mutates pausedSchedules & schedules, we might want to cloneDeep those parameters before we pass them in

PR #240, opened 2018-01-04
fix issue with clearing all

emily-plummer:
### Purpose
Fixes an issue where after clearing all times, trying to add a new time to a paused day throws an error.

### Review
@alvarezmelissa87 Could you take a quick look at this?

#### Staging Deployment


packages/settings/middleware.js
```diff
@@ -82,13 +82,6 @@ export default ({ dispatch, getState }) => next => (action) => {
       }));
       break;
     case actionTypes.CLEAR_ALL_TIMES:
-      dispatch(dataFetchActions.fetch({
```
emily-plummer: Since we're updating the schedules in updatePausedSchedules we don't need to also call this fetch

packages/settings/reducer.js
```diff
@@ -158,7 +158,7 @@ export default (state = initialState, action) => {
     case actionTypes.CLEAR_ALL_TIMES:
       pausedSchedules = deleteAllTimesFromSchedule(state.pausedSchedules);
       schedules = deleteAllTimesFromSchedule(state.schedules);
-      mergedSchedules = [];
+      mergedSchedules = mergeSchedules(schedules, pausedSchedules);
```
emily-plummer: Not sure why we set this to an empty array to begin with @alvarezmelissa87 but I vaguely remember discussing it. I believe having this set as an empty array is causing the issues with the pausedSchedules not being set when we want to add a time after clearing all.

alvarezmelissa87: Heya @emily-plummer - in this case, since both pausedSchedules and schedules should be the same - is there a reason we're doing the mergeSchedules here?

emily-plummer: Heya @alvarezmelissa87 good point! I think we can just assign margedSchedules to either `schedules` or `pausedSchedules` since they should both be arrays with every day and no times :)

alvarezmelissa87: Yeah, I think we can set it to schedules since all the times will be cleared out. 👍 Looks good! 😄

PR #247, opened 2018-01-17
bump session manager version

hharnisc:
### Purpose

New session manage handles versioned (and non-versioned) sessions!

### Notes

Versioned sessions are useful when we need to migrate session store

### Review

Mostly putting this on your radar :smile:


hharnisc:
merging since :green_apple:

PR #248, opened 2018-01-17
update lerna to fix broken publish

hharnisc:
### Purpose

Get the `lerna publish` command working again

### Notes

2.7.0 has a bug that breaks package publish, 2.7.1 should fix it: https://github.com/lerna/lerna/issues/1200

### Review

FYI


hamstu:
Looks great @hharnisc! Thanks for catching this one! 🌳

PR #249, opened 2018-01-17
Use legacy watch mode with nodemon

hamstu:
### Purpose

Recently `nodemon` has stopped restarting the server when things change. I suspect this is due to changes in Docker. As such I've switched nodemon to use the 'legacy' watching mode. This mode is slower as it uses polling, but that should make little to no difference in our case as we're only watching ~30 files in the server package.

### Notes

Hey @emily-plummer, sounds like this was happening to you too. Would love for you to test this out. I'll merge it into master and you'll just have to run `./dev rebuild publish`.

cc @alvarezmelissa87 as well.

#### Staging Deployment

### DIFF
--- a/Dockerfile.development
+++ b/Dockerfile.development
@@ -8,5 +8,5 @@ RUN npm install -g nodemon
 
 EXPOSE 80
 
-CMD nodemon --ignore 'index.html' --watch ./packages/server -e html,js,json packages/server/index.js
+CMD nodemon -V -L --ignore 'index.html' --watch ./packages/server -e html,js,json packages/server/index.js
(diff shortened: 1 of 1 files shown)


Dockerfile.development
```diff
@@ -8,5 +8,5 @@ RUN npm install -g nodemon
 
 EXPOSE 80
 
-CMD nodemon --ignore 'index.html' --watch ./packages/server -e html,js,json packages/server/index.js
+CMD nodemon -V -L --ignore 'index.html' --watch ./packages/server -e html,js,json packages/server/index.js
```
hamstu: -V = verbose
-L = legacy watch

PR #252, opened 2018-01-19
Task/add draft package base

alvarezmelissa87:
### Purpose
Adds basics of the `drafts` package. 
### Notes
Additional functionality to come in subsequent PR's in order to keep PR sizes manageable.
### Review
Hey @emily-plummer, @hamstu - would love your 👀  on this.
My plan is to merge this in as a first piece, then merge in additional functionality to keep PR's small. Since I'm not using this package anywhere yet, it should be safe to merge in and continue development on it that way.

Would love to hear if this feels okay. Thanks!
#### Staging Deployment


packages/drafts/index.js
```diff
@@ -0,0 +1,141 @@
...
+      isPastDue,
+      twentyFourHourTime,
+    }),
+    isRetweet: draft.retweet !== undefined,
+  };
+};
+
+// Could export this to utils and then pull it in and pass tab depending on which package uses it
+const formatPostLists = (profile, drafts, user) => {
+  const profileIsManager = profile.organizationRole === 1;
```
hamstu: I'd love to minimize the number of times we do `role == <someRoleNumber>` checks in the code. Ideally we'd have this data stored on the profile/user, maybe it could be added as part of the parser(s)?

Ideally we do only do it once in the front-end and then don't have to have the 'magic numbers' scattered about. 😄  (This is a real problem in `buffer-web`) This could also be something we consider adding on the API end, but again the same idea of 'single source of truth' applies.

packages/drafts/reducer.js
```diff
@@ -0,0 +1,170 @@
...
+      return state;
+  }
+};
+
+const profileReducer = (state = profileInitialState, action) => {
+  switch (action.type) {
+    case `draftPosts_${dataFetchActionTypes.FETCH_START}`:
+      return {
+        ...state,
+        loading: true, // !action.args.isFetchingMore,
```
hamstu: Curious about the comment on this line. 🤔

packages/utils/profileParser.js
```diff
@@ -4,6 +4,7 @@ module.exports = profile => ({
   avatar_https: profile.avatar_https,
   type: profile.service,
   handle: profile.service_username,
+  isManager: profile.organization_role === 1,
```
hamstu: 😮  Oh awesome! You already added this here. Would be great to refactor the earlier check to use this value then. ☝️ 

Also, while I'm thinking about this; I think we should have a `constants` package, which would be single file. In that file we'd have constants for things just like this. Then it's not the magic number `1`, it's `ROLE_MANAGER` which is equal to `1`. Does that make sense? We don't have a lot of constants right now, really, but we will certainly gain more over time.

packages/drafts/components/DraftList/index.jsx
```diff
@@ -0,0 +1,140 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+// import CSSTransitionGroup from 'react-addons-css-transition-group';
```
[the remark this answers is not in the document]
hamstu: Coming back to this, I'm guessing you're still thinking about/working on migrating the CSS animations?

packages/drafts/index.js
alvarezmelissa87: Oh good catch @hamstu! 😄  I added an explicit `isManager` field to the profile in the profileParser that way we can just call it directly and only do the changes there. Updating this now! 👍

packages/utils/profileParser.js
alvarezmelissa87: ah, great idea! Maybe we should add this as an actual task to flow? 🤔

hamstu: Yeah, no rush to get this in now! A task would be a great way to keep it in mind. 👍

alvarezmelissa87:
Heya @hamstu - I've gone ahead and updated with the suggestions! Thanks so much for taking a look! 😄

PR #255, opened 2018-01-25
remove bundle from git

emily-plummer:
### Purpose
Remove bundle from git :) 

#### Staging Deployment


hharnisc:
This also includes some fixes for bugsnag and updates the session-manager version

PR #256, opened 2018-01-30
[WIP] Create and consume a flag that configures local vs prod services

hharnisc:
### Purpose

Enables use of production containers locally

### Notes

The flag is only used server side and passed through using the environment RPC endpoint

### Review

Does this have any unintended side effects locally in other packages?


hharnisc:
Thinking about this more we don't really need this now :D

hamstu:
Hey Harrison, not sure if this is the same thing; but the title reminded me a of a conversation I had with @twanlass where he said that locally the Reply app actually points to production APIs, which has simplified local dev setup/testing. Have we ever considered this? Would be a simple 'switch' to add? Not saying we should go for it, but it's an interesting idea.

twanlass:
Nice one @hamstu 😄. To share some additional context too, we're able to boot up our react client locally which then points to a remote dev environment (remote backend, db, and other services). With just the client running locally, changes are fast and no complex setup is needed (no containers, vagrant, et el). 

cc @humberaquino

hamstu:
@twanlass Yeah, having an 'entirely containerless' mode sounds very interesting. Esp. if you're only working on front-end. Might be a bit harder with the way our stack is setup, but I think it's worth exploring.

PR #258, opened 2018-02-05
remove unused environment variables

hharnisc:
### Purpose

Cleans up some of the environment variables that used to be used for login related tasks

### Notes

Removed session service being part of the health check - this would have caused down time if the old session service was removed

### Review

Anything look strange here?


hharnisc:
Merging since :green_apple:  and working on the branch build on dev

PR #259, opened 2018-02-06
{WIP} Task/calendar ui

emily-plummer:
### Purpose
Adds the mini calendar! This is UI only right now, and is hidden behind a feature flip `mini_calendar` :D 

#### Staging Deployment


packages/queue/index.js
```diff
@@ -36,6 +36,8 @@ export default connect(
         environment: state.environment.environment,
         editMode: state.queue.editMode,
         editingPostId: state.queue.editingPostId,
+        showCalendar: currentProfile.showCalendar,
+        hasCalendarFeatureFlip: state.appSidebar.user.features.includes('mini_calendar'),
```
emily-plummer: @alvarezmelissa87 might be having a brain fart but wonder if there's a cleaner way to do this? I always get nervous chaining lots of things in JS.

packages/profile-page/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@bufferapp/profile-page",
-  "version": "1.2.10",
+  "version": "1.3.0",
   "description": "profile page package",
```
alvarezmelissa87: Heya @emily-plummer - more a double-check here, do we need to update this manually? Won't this update automatically once we run `npm run publish`? looping in @hamstu for a quick check.

packages/queue/components/MiniCalendar/index.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import DayPicker from 'react-day-picker';
+import 'react-day-picker/lib/style.css';
+
+const containerStyle = {
+  position: 'absolute',
+  right: '37px',
+  marginTop: '70px',
```
alvarezmelissa87: Curious if we should update this to use rem's instead to keep consistent? 🤔 Sorry if you've already thought of this! 😸

packages/profile-page/package.json
emily-plummer: I didn't update this one manually, this is a side effect of running the publish task trying to fix the errors with the composer package 🤔 I need to double check on this before merge - the task said it failed but it did make the version commits so not sure if I should manually change this back to what's listed on npm or leave it

emily-plummer: Looks like the publish didn't make it to npm so my hunch is we should probably manually change these versions back

packages/queue/components/MiniCalendar/index.jsx
emily-plummer: Good catch! These will need to be readjusted later on anyways once we hear back from design so I wasn't super precise here! When I go back to adjust I'll switch them to rems :)

PR #266, opened 2018-02-09
bump session manager version

hharnisc:
### Purpose

Move to default branch package version


hharnisc:
Package is identical just with a non-beta version - dev deploy looks good!

PR #271, opened 2018-03-07
Update session manager package

hharnisc:
### Purpose

Update session manager package

### Notes

The latest session manager package redirects to the `login` app if the session service cannot be reached

### Review

Did I miss any dependencies to update?


hamstu:
Looks good to me @hharnisc! Thanks for the update. 

Your emoji of the day: 🖇 

😄

PR #274, opened 2018-04-24
[WIP] Upgrade to Pro modal

msanroman:
Adds the upgrade to pro modal to Buffer Publish!


hamstu:
@msanroman I've pushed some more changes here. The modal now shows/hides from the 'Upgrade' link in the tab bar.

The tests related to the Stripe package are failing, however. Do you want to take a look at that when you have a chance?

PR #278, opened 2018-05-23
Remove Fullstory to follow GDPR rules

josem:
### Purpose
https://buffer.atlassian.net/browse/PUB-341

### Notes
Not sure how to deploy this change @hamstu in Publish, could you help me with that? :) 

### Review


josem:
I've removed the other code that was failing and it's looking better now @hamstu! Thanks so much! 💯

PR #281, opened 2018-05-28
Upgrade to Webpack 4 and remove some warnings when building

josem:
### Purpose
Upgrade to (almost) latest version of Webpack 4 and remove a couple of warnings when building the project. 

### Notes
There is one warning remaining regarding the postcss-loader I think but I haven't been able to fix it with the time I have now for this.

```
(node:22068) DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
```

#### Staging Deployment

https://task-upgrade-webpack-v4.publish.buffer.com :smile:_


josem:
Oh it seems there's an issue with one of the checks, I'll try to see what happens :)

josem:
I'm closing this PR for now since I don't have the time to debug the problem. I also realized that Webpack 4 didn't seem to improve the speed for our local build with this configuration so I might come back to this in between cycles in the future after chatting with the team.

PR #288, opened 2018-06-11
Show upgrade modal when queue limit is hit from composer

hamstu:
### Purpose

Just a bit o' code to catch the new 'event' from the composer when the user hits the 'See paid plans' button.

⭐️ See https://github.com/bufferapp/buffer-composer/pull/67 for the deets 

#### Staging Deployment

### DIFF
--- a/packages/queue/components/ComposerWrapper/index.jsx
+++ b/packages/queue/components/ComposerWrapper/index.jsx
@@ -56,4 +56,9 @@ export default connect(
     }
     return {};
   },
+  (dispatch) => ({
+    onEvent: (type, data) => {
+      dispatch({ type: 'COMPOSER_EVENT', eventType: type, data })
+    }
+  }),
 )(ComposerWrapper);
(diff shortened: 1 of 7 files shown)


packages/queue/components/ComposerWrapper/index.jsx
```diff
@@ -56,4 +56,9 @@ export default connect(
     }
     return {};
   },
+  (dispatch) => ({
+    onEvent: (type, data) => {
```
hamstu: 🤞 This `onEvent` method is passed as a prop to the child `<Composer>` a bit later, and is called anytime an event is broadcasted on the composer's internal 'pubsub'.

hamstu: https://github.com/bufferapp/buffer-composer/blob/48782fd6949b6255cb90903c29e7df0e689b7122/interfaces/buffer-publish.jsx#L27-L30

PR #292, opened 2018-06-16
Upgrade modal deep link

hamstu:
### Purpose

Support linking directly to upgrade modal in New Publish.

URL: https://publish.buffer.com/#upgrade-to-pro

### Notes

This is a very simple implementation for now. In the future we may need to pass a source/utm parameter, but we'll be able to modify it.

#### Staging Deployment


hamstu:
Great feedback @msanroman! I updated the structure (adding a middleware as you suggested) in https://github.com/bufferapp/buffer-publish/pull/292/commits/93be3bfa28cc736fbfab038cd5c71a88dfb39eb3. 👍

PR #295, opened 2018-06-20
[PUB-437] ✨ 2FA in Publish Settings

hamstu:
### Purpose

Setup and management of 2FA in Publish.

➕Also included some small improvements to the `ProfileBadge` component since it was bugging me how it was displayed (or not displayed for some services.)

### Notes

Task: [PUB-377](https://buffer.atlassian.net/browse/PUB-377)

### Review

Does the state machine make sense? Is it well tested?

#### Staging Deployment


packages/preferences/components/Security/index.jsx
```diff
@@ -1,6 +1,15 @@
 import React from 'react';
+
 import ChangePassword from '@bufferapp/change-password';
+import TwoFactorAuth from '@bufferapp/publish-two-factor-auth';
+import { Divider } from '@bufferapp/components';
 
-const Security = () => <ChangePassword />;
+const Security = () => (
+  <React.Fragment>
```
hharnisc: An even better way to future proof this in case they move fragment out like they did to `prop-types`:

```js
import React, { Fragment } from 'react';
//... 
<Fragment />
```

packages/two-factor-auth/components/Modal/screens/chooseMethod.jsx
```diff
@@ -0,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Text, Button } from '@bufferapp/components';
+
+const ChooseMethod = ({ transition }) => (
+  <React.Fragment>
```
hharnisc: Same as above :smile:

packages/two-factor-auth/components/Toggle/index.jsx
```diff
@@ -0,0 +1,29 @@
...
+import PropTypes from 'prop-types';
+
+import { Button } from '@bufferapp/components';
+
+const TwoFactorToggle = ({ machineState, enabled, onClick }) => {
+  const toggleEnabledStates = ['enabled', 'disabled'];
+  const disabled = !toggleEnabledStates.includes(machineState);
+  if (enabled) {
+    return (
+      <Button tertiary onClick={onClick} disabled={disabled}>
```
hharnisc: Looks like it will be easy to drop in the Toggle :smile:

packages/preferences/components/Security/index.jsx
hamstu: Great idea! Made the change. 👍

packages/profile-sidebar/components/ProfileBadge/index.jsx
```diff
@@ -4,51 +4,52 @@ import {
...
-  right: '0.05rem',
-  background: 'white',
-  borderRadius: '50%',
-  height: '0.90rem',
-  width: '0.90rem',
-};
+const badgeTypes = PropTypes.oneOf(constants.SERVICE_NAMES);
 
-const badgeTypes = PropTypes.oneOf(['twitter', 'facebook', 'instagram', 'linkedin']);
+const profileBadgeIconMap = new Map([
```
msanroman: So clean ✨

packages/two-factor-auth/components/Modal/screens/chooseMethod.jsx
```diff
@@ -0,0 +1,37 @@
...
+const ChooseMethod = ({ transition, setupApp, loading, editMode }) => (
+  <Fragment>
+    {loading && <Text size="large">Please wait...</Text>}
+    <div style={{ display: loading ? 'none' : 'block' }}>
+      <div style={{ textAlign: 'center' }}>
+        <Text size="large">
+          {!editMode && 'Enable Two Factor Authentication'}
+          {editMode && 'Change Two Factor Authentication'}
+        </Text>
+        <div style={{ margin: '12px 0' }}>
```
msanroman: Should we be using `rem`s for margins and paddings?

packages/two-factor-auth/machine.js
```diff
@@ -0,0 +1,120 @@
+const handleEnabled = state => (state.isEnabled ? 'enabled' : 'disabled');
+const handleEditMode = state => (state.editMode ? 'enabled' : 'recovery');
+const handleConfirmBack = state => (state.updateMethod === 'app' ? 'setupApp' : 'setupSMS');
+
+const machine = {
+  disabled: {
+    ENABLE: 'chooseMethod',
```
msanroman: I think it'd be great to move all these strings to constants, since we are using them for multiple things (not only setting up the machine, but comparing then the dynamic "current" state with these strings as well).

packages/two-factor-auth/machine.js
```diff
@@ -0,0 +1,120 @@
...
+  state,
+  transitionName,
+  nextMachineState,
+  /* params, */
+}) => {
+  const { machineState: currentMachineState } = state;
+  let newState = {};
+
+  // Clear any errors when closed
+  if (['CLOSE', 'BACK'].includes(transitionName)) {
```
msanroman: Like we were chatting about yesterday (and maybe not for this specific case, as things are pretty much set in stone, and it won't need to change often), it'd be great to keep in mind how we can create some patterns to prevent this `transitionName` and `nextMachineState` checks. Instead, having some form of callbacks for each state and transitions that trigger whatever they need to do.

packages/two-factor-auth/middleware.js
```diff
@@ -0,0 +1,108 @@
...
+
+export default ({ dispatch, getState }) => next => (action) => {
+  next(action);
+  switch (action.type) {
+    case actionTypes.TRANSITION:
+      /**
+       * When the user clicks the disable toggle we hit the API
+       * with an 'off' method to disable TFA
+       */
+      if (getState().twoFactorAuth.machineState === 'disabled' &&
```
msanroman: It would be great to also see some ways we could move these `if` statements away from the `switch/case` statements, as this feels a bit harder to read through and maintain.

PR #296, opened 2018-06-22
Use new Account Management upgrade endpoint

msanroman:
### Purpose

Uses the new Account Management upgrade endpoint specified on
https://github.com/bufferapp/README/blob/master/billing/api-endpoints.md
and released on https://github.com/bufferapp/buffer-web/pull/15173.

The RPC endpoint encapsulates the config details that the user can't
choose (product is "publish" and the plan is "pro").

### DIFF
--- a/packages/server/rpc/upgradeToPro/index.js
+++ b/packages/server/rpc/upgradeToPro/index.js
@@ -12,7 +12,7 @@ module.exports = method(
         method: 'POST',
         strictSSL: process.env.NODE_ENV !== 'development',
         json: true,
-        body: {
+        form: {
           cycle,
           stripeToken: token,
           access_token: session.publish.accessToken,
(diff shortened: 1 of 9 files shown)


packages/server/rpc/upgradeToPro/index.js
```diff
@@ -6,16 +6,18 @@ module.exports = method(
...
       strictSSL: process.env.NODE_ENV !== 'development',
       body: {
         cycle,
         stripeToken: token,
         access_token: session.publish.accessToken,
+        product: 'publish',
+        plan: 'pro',
       },
     })
     .then(result => JSON.parse(result))
```
philippemiguet: Hey @msanroman - Just so you know, I've added to the [README](https://github.com/bufferapp/README/blob/master/billing/api-endpoints.md) an example of what the result will look when an error occurs :)

msanroman:
It's working great locally @philippemiguet! Thanks so much for your hard work building this new API! 👏

PR #297, opened 2018-06-28
Task/two factor auth settings recovery code

hharnisc:
### Purpose

Implement and wire up two factory recovery code component

https://buffer.atlassian.net/browse/PUB-441

### Notes

Created a component in publish shared components: AutoSelectText

### Review

How's the UX around automatically copying text?


hharnisc:
auto-select

PR #301, opened 2018-07-10
[PUB-378] Add Apps & Extras preferences tab

josem:
### Purpose
Add this tab to preferences:
image

(Based on this design: https://buffer.invisionapp.com/share/X6K7ALUCB83#/screens/301237410).

### Notes
- I still need to adjust the styling but the rest is working.
- I've added [some constants](https://github.com/bufferapp/buffer-publish/pull/301/files#diff-6245488a19eee7730cc1e2e88061c758) for the names of the tabs. //cc @hharnisc 

### Review
I'd love to have @msanroman and @hharnisc's reviews if they have some time. Please, let me know any potential improvements or mistakes you might see :)

#### Staging Deployment

https://task-apps-extras-settings.publish.buffer.com :smile:_


packages/preferences/components/Preferences/index.jsx
```diff
@@ -21,10 +23,10 @@ const PreferenceContent = ({ tabId, onUnknownTab }) => {
           <CloseAccount />
         </div>
       );
-    case 'security':
+    case TabsNames.SECURITY:
       return <Security />;
-    case 'appsandextras':
-      return <Text>Apps & Extras Content Goes Here</Text>;
+    case TabsNames.APPS_EXTRAS:
```
hharnisc: Nice work here @josem :raised_hands:

packages/preferences/components/Preferences/index.jsx
```diff
@@ -70,9 +72,9 @@ const Preferences = ({
     >
       <div id={'tabs'}>
         <Tabs selectedTabId={selectedTabId} onTabClick={onTabClick}>
-          <Tab tabId={'general'}>General</Tab>
-          <Tab tabId={'security'}>Security</Tab>
-          <Tab tabId={'appsandextras'}>Apps & Extras</Tab>
+          <Tab tabId={TabsNames.GENERAL}>General</Tab>
+          <Tab tabId={TabsNames.SECURITY}>Security</Tab>
+          <Tab tabId={TabsNames.APPS_EXTRAS}>Apps & Extras</Tab>
```
hharnisc: :100:

packages/manage-apps-extras/reducer.js
```diff
@@ -0,0 +1,65 @@
...
+    case `connectedApps_${dataFetchActionTypes.FETCH_SUCCESS}`: {
+      return {
+        ...state,
+        connectedApps: action.result,
+      };
+    }
+    case `revokeConnectedApp_${dataFetchActionTypes.FETCH_SUCCESS}`:
+      return {
+        ...state,
+        submitting: true,
```
hharnisc: I think this is going to get stuck in the submitting state. I think the flow would look something like this:

```js
case `revokeConnectedApp_${dataFetchActionTypes.FETCH_START}`:
      return {
        ...state,
        submitting: true,
     }
case `revokeConnectedApp_${dataFetchActionTypes.FETCH_SUCCESS}`:
case `revokeConnectedApp_${dataFetchActionTypes.FETCH_FAIL}`:
      return {
        ...state,
        submitting: false,
     }
```

packages/manage-apps-extras/reducer.js
```diff
@@ -0,0 +1,65 @@
...
+  REQUEST_OPEN_MODAL: 0,
+  REQUEST_CLOSE_MODAL: 0,
+  REQUEST_REVOKE_APP: 0,
+});
+
+const initialState = {
+  showModalAppId: null,
+  showModalAppName: '',
+  connectedApps: [],
+};
```
hharnisc: would be great to have the initial `submitting: false` set here :smile:

packages/manage-apps-extras/components/AppsManager/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+  connectedApps,
+  showModalAppId,
+  showModalAppName,
+  onRequestOpenModal,
+  onRequestCloseModal,
+  onConfirmRevokeApp,
+  submitting,
+}) => (
+  <div
+    style={{
```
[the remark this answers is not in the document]
hharnisc: After doing polishing on the other components, we've got a `Row` component (in the shared components package) you could use instead of the div

https://github.com/bufferapp/buffer-publish/blob/master/packages/shared-components/Row/index.jsx

All of the other preferences tabs use this one now :smile: so they'll all look the same if you use the `Row`

packages/manage-apps-extras/components/AppsAndExtras/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+  onRequestOpenModal,
+  showModalAppId,
+  showModalAppName,
+  onRequestCloseModal,
+  onConfirmRevokeApp,
+  submitting,
+}) => (
+  <div>
+    <ExtrasLinks />
+    <AppsManager
```
msanroman: If you're just passing all the props from this component to `AppsManager` a quicker way to handle it would be:

```
const AppsAndExtras = (props) => (
  <div>
    <ExtrasLinks />
    <AppsManager {...props} />
  </div>
);
```

This destructures the `props` object and passes all the props to the component in the same way you're doing :)

packages/manage-apps-extras/reducer.test.js
```diff
@@ -0,0 +1,123 @@
...
+    };
+    const action = {
+      type: actionTypes.REQUEST_CLOSE_MODAL,
+    };
+    deepFreeze(action);
+    expect(reducer(undefined, action))
+      .toEqual(stateAfter);
+  });
+
+  it('should handle REQUEST REVOKE APP action type', () => {
```
msanroman: I think it'd be great to rename these tests to what's really intended (documenting the state change, rather than the "mechanics" of action handling). In this case, I'd say something like `REQUEST_REVOKE_APP removes the selected appId`, or something like that :)

packages/preferences/components/AppsAndExtras/index.jsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+import ManageAppsAndExtras from '@bufferapp/manage-apps-extras';
+
+const AppsAndExtras = () => <ManageAppsAndExtras />;
```
msanroman: If this component is just rendering another one maybe we can just reference`ManageAppsAndExtras` where `AppsAndExtras` is being used?

hharnisc:
Just took a quick look at the changes you've made! This is looking great @josem :shipit:

PR #320, opened 2018-08-10
[PUB-541] Add Mini-calendar front-end tweaks

anafilipadealmeida:
### Purpose
Fix front-end issues on MiniCalendar.

### Notes
Task: [PUB-541](https://buffer.atlassian.net/browse/PUB-541)


anafilipadealmeida:
It slightly decreases the code coverage, but this has only a few front-end tweaks. As soon as the feature is finished and fully functional, we're gonna add tests to it.

PR #322, opened 2018-08-16
[PUB-573] Add composer notifications when scheduling a post

anafilipadealmeida:
### Purpose
Add composer notifications when scheduling a post.

### DIFF
--- a/packages/queue/middleware.test.js
+++ b/packages/queue/middleware.test.js
@@ -1,14 +1,56 @@
 import { actionTypes as notificationActionTypes } from '@bufferapp/notifications';
+import { actionTypes as profileActionTypes } from '@bufferapp/publish-profile-sidebar';
+import {
+  actions as dataFetchActions,
+  actionTypes as dataFetchActionTypes,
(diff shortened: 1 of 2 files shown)


anafilipadealmeida:
@hamstu I struggled a bit to name each unit test, I took a look at other middleware.test.js files but couldn't find a pattern. Suggestions are very welcome!

packages/queue/middleware.test.js
```diff
@@ -1,8 +1,340 @@
...
+      .toBeCalledWith(action);
+    expect(dispatch)
+      .toBeCalledWith(expect.objectContaining({
+        type: notificationActionTypes.CREATE_NOTIFICATION,
+        notificationType: 'success',
+        message: action.data.message,
+    }));
+  });
+
+  it('should requeuePost_${dataFetchActionTypes.FETCH_SUCCESS}', () => {
```
hamstu: Ah, yes some of these are quite confusing! 😅 
I would rename this one to:

```
it('should fetch posts again after a post is requeued', ...)
```

PR #324, opened 2018-08-17
[PUB-623] Don't show app switcher for NTB-F Users

MayaUribe:
### Purpose
When feature ‘new_publish_new_buffer_free_users’ is present prevent showing the app switcher in New Publish.

### Review
Verify that the app switcher doesn't display for New Buffer Free Users.

#### Staging Deployment

### DIFF
--- a/packages/app-switcher/reducer.test.js
+++ b/packages/app-switcher/reducer.test.js
@@ -1,0 +1,72 @@
+import {
+  actionTypes as dataFetchActionTypes,
+  actions as dataFetchActions,
+} from '@bufferapp/async-data-fetch';
+
+import reducer from './reducer';
+
+describe('reducer', () => {
+  describe('actions', () => {
+    it('user_fetchSuccess triggers a FETCH_SUCCESS action', () => {
+      const id = 1;
+      const name = 'sendFeedback';
+      const result = 'result';
+      const args = {
+        body: 'test',
+      };
+      expect(dataFetchActions.fetchSuccess({ name, args, id, result })).toEqual({
+        type: `user_${dataFetchActionTypes.FETCH_SUCCESS}`,
+        name,
+        args,
+        id,
+        result,
+      });
+    });
+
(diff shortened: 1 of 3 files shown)


packages/app-switcher/reducer.test.js
```diff
@@ -0,0 +1,72 @@
+import {
+  actionTypes as dataFetchActionTypes,
+  actions as dataFetchActions,
+} from '@bufferapp/async-data-fetch';
+
+import reducer from './reducer';
+
+describe('reducer', () => {
+  describe('actions', () => {
+    it('user_fetchSuccess triggers a FETCH_SUCCESS action', () => {
```
hharnisc: Hey @MayaUribe great working starting to add tests, we need these :raised_hands:  Looks like these tests are actually testing `@bufferapp/async-data-fetch` package rather than the `app-switcher` package. I'd expect these tests to be testing functions from the reducer that was imported on line 6 :smile: (highlighted the functions below).

Here's an example of testing another reducer: https://github.com/bufferapp/buffer-publish/blob/master/packages/app-sidebar/reducer.test.js

The idea is you pass in the intial state (stateBefore) and the action to the `reducer` and the check it against the expected state (stateAfter)

So you're looking to test:

Reducer :point_right: https://github.com/bufferapp/buffer-publish/blob/master/packages/app-switcher/reducer.js#L17

Action Creator: :point_right: https://github.com/bufferapp/buffer-publish/blob/master/packages/app-switcher/reducer.js#L51

How does that sound? :smile:

MayaUribe: Sounds great actually! thank you a lot for the help here 💯:D

MayaUribe: Hey @hamstu, since Harrison is on his time off, would you mind checking this one out? 😃Thank you a lot!

MayaUribe:
@hamstu would you be up to reviewing the test cases? 😃😃

hamstu:
This is good to merge @MayaUribe! 💯

PR #327, opened 2018-08-22
[PUB-632] Adding condition in new Publish for new feature flip

MayaUribe:
…Buffer Free Users” to Buffer Classic if they have the “new_publish_new_buffer_free_users” feature flip.

### Purpose
Adding a condition in new Publish, so that it doesn’t redirect “New to Buffer Free Users” to Buffer Classic if they have the “new_publish_new_buffer_free_users” feature flip.

### Review
Verify that the condition is placed correctly.

#### Staging Deployment


MayaUribe:
Hey @hamstu what should I do regarding those two failing checks?

PR #335, opened 2018-08-30
Update React Hot Loader + Tweaks ⚡️

hamstu:
### Purpose

* 🌶 Update `react-hot-loader` and get it working again (*See notes below).
* 🔨 Fix small UI bugs, shadows, misalignment, etc.
* 👍 Update the composer dependency and make changes on publish side to clean up all the noisy composer PropTypes / Immutable warnings. (https://github.com/bufferapp/buffer-composer/pull/85)
* Stop rebuilding `grpc` every time and add a command to do it when needed (`yarn run rebuild-grpc`). Not an ideal solution yet, but should prevent constant need to rebuild by taking out the `postinstall`.

### Notes

You may notice that only `profile-page` and it's children are enabled for hot reloading. I tried making the entry point 'App' component hot reloadable, but it doesn't work. Something to do with the way the other components are linked together in the dependency graph means it just triggers a full reload when they change, Whereas marking specific components, such as the `profile-page` _does_ work.

image

Might be something we can fix to work from the top, but it's working for now on a per package basis, and just wrapping more components with `hot()` is pretty easy if we want it in other places.

#### Staging Deployment


package.json
```diff
@@ -113,7 +113,7 @@
   },
   "jest": {
     "transformIgnorePatterns": [
-      "node_modules/(?!@bufferapp/*)"
+      "/node_modules(?!\/@bufferapp\/performance-tracking)(?!\/@bufferapp\/async-data-fetch)(?!\/@bufferapp\/components)(?!\/@bufferapp\/web-components)(?!\/@bufferapp\/composer)/"
```
hamstu: Being more explicit here because otherwise Jest tries to transpile the newly scoped package I created [`@bufferapp/react-simple-dropdown`](https://github.com/bufferapp/react-simple-dropdown.) which is used by the composer.

package.json
```diff
@@ -4,19 +4,19 @@
   "version": "0.0.1",
   "description": "🍩",
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "yarn",
     "clean": "lerna clean",
     "test": "if [ -n \"$CI\" ]; then yarn && jest --runInBand; else jest; fi",
     "test-update": "jest -u",
-    "init": "yarn install && lerna bootstrap --loglevel verbose",
+    "init": "yarn",
```
hamstu: Not sure if even having `bootstrap` or `init` commands is even needed now, since `yarn` alone does all we need. 🤔

packages/queue/components/ComposerPopover/index.jsx
```diff
@@ -9,8 +9,8 @@ const ComposerPopover = ({
   preserveComposerStateOnClose,
 }) => (
   <Popover
-    left={'21rem'}
-    top={'4.7rem'}
+    left={'337px'}
+    top={'73px'}
```
hamstu: Line up the composer better! #pixelperfect

packages/queue/components/ComposerWrapper/index.jsx
```diff
@@ -1,44 +1,14 @@
...
-};
-
-ComposerWrapper.defaultProps = {
-  profiles: [],
-  enabledApplicationModes: [],
-  csrfToken: '1234', // dummy string for now since MC requires csrfToken
-  post: {},
-};
+ComposerWrapper.propTypes = Composer.propTypes;
+ComposerWrapper.defaultProps = Composer.defaultProps;
```
hamstu: Make this more of a lightweight wrapper by keeping the propTypes/defaults in the composer package.

packages/server/webpack.config.common.js
```diff
@@ -19,6 +19,9 @@ module.exports = {
         exclude: /node_modules(?!\/@bufferapp\/performance-tracking)(?!\/@bufferapp\/async-data-fetch)(?!\/@bufferapp\/components)(?!\/@bufferapp\/web-components)(?!\/@bufferapp\/composer)/,
         use: {
           loader: 'babel-loader',
+          options: {
+            cacheDirectory: true,
+          },
```
hamstu: I think this helps? 🤷‍♂️ 

Didn't do any intensive testing, but I saw it somewhere and thought it would be worth trying.

packages/server/webpack.config.dev.js
```diff
@@ -4,32 +4,31 @@ const merge = require('webpack-merge');
 // const { BundleAnalyzerPlugin } = require('webpack-bundle-analyzer');
 const common = require('./webpack.config.common.js');
 
-const merged = merge.strategy({
-  plugins: 'prepend',
-  'entry.bundle': 'prepend',
-})(common, {
-  entry: {
-    bundle: ['react-hot-loader/patch'],
```
hamstu: This patch is no longer needed in the latest `react-hot-loader`.

PR #341, opened 2018-09-05
Hotfix: Prevent static asset issues from failing in prod

djfarrelly:
### Purpose
Prevent failing deploys like the previous one where `yarn run init` failed and static assets was empty in production.

### Notes
- Decided to use the `set -e` approach rather than the more verbose 6cdb52e23fdcb85fa5ccdb7d5932030ee76d5e86

### Review
TBD


helm/values.yaml
```diff
@@ -1,6 +1,6 @@
 name: buffer-publish # Override to be the name of the application
 track: stable # stable | canary
-replicaCount: 1
+replicaCount: 5
```
djfarrelly: We're in production now, we should have multiple replicas 👍

PR #343, opened 2018-09-06
Server package uses @bufferapp/buffer-rpc

hharnisc:
### Purpose

Replace `micro-rpc` with `buffer-rpc` which is made to work with express.

### Notes

- Express does not allow us to use things like async/await so buffer-rpc mostly does things with promises
- required a small change to the checkToken middleWare

### Review

How's this acting on the staging branch deployment?

#### Staging Deployment


hharnisc:
Ah yep, testing are passing locally @djfarrelly - adding Travis over there now

PR #344, opened 2018-09-07
[PUB-542] Add number of posts to Mini Calendar

anafilipadealmeida:
### Purpose
Retrieve live data from the API to get number of posts per day (when loading a month in the mini calendar)

### Notes
Task: [PUB-542](https://buffer.atlassian.net/browse/PUB-542)


packages/queue/components/MiniCalendar/index.jsx
```diff
@@ -57,6 +52,7 @@ const NavBar = ({
   onNextClick,
   className,
   localeUtils,
+  getNumberPostsByDate,
```
hamstu: 💅 Tiny nitpick here, but I wonder if this method might read better as `getNumberOfPostsByDate`?

packages/queue/components/QueuedPosts/index.jsx
```diff
@@ -89,13 +91,12 @@ const QueuedPosts = ({
...
-              {showCalendar ? 'Hide Calendar' : 'Show Calendar'}
-            </Button>
-          </div>}
-        {showCalendar && hasCalendarFeatureFlip && <MiniCalendar />}
+        <div style={buttonStyle}>
+          <Button secondary onClick={onCalendarToggleClick}>
+            {showCalendar ? 'Hide Calendar' : 'Show Calendar'}
+          </Button>
+        </div>
+        {showCalendar && <MiniCalendar numberPostsByDate={numberPostsByDate} onMiniCalendarMonthChange={onMiniCalendarMonthChange}  />}
```
hamstu: Since it's a prop on the `MiniCalendar` component it might be a bit simpler to just call it `onMonthChange`. What do you think? :)

packages/queue/components/MiniCalendar/index.jsx
```diff
@@ -11,6 +11,7 @@ import { ArrowRightIcon, ArrowLeftIcon } from '@bufferapp/components/Icon/Icons'
 import { Button } from '@bufferapp/components';
 import 'react-day-picker/lib/style.css'; // Not being used.
 import moment from 'moment-timezone';
+import { actions } from '../../reducer';
```
hamstu: I don't think this `import` is needed, since you have passed down methods that dispatch your action(s) (which is perfect! 👌)

packages/queue/components/MiniCalendar/index.jsx
```diff
@@ -96,39 +98,54 @@ const NavBar = ({
...
-    />
-  </div>;
+const MiniCalendar = ({numberPostsByDate, onMiniCalendarMonthChange}) => {
+
+  /* Requests the number of posts for the current month when open the calendar */
+  if(!numberPostsByDate) {
+    const startDate = moment().startOf('month').unix();
+    const endDate = moment().endOf('month').unix();
+    onMiniCalendarMonthChange(startDate, endDate);
+  };
```
hamstu: Instead of fetching the posts in the initial render here; another way would be to listen in on the queue package's middleware for `actionTypes.TOGGLE_CALENDAR`and dispatch the action to fetch there. Then you're better able to separate the data fetching logic.

hamstu:
⚠️  It also looks like the tests are failing, so be sure to get those sorted out before merging as well.

packages/queue/components/QueuedPosts/index.jsx
anafilipadealmeida: 👍

PR #345, opened 2018-09-10
[PUB-612] Add full story back into new publish

karelm:
### Purpose

Add Full story back to new publish, but exclude any data that may be sensitive to users from the recording, such as email, username, and phone numbers

### Notes

Task: [PUB-612](https://buffer.atlassian.net/browse/PUB-612)
#### Staging Deployment


packages/shared-components/SensitiveData/index.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const SensitiveData = ({ children }) => children ? <span className="sensitiveData">{children}</span> : null;
```
msanroman: Beautiful!

PR #351, opened 2018-09-17
Maintenance page

hharnisc:
### Purpose

Display a maintenance page when the application is in maintenance mode

### Notes

@josem feel free to take over on this when you start you day :wink: 

#### Staging Deployment


hharnisc:
@josem still working on getting the routes setup so going to `/maintenance` shows the `MaintenancePage` component.

Also working on the redirect if the maintenance response is detected.

hharnisc:
Hey @josem I decided to take a different approach than what was in the branch and needed to force push.

The big difference being that the maintenance page is being served from the backend.

hharnisc:
@djfarrelly @josem here's the ultra-lean maintenance page

image

This pages gets redirected to when someone is already signed into publish and makes a request against the Buffer API. It will also redirect back to the application when we flip back out of maintenance mode. So refreshing the page when we're back online will get them back into the application.

If you're cool with the _design_ this is ready to go.

hharnisc:
Merging this since the tests are passing and have manually verified the build via the dev deployment.

PR #354, opened 2018-09-20
[PUB-500] Addressing issue with Drag and Drop not working when intended

karelm:
### Purpose

We are now reloading the post list in the queue when the scheduled times get modified in the settings panel, to ensure that we are always have the right schedule post times for the posts, and the Drag and Drop works as expected as a consequence

### DIFF
--- a/packages/profile-page/components/ProfilePage/index.jsx
+++ b/packages/profile-page/components/ProfilePage/index.jsx
@@ -108,7 +108,7 @@ const ProfilePage = ({
           growthSpace={1}
         >
           <div style={tabContentStyle}>
-            {TabContent({ tabId, profileId })}
+            <TabContent tabId={tabId} profileId={profileId} />
             {showLoadMoreButton &&
               <div style={buttonContainerStyle}>
                 <Button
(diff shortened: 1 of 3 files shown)


packages/profile-page/components/ProfilePage/index.jsx
```diff
@@ -108,7 +108,7 @@ const ProfilePage = ({
           growthSpace={1}
         >
           <div style={tabContentStyle}>
-            {TabContent({ tabId, profileId })}
+            <TabContent tabId={tabId} profileId={profileId} />
```
hharnisc: Nice! Love seeing cleanup changes like this :raised_hands:

packages/queue/middleware.js
```diff
@@ -24,6 +24,15 @@ export default ({ dispatch, getState }) => next => (action) => {
         },
       }));
       break;
+    case `updatePausedSchedules_${dataFetchActionTypes.FETCH_SUCCESS}`:
```
hharnisc: Great catch here @karelm - interesting to see updating the paused schedules caused queued posts to get stale and break DnD. Glad to see this one fixed :taco:

packages/queue/middleware.test.js
```diff
@@ -50,6 +50,26 @@ describe('middleware', () => {
     }));
   });
 
+  it('should fetch queuedPosts if updatePausedSchedules is successful', () => {
```
hharnisc: Excellent to see tests for this :muscle:

PR #356, opened 2018-09-21
hide (coming soon) on analyze in the sidebar

federicoweber:
### Purpose
To hide the '(Coming soon)' string when we are on Analyze, as in that case we know the user has access to Analyze.

### Notes
Hey @hamstu super small one :smile:, just want to make sure I'm not missing any cases where we could have that icon active and not being an Analyze customer.

Also tried to run `jest -u` to update the snapshots but I'm getting lots of errors parsing files from node_modules. Any reason for that?

Thanks!

#### Staging Deployment


federicoweber:
thank you @hamstu, and no worries at all :smile:. Really weird I could not run those tests at all.

PR #358, opened 2018-09-25
PUB-704 + PUB-705 | Implementing rpc logic + Free vs Pro component in UI

karelm:
### Purpose

Implement RPC calls to fetch plan features from the api endpoint 
Implement FeatureLoader component (yes the name is terrible) to show how we handle Free vs Pro in the sidebar

### Notes

I've created tests for FeatureLoader, so check those out for use case and how its all handled... may need to make some changes to the overall logic, but I think its okay as a starting point, so you can display based on plan, or feature supplied, so maybe its not a featureloader as it is a supported plan and feature loader, but I dont think that name works as smoothly, and having 2 seperate components seemed not quite the right approach here.


packages/server/rpc/index.js
```diff
@@ -1,66 +1,69 @@
...
+  composerApiProxyMethod,
+  connectedApps,
   deletePostMethod,
-  sharePostNowMethod,
+  draftPostsMethod,
   enabledApplicationModesMethod,
-  composerApiProxyMethod,
   environmentMethod,
-  updateScheduleMethod,
+  featureMethod,
```
lolaodelola: Is this the new features package?

packages/utils/index.js
```diff
@@ -11,6 +12,7 @@ module.exports = {
   date,
   postParser,
   profileParser,
+  featureParser,
```
lolaodelola: or is this the new features package? or are they both?

packages/server/rpc/index.js
karelm: Yes, this is the new featureMethod RPC call... but I haven't surfaced it, because I'm not using it anywhere yet, and it seemed wasteful to surface it until I had a purpose for doing so

packages/product-features/components/FeatureLoader/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+   children,
+   fallback,
+   supportedPlans,
+ }) => {
+  const {
+    planName,
+    features,
+  } = productFeatures;
+
+  if (typeof supportedPlans !== 'undefined') {
```
karelm: I should probably move this if statement logic into the reducer from here so that its reusable and from other places

packages/product-features/components/FeatureLoader/index.jsx
```diff
@@ -0,0 +1,59 @@
...
+    const supportedPlanList = (typeof supportedPlans === 'string' ? [supportedPlans] : supportedPlans)
+      .map(p => p.toLowerCase());
+    const currentPlan = planName.toLowerCase();
+    if (!supportedPlanList.some((plan) => plan === currentPlan)) {
+      return fallback;
+    }
+  }
+
+  if (typeof supportedFeatures !== 'undefined') {
```
karelm: I should probably also move this into the reducer logic too so that its reusable in other places, e.g. ProductFeature.IsInSupportedFeatures(['feature_one'])

karelm:
I was originally thinking the same @josem but after reading about having profile features, and user features, and plan features... I thought it was probably best to separate it into its own isolated component, then the features hopefully wont spill out into code that doesn't necessarily need it. But I guess only time will tell whether thats the case or not :)

But thanks for the review, I'll get it shipped! :shipit:

PR #360, opened 2018-09-25
[PUB-715] Add Re-buffer button

MayaUribe:
### Purpose
Add Re-buffer button to each update for "Sent Posts" tab.


packages/shared-components/PostList/index.jsx
```diff
@@ -11,7 +12,14 @@ import LinkPost from '../LinkPost';
 import VideoPost from '../VideoPost';
 import PostDragWrapper from '../PostDragWrapper';
 
+const reBufferWrapperStyle = {
+  paddingLeft: '1rem',
+  minWidth: '120px',
+};
+
 const postStyle = {
+  display: 'flex',
```
anafilipadealmeida: Since you added these styles, the snapshots don't match anymore (If you run `yarn run test` you can see that the 8 snapshots failing are due to the change in the `postStyle` const). If you run `yarn run test-update` you'll update them, and then just need to commit those changes in `__snapshots__/snapshot.test.js.snap` 😄

MayaUribe: Yay!! Thank you so much Ana! 💯🥇

PR #361, opened 2018-09-25
Use non-interactive yarn install flag

hharnisc:
### Purpose

Interactive mode can cause CI/CD to hang if an interaction is required by yarn

### Notes

This flag should allow it to fail -- and be caught by CI/CD as a failure

### Review

Did it fail as expected?

#### Staging Deployment


hharnisc:
Merging this now since it causes the build to fail rather than hang.

PR #363, opened 2018-09-26
[PUB-705] Switch sidebar from displaying publish free for pro users

karelm:
### Purpose

Switch the sidebar from saying publish free for pro users to saying publish pro instead using the new RPC endpoint

### Notes

Work in progress, feedback always welcome, definitely not stable at present / fully functional 😄


packages/server/rpc/index.js
```diff
@@ -1,66 +1,69 @@
 const { rpc } = require('@bufferapp/buffer-rpc');
```
karelm: The diff in this file is because i alphabetised the const's and utils in the rpc module, I need to revert this change

karelm:
Merging into 704 (#358) + will merge in together

PR #364, opened 2018-09-27
[PUB-714] Add post's stats bar

anafilipadealmeida:
### Purpose
Add post's stats bar to be displayed for Pro Users

### Notes
Task: [PUB-714](https://buffer.atlassian.net/browse/PUB-714)

### DIFF
--- a/packages/shared-components/Post/index.jsx
+++ b/packages/shared-components/Post/index.jsx
@@ -170,7 +170,7 @@ const Post = ({
         <FeatureLoader
           supportedFeatures={'post_stats'}
         >
-          {sent &&
+          {isSent && !postDetails.isRetweet &&
             <PostStats
               statistics={statistics}
               profileService={profileService}
--- a/packages/shared-components/Post/index.jsx
+++ b/packages/shared-components/Post/index.jsx
@@ -20,6 +20,7 @@ import {
 import PostFooter from '../PostFooter';
 import PostStats from '../PostStats';
 import RetweetPanel from '../RetweetPanel';
+import FeatureLoader from '@bufferapp/product-features';
 
 const getPostContainerStyle = ({ dragging, hovering }) => ({
   display: 'flex',
@@ -113,8 +114,6 @@ const renderContent = ({
 
 /* eslint-enable react/prop-types */
 
-const isUserPro = false;
-
 const Post = ({
   children,
   isConfirmingDelete,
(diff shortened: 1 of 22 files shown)


packages/shared-components/Post/index.jsx
```diff
@@ -164,6 +168,12 @@ const Post = ({
           dragging={dragging}
           onRequeueClick={onRequeueClick}
         />
+        {sent && isUserPro &&
```
anafilipadealmeida: Temporary conditions: will be changed once the FeatureLoader and the sent pro fix are up

packages/shared-components/QueueItems/story.jsx
```diff
@@ -9,9 +9,18 @@ import {
   postLists,
   postListsNoHeaders,
 } from './postData';
+import createStore from '@bufferapp/publish-store';
+import { Provider } from 'react-redux';
+
+const store = createStore();
```
hamstu: I notice you've added the store to all the post types. It's a pretty big chunk of code to bring in to every story. Is there a way we can maybe fake just what we need instead?

packages/shared-components/PostStats/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+  
+  return (
+    <div style={statsBarStyle}>
+      {Object.keys(titles).map((typeStats) => createElement(typeStats))}
+    </div>
+  );
+};
+
+PostStats.propTypes = {
+  profileService: PropTypes.oneOf(['twitter', 'linkedin', 'google', 'facebook', 'instagram', 'pinterest']),
```
hamstu: Just a small one, but we have a constants to make this a little shorter and more DRY. :)
https://github.com/bufferapp/buffer-publish/blob/5e62afec097f784d1516fa1a42c722d3c683165a/packages/profile-sidebar/components/ProfileBadge/index.jsx#L35

anafilipadealmeida: Oh, great! I didn't know about that, thanks for showing it to me! 👍

packages/shared-components/PostStats/index.jsx
```diff
@@ -0,0 +1,106 @@
...
+  display: 'flex',
+  alignItems: 'baseline',
+  justifyContent: 'center',
+  borderTop: `${borderWidth} solid ${mystic}`,
+  padding: '0.5rem',
+  borderRight: `${borderWidth} solid ${mystic}`,
+};
+
+/* Abbreviates numbers, eg: if we want 2 decimal places, it displays 1240 as 1.24k */
+const abbreviateNumber = (number, decPlaces) => {
```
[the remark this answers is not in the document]
anafilipadealmeida: 💯 Great suggestion, gonna do it!

PR #365, opened 2018-09-28
Hide action buttons on footer from sent tab

MayaUribe:
### Purpose
Hide action buttons on footer from Sent Tab.

### DIFF
--- a/packages/shared-components/PostFooter/index.jsx
+++ b/packages/shared-components/PostFooter/index.jsx
@@ -73,7 +73,6 @@ const PostFooter = ({
   onEditClick,
   onShareNowClick,
   postDetails,
-  sent,
   dragging,
   isSent,
 }) => {
@@ -117,7 +116,6 @@ PostFooter.propTypes = {
     error: PropTypes.string,
     postAction: PropTypes.string,
   }).isRequired,
-  sent: PropTypes.bool.isRequired,
   dragging: PropTypes.bool,
   onRequeueClick: PropTypes.func,
   isSent: PropTypes.bool,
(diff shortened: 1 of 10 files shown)


MayaUribe:
Hey @hamstu whenever you have time, would you mind checking this one out? 😄
I would like to merge this one before the changes Ana and I have for our current tasks.

Thank you!

packages/shared-components/PostFooter/index.jsx
```diff
@@ -75,6 +75,7 @@ const PostFooter = ({
   postDetails,
   sent,
   dragging,
+  isSent,
```
karelm: isSent has been added, and the sent prop already exists, and its mandatory... do we need to remove the sent variable from here as it doesnt look like its being used anywhere?

karelm:
@anafilipadealmeida + @MayaUribe I think that my comment about sent vs isSent likely applies to multiple files, but I've only highlighted it in one place... I think it would be great if that could get cleaned up here, but I think this is all great work, so if you think its still needed / used by something, then ignore my comment

otherwise I think its okay to get shipped 💯

packages/shared-components/PostFooter/index.jsx
MayaUribe: Hey there @karelm! yes, that sounds about right, sent is no longer used so I will clean that up 😃thank you for catching that up!

MayaUribe: @karelm would you mind taking a quick look so I merge this one out? :)

PR #368, opened 2018-10-01
[PUB-761] Implementing higher order component for feature loading

karelm:
### Purpose

Creating a higher order component to use for feature loading, when you want to handle loading in the feature yourself inside a component, instead of being told to draw or not

### Notes

JIRA task: https://buffer.atlassian.net/browse/PUB-761


packages/product-features/README.md
```diff
@@ -21,3 +27,59 @@ To see more of the usecases supported, check out the story.jsx file...
...
+    };
+
+    TextComponent.propTypes = {
+      features: PropTypes.any.isRequired,
+    };
+    
+    const TextComponentWithFeatureLoader = WithFeatureLoader(TextComponent);
+
+    // Now when you want to use it, you can just use it like any other component
+    return <TextComponentWithFeatureLoader size={'large'} productFeatures={fakeFeatures} />;
```
josem: One quick question @karelm! Sorry if it's documented elsewhere. 

Would we need to pass the plan as well here?

karelm: Good question @jose that information comes from the redux state that's already injected into the component, so unless you chose to use the raw component, WithFeatureLoader already knows the users current plans from the rpc call made to the features endpoint. And that triggers in response to hearing a user event being fired.

Hopefully that makes sense 🙂

josem: Oh yes sorry! I didn't spot that, makes total sense. Thanks so much!

PR #369, opened 2018-10-02
[PUB-716] Send again functionality

MayaUribe:
### Purpose
Display send again button on "Sent Posts" tab for Pro users.


MayaUribe:
Hey, @karelm! do you think you could take a look at this one? specially the part where I'm using the FeatureLoader, thank you a lot!

packages/shared-components/PostList/index.jsx
```diff
@@ -116,19 +117,22 @@ const PostList = ({
...
               onImageClose,
               onDropPost,
+              onShareAgainClick,
               isSent,
             })
           }
-          {isBusinessUser ?
+          <FeatureLoader
+            supportedPlans={'pro'}
+          >
```
karelm: is rebuffer a feature (as in the way that mini calendar is a feature flag 'mini_calendar') or is it actually just a consequence of being a pro user? if its the second then I think this is great... if its not, then you may need to add a new feature flag into the web project, and use supportedFeatures instead

But your usage is perfect 👍

karelm: also @MayaUribe if you want to test if its funcitoning correctly, go into the buffer admin screen on your local site, and change from whatever plan you are on in the billing screen to a pro user, and then you should be able to see the component appear and disappear before your eyes 👀

packages/shared-components/PostList/story.jsx
```diff
@@ -143,20 +142,4 @@ storiesOf('PostList', module)
       onImageClickPrev={action('onImageClickPrev')}
       onImageClose={action('onImageClose')}
     />
-  ))
```
karelm: how come this test is being removed here? just curious really?

packages/sent/index.js
```diff
@@ -35,11 +36,23 @@ export default connect(
...
         total: currentProfile.total,
-        isBusinessUser: state.appSidebar.user.is_business_user,
+        showComposer: state.sent.showComposer,
+        editMode: state.sent.editMode,
       };
     }
     return {};
   },
+  (dispatch, ownProps) => ({
+    onShareAgainClick: (post) => {
```
karelm: I think you could possibly use prop spreading here and use ({post}) => instead of (post) => but I think its minor

karelm:
@MayaUribe I think this is great work you have done here, I've got a couple of comments, but nothing major.

I would love to see some more tests around this functionality added, even if thats just the addition of a story or two to support the use-cases we expect to see here, but I don't think thats critical

I think its looking good to go 💯

packages/shared-components/PostList/story.jsx
MayaUribe: Because it wasn't really adding value, as it was before, I had an "isBusiness" flag I needed to check, but when I removed that, then that test ended up being exactly the same as the others, so it didn't make much sense.

packages/shared-components/PostList/index.jsx
MayaUribe: This is an awesome question, I need to confirm that one 😄

karelm: I do t think it stops us getting it out, just that we may need to circle back later

karelm:
I cant find the approve button... but I am happy with this work! Great stuff @MayaUribe 

Lets get it merged! 🚀

PR #370, opened 2018-10-03
Change sent post footer styles

anafilipadealmeida:
### Purpose
Add different styles to the footer of sent posts

### Notes
Also includes small tweaks to the stats bar design and an update to the stories due to changes on `sent` prop in https://github.com/bufferapp/buffer-publish/pull/365

### DIFF
--- a/packages/shared-components/ImagePost/story.jsx
+++ b/packages/shared-components/ImagePost/story.jsx
@@ -40,6 +40,11 @@ const postDetails = {
   postAction: 'This post will be sent at 9:21 (GMT)',
 };
 
+const postDetailsSent = {
+  isRetweet: false,
+  postAction: 'This post was sent at 9:21 (GMT)',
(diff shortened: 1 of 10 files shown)


packages/shared-components/ImagePost/story.jsx
```diff
@@ -80,7 +85,7 @@ storiesOf('ImagePost', module)
       onDeleteConfirmClick={action('delete-confirm-click')}
       onEditClick={action('edit-click')}
       onShareNowClick={action('share-now-click')}
-      sent={false}
```
anafilipadealmeida: @MayaUribe Just a small note: The stories I'm updating in this PR should probably have been done when we changed the `sent` prop to `isSent` (https://github.com/bufferapp/buffer-publish/pull/365). Nonetheless, it's fixed now 😄

PR #372, opened 2018-10-04
[PUB-801] Display custom footer message when IG reminder

anafilipadealmeida:
### Purpose
Add a custom message on the footer of each Instagram Reminder Post

### DIFF
--- a/packages/shared-components/PostFooter/index.jsx
+++ b/packages/shared-components/PostFooter/index.jsx
@@ -4,6 +4,7 @@ import {
   Text,
   WarningIcon,
   ClockIcon,
+  CircleInstReminderIcon,
 } from '@bufferapp/components';
 import {
   borderWidth,
@@ -45,16 +46,6 @@ const postControlsStyle = {
 
 /* eslint-disable react/prop-types */
 
-const renderCustomScheduledIcon = (isSent) =>
-  (<div style={postActionDetailsIconStyle}>
-    <ClockIcon color={isSent ? 'shuttleGray' : 'outerSpace'} />
-  </div>);
-
-const renderErrorIcon = () =>
-  (<div style={postActionDetailsIconStyle}>
-    <WarningIcon color={'torchRed'} />
-  </div>);
-
 const renderText = ({ postDetails }, hasError, isSent) =>
   (<span>
     <Text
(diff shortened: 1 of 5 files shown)


packages/shared-components/PostFooter/index.jsx
```diff
@@ -65,6 +56,16 @@ const renderText = ({ postDetails }, hasError, isSent) =>
     </Text>
   </span>);
 
+const renderIcon = (hasError, isSent, isCustomScheduled, isInstagramReminder) => {
+  if (!hasError && !isCustomScheduled && !isInstagramReminder) return;
+
+  return (<div style={postActionDetailsIconStyle}>
+    {hasError ? <WarningIcon color={'torchRed'} /> : null}
+    {isInstagramReminder && !hasError ? <CircleInstReminderIcon color={'instagram'} /> : null}
```
karelm: hey @anafilipadealmeida is this meant to be color={'instagram'} here? just checking 😄

packages/shared-components/PostFooter/index.jsx
```diff
@@ -83,10 +84,10 @@ const PostFooter = ({
 }) => {
   const hasError = postDetails.error && postDetails.error.length > 0;
   const isCustomScheduled = postDetails.isCustomScheduled;
+  const isInstagramReminder = postDetails.isInstagramReminder;
   return (<div style={isSent? sentPostDetailsStyle : getPostDetailsStyle(dragging)}>
     <div style={postActionDetailsStyle}>
-      {hasError && renderErrorIcon()}
-      {isCustomScheduled && !hasError && renderCustomScheduledIcon(isSent)}
+      {renderIcon(hasError, isSent, isCustomScheduled, isInstagramReminder)}
```
karelm: thats so much tidier! 👍

packages/shared-components/PostFooter/index.jsx
anafilipadealmeida: Yes 🤣 It's equivalent to #e53c5f, you can see the full list on `buffer-components/style/color.js` !

packages/utils/postParser.js
```diff
@@ -28,9 +28,14 @@ const getPostActionString = ({ post }) => {
     },
   );
 
-  if (post.scheduled_at) {
+  if (post.scheduled_at && post.can_send_direct !== false) {
```
karelm: does this need to use !== false as its kind of hard to read a double negative... can this be simplified with a !!post.can_send_direct ?

karelm: just an addition here `undefined !== false` and `null !== false`... just wanted to check that with you that you are looking for anything other than explicitly false here

anafilipadealmeida: I wanted this to run with any other values, just not with false. Sometimes this property is undefined, and I still wanted it to run in that situation.

PR #373, opened 2018-10-04
Attempt to fix grpc building errors

karelm:
### Purpose

grpc messaging for package.json


package.json
```diff
@@ -16,7 +16,8 @@
     "watch": "webpack-dev-server --config ./packages/server/webpack.config.dev.js --colors --display-error-details",
     "build": "webpack --config ./packages/server/webpack.config.prod.js --output-path=`pwd`",
     "updated": "lerna updated",
-    "rebuild-grpc": "docker exec -it bufferdev_publish_1 bash -c 'npm rebuild grpc --target_arch=x64 --target_platform=linux --target_libc=glibc'"
+    "rebuild-grpc": "docker exec -it bufferdev_publish_1 bash -c 'npm rebuild grpc --target_arch=x64 --target_platform=linux --target_libc=glibc'",
+    "preinstall": "MAGIC=`uname` && if [ $MAGIC != 'Linux' ]; then echo 'Please build dependencies from inside the container using: \n`./dev deps publish`' && exit 1; fi;"
```
hharnisc: Might be better to have a more descriptive name like DETECTED_OS or something similar. 

Looks like a snapshot failed here too

karelm: Certainly @hharnisc, I just thought it best to commit and push up while I remembered. I'll update all the things in the morning, and hopefully the snapshots will be sorted if I merge in master as I've not actually made any actual code changes... So would be a bit surprised if this broke some tests, but I've been wrong before 🙂

karelm: It looks like somehow master had failed tests checked in, I dont know how it got into that state exactly, but hopefully this will succeed

karelm:
I also dont think that the hint of `./dev rebuild publish` actually works for the grpc module rebuild either 😞

this does: `docker exec -it bufferdev_publish_1 bash -c 'npm rebuild grpc --target_arch=x64 --target_platform=linux --target_libc=glibc'`

hamstu:
Do you mean `./dev deps publish` @karelm? Yeah I don't expect that to work correctly until we upgrade the base image in the `Dockerfile` as @djfarrelly mentioned (since the current base image has a very outdated yarn). And yeah I do think we should do them in the same PR @djfarrelly :)

For what it's worth though, I just tried to do `./dev deps publish` locally as an experiment (after quickly updating the base images) and while `yarn` is now running and installing without errors, it's **_painfully_** slow. It probably takes 1-2 mins tops on my machine outside the container, but inside it took a whopping **11 minutes**. 😱

hamstu:
Gonna close this for now in lieu of other fixes in progress. 👍

PR #374, opened 2018-10-05
[PUB-709] Infinite scroll for loading posts

josem:
### Purpose
This PR removes the button to load more and instead uses the event of reaching the bottom of the list when scrolling down to load the next block (if there's one).

It also removes the 30 days limit for free users vs. not limit for Pro and business users from the frontend when loading sent posts since that's already handled on the backend so it was not needed.

### Review
Hey @hamstu! I'd love to have your eyes on this when you have some time :) 

### Note
The build is failing because I think the snapshots are outdated but I'm not sure how to update them from local. I've tried with `yarn run test-update` from root but nothing changes 🤔 

#### Staging Deployment
This repo is CI/CD enabled and staging deployment should be available at:
https://pub-709-sent-posts-30-days.publish.buffer.com


hamstu:
Ah, I see there's conflicts, so you will probably have to update the snapshots again after resolving them. Let me know as I'll still be connected throughout the conference, or if @anafilipadealmeida is online she could help too until we get your test update working right. 🙏 😄

josem:
Thanks so much @hamstu! Ana helped me to find that there was a problem with the Enzyme versions in Publish and the composer package so when you run the tests from Publish, the composer ones failed (eight of them exactly). 

I've upgraded Enzyme to the same version we use in composer so now we can run the tests well locally.

PR #379, opened 2018-10-10
[PUB-769] Display pinterest bar in queue with shared component PostMetaBar

albennett:
### Purpose

Display pinterest bar in queue with relevant information: 

- “Pinning to” phrase

- Thumbnail of Board

- Board name

- “Source” and URL

### Notes

Since the locationBar with Instagram is very similar to the pinterest bar, I created a shared component PostMetaBar. 
[PUB-769](https://buffer.atlassian.net/browse/PUB-769)

#### Staging Deployment


packages/shared-components/Post/index.jsx
```diff
@@ -76,6 +60,48 @@ const commentStyle = {
 
 /* eslint-disable react/prop-types */
 
+const stripUrlToDomainName = (sourceUrl) => {
```
albennett: Planning on writing some tests for this tmr

packages/shared-components/Post/index.jsx
```diff
@@ -76,6 +60,48 @@ const commentStyle = {
...
+  profileService,
+  dragging,
+  locationName,
+  sourceUrl,
+  subprofileID,
+  subprofiles,
+}) => {
+  let postMetaBarObj = null;
+  if (profileService === 'instagram' && locationName !== null) {
+    postMetaBarObj = { leftContent: { title: 'Location:', text: locationName }, dragging };
```
karelm: I don't think that dragging is needed here or in the other `postmetaBarObj`

packages/shared-components/Post/index.jsx
```diff
@@ -76,6 +60,48 @@ const commentStyle = {
...
+  /* the way we're handling sourceUrls, they shouldnt contain www., but just a safety check */
+  if (a.href.includes('www.')) a.href.replace('www.', '');
+  /* domainnames don't allow ".", so remove it & everything after it. With using split,
+  there's no issue if a "." isn't present in url  */
+  const domainName = a.hostname.split('.')[0];
+  const capitalizedDomainName = domainName.charAt(0).toUpperCase() + domainName.slice(1);
+  return capitalizedDomainName;
+};
+
+const renderPostMetaBar = ({
```
karelm: @albennett this is looking great! 

Could I perhaps suggest another approach that may be worth looking into. 
The suggestion I'm going to make is just checking for `profileService` with a switch/if statement.
Then returning an appropriate `<PinterestPostMetaBar>` / `<InstagramPostMetaBar>` / `null` based on scendario...
The benefit to this kind of approach is that you have cleaner separation in your code, and additionally you have an explicit list of dependencies attached to each branch of code, so you will know that pinterest has `locationName` as an explicit dependency, etc... otherwise this function gets full of decisions for all of the code, but actually it just needs to work out which `PostMetaBar` to render instead, and leave each bespoke `<profileService>PostMetaBar` to decide how best to render itself.

I hope that makes sense, and if you think its a good idea feel free to take it on, if not then thats fine too 😄 
Also, if you need me to clarify anything here, let me know and I'll gladly assist.

Otherwise, great work here 👍

karelm:
Hey @albennett I'm loving your work here. 
I've got a couple of suggestions for you, and a question or two in there too for you. I think perhaps consider them a bit and let me know whether you want to take them on or not. Either way is fine really.

Otherwise, great stuff. I look forward to seeing the tests for `stripUrlToDomainName`, and I think you may possibly need to update the tests on this branch, but otherwise I think its all great! 💯

packages/shared-components/Post/index.jsx
```diff
@@ -183,11 +196,14 @@ const Post = ({
           draggable,
           dragging,
         })}
-        {profileService === 'instagram' && locationName != null &&
-        <LocationBar
-          locationName={locationName}
-          dragging={dragging}
-        />}
+        {renderPostMetaBar({
```
karelm: essentially renderPostMetaBar is already a functional component...

It takes in a props object, and renders a component or null...

` fn(props) => dostuff && <component> || null `

So I think if we want to be full on React here, it would be great to rename the function from `renderPostMetaBar` to `RenderPostMetaBar` then you can call the function as a component

`<RenderPostMetaBar profileService={profileService} ... />`
or 
`<RenderPostMetaBar { ...{profileService, dragging, ...}} />`

This way it makes it more like the "React Way" (tm) and later if we want to pull out the component, it should just be simple to hoist it out of this file and into a standalone shareable component.

If you think this is a good idea, then go for it, if not, then thats fine too... this is great either way, and both are right

packages/shared-components/Post/index.jsx
karelm: I think its worth mentioning it here what we discussed on the call, domains like images.google.com will show images, and bbc.co.uk will show bbc, so i think its probably worth just removing the www but if it exists, but keeping the rest. Maybe take a look at how the buffer-web codebase does this, and see whether theres a property there, or whether its just the sourceUrl being output directly. 

Perhaps this function can be removed if its not necessary, but may be worth bringing up with Dave and Super if this changes the vision set out in the design.

I hope this is of use, even if just for recollection later 😄

packages/shared-components/Post/index.jsx
albennett: ah you're correct! thank you!

packages/shared-components/PostMetaBar/index.jsx
```diff
@@ -0,0 +1,85 @@
...
+    <Text
+      size={'small'}
+      color={'black'}
+    >
+      {leftContent.title}
+    </Text>
+    <div style={getImageWrapperStyle(leftContent.avatarUrl)}>
+      { leftContent.avatarUrl &&
+        <Image
+          height="1.3rem"
```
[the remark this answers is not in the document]
albennett: Hey @karelm , great suggestion. I looked as well, and didn't see anything. I'm going to add it to my "tasks in-between cycles" list to refactor components styles with this and a few more possible variables I've come across in publish. At the moment, I'm going to keep as is. :)

packages/shared-components/Post/RenderPostMetaBar/index.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+const RenderPostMetaBar = ({
+  profileService,
+  locationName,
+  sourceUrl,
+  subprofileID,
+  subprofiles,
+  dragging,
+}) => {
+  if (profileService === 'instagram' && locationName !== null) {
```
karelm: If I've got one last minor thing to add here

You've put the if condition about locationName outside the InstagramPostMetaBar, but I think this is the type of logic that could potentially live inside the InstagramPostMetaBar and any other `[service]PostMetaBar` components, because I think that it should be the component itself deciding whether it should render or not. 
There are exceptions to this too, but I think the important part as I see it is around the profileService that triggers the rendering, once we know what profileService we are in, its up to the component to be in charge of its rendering choices.
I may well be wrong here, but I think this gives the best form of encapsulation, as you are dealing with your own scenarios yourself. Then it allows you to test what happens when you provide null / undefined inputs, and complete inputs to a component.

I definitely don't mind if this gets solved at another juncture at all. I think its just good to share this for informational purposes.

karelm:
@albennett I'm absolutely loving this, its so much cleaner and clearer about where dependencies lie, what data each component requires and how its being handled. 
I've added a comment here: https://github.com/bufferapp/buffer-publish/pull/379/files#r225204430 but I think this is totally fine to get out in its present state! 👍 

Great work 💯

packages/shared-components/Post/RenderPostMetaBar/index.jsx
albennett: That makes sense, thanks @karelm 😄

PR #380, opened 2018-10-12
Feature/pub 784 general tab

lolaodelola:
### Purpose
Display the secondary nav under 'Settings' including 'General' & 'Posting Schedule'. Ensure that the correct tab content & url shows when the tab is selected.

### Notes
- As mentioned in the jira ticket, although this particular thing is finished, I don't think it's ready to be merged into master as we definitely don't want users to see it like this. I thought about adding a feature flip but this isn't a full feature. I propose that this becomes the Profile Settings' (epic on Jira) master branch, where the other features related to this will branch from (& merged into) & once enough of the work is completed merge everything back into real master. Reviews will still happen on feature branches but it does mean that the final merge back into real master will be chunky. If anyone has any other alternatives or preferences, please do let me know.

- When the settings tab is selected, the "General" tab content is the default however, the url is still `buffer.com/profile/:profile_id/tab/settings` instead of `buffer.com/profile/:profile_id/tab/settings/general` does this matter? Both urls lead to the same place.

- There are places I've essentially just repeated code, particularly in the routes file, & the only difference being the argument of `childTabId`, when I've tried to incorporate `childTabId` in existing route generating functions, something breaks. This works for now so isn't a real priority, however, a refactor could be just having one function, that handles the option of `childTabId`. Alternatively, this is more readable, & concerns itself with only one thing.

EDIT:
- I had 100% passing tests (see commit: [65f8049](https://github.com/bufferapp/buffer-publish/pull/380/commits/65f8049fcd4e6aab25865a3212fc832189db8f4a)), prompted me to resolve a minor conflict about what component to display, then 13 tests failed 😨so I got rid of that merge, instead i'll run a rebase.

- The rebase worked better because it allowed me to investigate the offending conflict, all passing now 🙌 

#### Staging Deployment


Dockerfile
```diff
@@ -13,4 +13,4 @@ COPY staticAssets.json /usr/src/app
 
 EXPOSE 80
 
-CMD node ./index.js
+CMD node ./index.jsx
```
karelm: Hey @lolaodelola,

I've had a look at the dockerfile and I _think_ that its copying assets from the `/packages/server/` directory and then running `node` against `index.js` from that directory. I may be wrong but I'm not sure this is quite right in this dockerfile or the development dockerfile either... With that being said, I'd be happy to know if you know different and can share some context on this 😃

lolaodelola: Hey Karel, so this was auto generated & i'm not exactly sure how/why so I can do some investigation. Maybe I'm using the wrong prefix for one of the new files.

karelm:
hey @lolaodelola I think something has definitely gone a bit awry with your references... as lots of things appear like they were originally pointed at lerna packages now have relative imports in their place... looks like your IDE is causing some additional pains for you there. changing bufferapp/<module> to ../../<module> in lots of places

karelm:
Hi @lolaodelola, 
I've finally got to the end of the files in this PR + I can see that this PR has some small things that could do with sorting... but the truth is that this is a Brilliant bit of work!!! 🎉 

I know that React is not something you have mastered, and thats nothing to worry about, its a learning journey for everyone on the team, but its a fantastic first effort for this. Even better it works when running it locally, so thats a superb effort. 💯 🥇 

If you are interested we could perhaps do some pair programming next week to go over the bits and pieces here that could do with some attention and address them together? That way I can talk to you about my reasoning and perhaps cover what each thing is doing in a more meaningful way.

I've created a PR + closed it down with a fast set of changes I think may be worth applying to your PR: https://github.com/bufferapp/buffer-publish/pull/382 
I think I could show you how to deal with some of the reverts in a fast (+ simple-ish) way, similar to how I've made changes there myself. It wont clean up all the issues you've faced, but it took me somewhere between 10 and 30 minutes to do a quick cleanup job on the code there and it very rapidly addresses the problems that are mainly impacting your work.

I hope thats okay, let me know if that sounds good to you + we can work through those together

lolaodelola:
That would be great @karelm :D

Dockerfile
lolaodelola: Hey folks, so I just realised it was my PHPStorm being clever, everytime I changed the file name it changed it everywhere 🤦‍♀️

lolaodelola:
Thanks for your reviews @karelm @hamstu @hharnisc 😄😄😄 

I'll come back to this PR once my second PR enters the review stage. The file changes & linting is stuff is PHPStorm trying to be clever, I think there are some presets/settings I need to amend.

lolaodelola:
Closing this branch as there's another branch with more up-to-date changes: https://github.com/bufferapp/buffer-publish/pull/385

PR #381, opened 2018-10-12
[PUB-807] Update tooltip's limit of accounts

anafilipadealmeida:
### Purpose
Fix tooltip to display number of accounts allowed.

### Notes
Task: [PUB-807](https://buffer.atlassian.net/browse/PUB-807)

### DIFF
--- a/packages/profile-sidebar/components/ProfileSidebar/index.jsx
+++ b/packages/profile-sidebar/components/ProfileSidebar/index.jsx
@@ -52,7 +52,7 @@ const buttonDividerStyle = {
   marginBottom: '1rem',
 };
 
-const renderLockedHeader = translations => (
+const renderLockedHeader = ({translations, profileLimit}) => (
   <div style={lockedAccountHeaderStyle}>
     <Text size={'small'}>
       {translations.lockedList}
@@ -67,7 +67,9 @@ const renderLockedHeader = translations => (
         label={translations.lockedList}
       >
         <div style={{ padding: '.5rem .25rem' }}>
-          {translations.lockedListTooltip}
+          {translations.lockedListTooltip1}
+          {profileLimit}
+          {translations.lockedListTooltip2}
         </div>
       </IconArrowPopover>
     </div>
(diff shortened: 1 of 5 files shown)


__snapshots__/snapshot.test.js.snap
```diff
@@ -61513,7 +61513,7 @@ exports[`Snapshots ProfileSidebar should display a list of profiles 1`] = `
                   }
                 }
               >
-                Sorry, the free plan lets you access up to two social accounts at the same time (and any business accounts you’re a team member on). We’ll keep these other ones safe and sound until you’re ready to upgrade!
+                Sorry, your current plan lets you access up to 3 social accounts at the same time (and any business accounts you’re a team member on as long as the owner is on a Business Plan). We’ll keep these other ones safe and sound until you’re ready to upgrade!
```
anafilipadealmeida: Added a "as long as the owner is on a Business Plan" to make it clear for the user. Imagining a free user (limit: 3 social accounts) that has 2 social accounts and is a team member of a business account: If the owner of the business plan downgrades to free, this user would until now see 1 blocked account and not know at first why there's a blocked account if he didn't reach the limit.

packages/profile-sidebar/components/ProfileSidebar/index.jsx
```diff
@@ -52,22 +52,22 @@ const buttonDividerStyle = {
   marginBottom: '1rem',
 };
 
-const renderLockedHeader = translations => (
+const renderLockedHeader = ({translations, profileLimit}) => (
   <div style={lockedAccountHeaderStyle}>
     <Text size={'small'}>
       {translations.lockedList}
     </Text>
     <div style={{ position: 'absolute', marginLeft: '13rem' }}>
```
anafilipadealmeida: I discovered that there's a position issue with the tooltip: if you have multiple accounts blocked and you have to scroll to see them, the tooltip also moves. I've tried a couple of different things to solve it, but none was a perfect solution. My suggestion would be to add a new Jira ticket for it, so I can revisit it later or we can have a fresh pair of eyes on this.

josem: I agree @anafilipadealmeida! Since the goal of this change was to include Pro we can create a new task for this with screenshots. Would you be up for this when you have some time? I think we could mention there Super to prioritize it.

PR #386, opened 2018-10-15
[PUB-822] Add year to queue when post is scheduled in future year

anafilipadealmeida:
### Purpose
Add year to queue when post is scheduled in future year.

### Notes
Task: [PUB-822](https://buffer.atlassian.net/browse/PUB-822)


hamstu:
Great fix @anafilipadealmeida! Not to diminish the awesome fix you made here, but this feels like something we _might_ want to handle on the backend instead. (We generate the `post.day` field in the backend.)

Not saying it should be that way per-se, but it would simplify the front-end a bit, and I'm all for doing that wherever possible! 😄

hamstu:
Oops, I didn't mean to close this! So sorry (reopening it.)

anafilipadealmeida:
@hamstu I haven't changed the post.day in the backend because that string (that doesn't include the year) is being used in a lot of places in that format. Since we want to keep the dashboard without clutter, the only circumstance where we want to see the year is in the queue's day title of a future year (not even in the post footer on each post) so I thought this approach would be the cleanest. But would love to hear any thoughts or different approaches to it! 😄

hamstu:
I think I stalled this one a bit @anafilipadealmeida! Shall we pair on investigating the backend option here?

PR #399, opened 2018-10-23
[PUB-865] Add composer to drafts tab

anafilipadealmeida:
### Purpose
Add composer to Drafts tab

### Notes
Task: [PUB-865](https://buffer.atlassian.net/browse/PUB-865)
Adding now a temporary ComposerPopover and a ComposerWrapper for drafts just so it works. My next task is gonna be to create a new package for the composer to be used by drafts, queue and sent packages (that's why I'm already passing a prop type in this PR, to start this kind of migration process).


anafilipadealmeida:
I think I found an old bug with composers in different tabs: if you submit a post and move to the drafts tab you can't submit the draft (and vice versa). I'm merging this and creating a ticket for it!

PR #400, opened 2018-10-23
[WIP] Fix `grpc` binary errors by separating server package

hamstu:
### Purpose

Proposal to fix issues with `grpc` binary errors in publish by separating server from other packages.

More details: https://paper.dropbox.com/doc/RFC-Proposal-for-fixing-grpc-binary-errors--AP8Ok86ht7Ve1XrC2UVH5UBXAg-iBYHmT3yCsyR1iX6MOCeS

### Notes

cc @msanroman as I think this approach would be good for Analyze as well.

### Review

Pending changes from @hharnisc, will rebase from master and merge this in.

#### Staging Deployment


Dockerfile
```diff
@@ -1,14 +1,14 @@
-FROM node:8.1.2-alpine
+FROM node:8.12.0-alpine
```
hamstu: Updated to latest image. ✨

Dockerfile.development
```diff
@@ -1,4 +1,4 @@
-FROM node:8.1.2-slim
+FROM node:8.12.0-alpine
```
hamstu: Match version in prod.

Dockerfile.development
```diff
@@ -8,5 +8,5 @@ RUN npm install -g nodemon
 
 EXPOSE 80
 
-CMD nodemon -V -L --ignore 'index.html' --watch ./packages/server -e html,js,json packages/server/index.js
+CMD nodemon -V -L --ignore 'index.html' --watch ./ -e html,js,json index.js
```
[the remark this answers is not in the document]
hamstu: Good one! 👍

hamstu:
Thanks for the review @djfarrelly! Great thoughts around all the changes for devs. This shouldn't change the dev setup at all though. As I mentioned on the other PR we're already running yarn and bundling outside the container. For new devs, the `./dev setup` command will run all the `./dev deps` and get the server working, so the only real change is knowing to run `./dev deps publish` again if you change the server's package.json, which isn't something I expect we'll do very often. Does that make sense?

Re: Analyze; I talked to @msanroman today about this a bit, as I think being consistent with Analyze makes a lot of sense (and James had the `grpc` issue just the other day...) Would love to know what he thinks.

If / when we merge something like this I'd pair it with a very clear announcement and instructions so that the team knows how to update their code/branches. 

I know @hharnisc wanted to mull this one over, so keen to hear any of this thoughts as well. I'm happy to be challenged or hear other ideas on any of this. This is just one way I could see it working well for the team.

hharnisc:
@hamstu @djfarrelly have had some time to think about this approach. While this would fix the problem short term for Publish, I think fixing it in https://github.com/bufferapp/buffer-js-buffermetrics is going to be the better medium to long term approach (long term approach would be native gRPC library).

This would fix it for Publish, Analyze, Account (and any other project we decide to use this library in) rather than have to do this for each application for both the application and the `buffer-dev` configuration.

This approach defined in this PR also means 2 docker containers need to run with 2 different initialization processes (sure one command could run both, but there are still 2 things to care about and debug). We'll have to do this for each application as well.

So generally speaking I think we should hold off on this approach @hamstu until we show that we can't modify https://github.com/bufferapp/buffer-js-buffermetrics for some reason.

karamcnair:
@hharnisc - I love the idea of fixing this in one place but I'm unclear on who actually owns and makes changes to buffermetrics, and if that's the right place to fix it, who does that and how does it get scheduled? Is that something that's clear to others? If there's an immediate tradeoff between 'this makes things better but not perfect right now' and 'someday it will be much better', I'd love to see if we can blend those to 'this makes things much better' :) or at least improve things in the way we know.

If this change wouldn't require more work in a future change to revert this code, it doesn't feel like it's bad to put this in, does it?

@djfarrelly I think Hamish mentioned that he planned to communicate this in a lot of detail, and since this is something folks are tripping over regularly in its current state, I wonder if moving ahead with this makes sense anyway. What do you think?

PR #409, opened 2018-10-26
Task/add link shortener

karelm:
### Purpose

Implementing Link Shortener selection screen options in general settings tab

(still need to plumb in the connect bit.ly button)

### Notes

JIRA:
- https://buffer.atlassian.net/browse/PUB-790
- https://buffer.atlassian.net/browse/PUB-791


packages/general-settings/components/LinkShortening/index.jsx
```diff
@@ -0,0 +1,100 @@
...
+      >
+        <div style={textWrapperStyle}>
+          <div
+            style={{
+              padding: '10px',
+              background: '#efefef',
+              borderRadius: '6px',
+              backgroundClip: 'padding-box',
+              width: '100%',
+            }}
```
lolaodelola: Should these rules be abstracted into their own const?

packages/general-settings/components/LinkShortening/LinkShorteningWrapper.jsx
```diff
@@ -0,0 +1,129 @@
...
+  return (
+    <div
+      style={{
+        display: 'flex',
+        flex: '1 1 0%',
+        flexDirection: 'row',
+        marginBottom: '0.5rem',
+        width: '100%',
+        alignItems: 'center',
+      }}
```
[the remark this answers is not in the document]
karelm: 👍 Totally, missed this one!!! thanks @lolaodelola

packages/general-settings/middleware.js
```diff
@@ -1,12 +1,34 @@
...
   switch (action.type) {
     case actionTypes.SET_DIRECT_POSTING:
       window.location = getURL.getInstagramDirectPostingURL(action.profileId);
       break;
+    case profileActionTypes.SELECT_PROFILE:
+      dispatch(dataFetchActions.fetch({
+        name: 'getLinkShortener',
+        args: {
+          profileId: action.profile.id,
+        },
```
lolaodelola: Can you get the `profileId` directly out of action here, like:
```suggestion
        profileId: action.profileId
```

packages/general-settings/middleware.js
```diff
@@ -1,12 +1,34 @@
...
   next(action);
   switch (action.type) {
     case actionTypes.SET_DIRECT_POSTING:
       window.location = getURL.getInstagramDirectPostingURL(action.profileId);
       break;
+    case profileActionTypes.SELECT_PROFILE:
+      dispatch(dataFetchActions.fetch({
+        name: 'getLinkShortener',
+        args: {
+          profileId: action.profile.id,
```
lolaodelola: Can you get the `profileId` directly out of action here, like:
```suggestion
        profileId: action.profileId
```

lolaodelola:
This looks good to me, Karel! Well done 💯

packages/general-settings/middleware.js
karelm: Hey Lola, great question here, I had a look and I think that since this is triggering from the SELECT_PROFILE action, it looks like this is the place in the data that the id is stored, and seems consistent with other places in our codebase where this is being accessed... with that being said, its quite a difficult challenge to work out where to source data as there is no definition of the data objects we are seeing, so it makes it quite challenging to find the right piece of data here, so thankyou for mentioning it, as it meant i could double check 👍

PR #414, opened 2018-10-30
Fix default element when loading the app for the first time

josem:
### Purpose
Fix https://buffer.atlassian.net/browse/PUB-735 by changing the default element that it's loaded and sending an event when it's rendered so we prevent the blank page when going back to https://publish.buffer.com

(Please, see the ticket for more context)

### Review
Hey @karelm! I'd love to have your eyes on this :) 

#### Staging Deployment
https://fix-flash-screen-loading.publish.buffer.com


packages/profile-sidebar/middleware.js
```diff
@@ -19,35 +23,49 @@ import {
 export default ({ dispatch, getState }) => next => (action) => {
   next(action);
   switch (action.type) {
-    case 'APP_INIT':
+    case 'APP_INIT': {
       dispatch(dataFetchActions.fetch({
         name: 'profiles',
       }));
+      dispatch({ type: initialLoadingActionTypes.PROFILE_LOADING_REDIRECT });
```
karelm: Hey @josem, I'm just wondering why this action is necessary? should the default component not trigger this event itself when it loads?

packages/profile-sidebar/middleware.js
```diff
@@ -19,35 +23,49 @@ import {
...
       break;
+    }
     case `profiles_${dataFetchActionTypes.FETCH_SUCCESS}`: {
+      dispatch({ type: initialLoadingActionTypes.PROFILE_LOADING_REDIRECT });
```
karelm: I think you should just be able to remove lines 34, 35, and 36 here and just fall straight through into the `case initialLoadingActionTypes.PROFILE_LOADING_REDIRECT` as thats basically what you are doing anyway, and it would save you from adding an extra step in the call stack? 

I may be wrong here, just wondered if that would be a better solution at all?

karelm:
Hey @josem I've added a couple of questions/comments, but I think this looks good overall!! 💪 

I think its good to 🚀 but if you think the comments warrant some added TLC, feel free to make those changes. Otherwise LGTM 👍

josem:
Thanks so much @karelm for your review! Great comments 💯 

I've made those adjustments and will deploy this now :)

PR #415, opened 2018-10-30
Feature/pub 739 ig banner

lolaodelola:
### Purpose
Display IG banner if direct posting is not enabled

### Notes
screenshot 2018-10-29 at 17 49 49

### Review
Hey @albennett @anafilipadealmeida @hamstu 😄 
#### Staging Deployment

### DIFF
--- a/packages/queue/components/InstagramDirectPostingBanner/index.jsx
+++ b/packages/queue/components/InstagramDirectPostingBanner/index.jsx
@@ -1,48 +1,52 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
-  Button, Divider,
-  Text, Link,
+  Text, Link, CircleInstagramIcon,
 } from '@bufferapp/components';
-
-const instagramDirectPostingStyle = {
-  display: 'flex',
-  flex: 1,
-  flexDirection: 'row',
-  marginBottom: '0.5rem',
-};
 
 const textWrapperStyle = {
   display: 'flex',
-  marginBottom: '0.5rem',
+  marginLeft: '0.3rem',
 };
 
-const setUpDirectPostingStyle = {
-  marginBottom: '1.5rem',
-  marginTop: '1rem',
-  textAlign: 'right',
-  whiteSpace: 'nowrap',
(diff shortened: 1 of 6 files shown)


packages/queue/components/InstagramDirectPostingBanner/index.jsx
```diff
@@ -0,0 +1,57 @@
...
+  display: 'flex',
+  marginLeft: '0.3rem',
+};
+
+const linkWrapperStyle = {
+  marginLeft: '0.3rem',
+};
+
+const bannerWrapper = {
+  border: '1px solid #CAD4DB',
```
albennett: If you haven't checked yet, I wonder if any of these colors/styles are in the buffer-components style folder.

packages/queue/middleware.js
```diff
@@ -6,6 +6,7 @@ import {
 } from '@bufferapp/async-data-fetch';
 import { actions as notificationActions } from '@bufferapp/notifications';
 import { actionTypes, actions } from './reducer';
+import {getURL} from "@bufferapp/publish-formatters";
```
hamstu: Small style thing, just spacing around the braces! 💅 

```suggestion
import { getURL } from "@bufferapp/publish-formatters";
```

packages/queue/reducer.js
```diff
@@ -380,6 +382,12 @@ const profileReducer = (state = profileInitialState, action) => {
         ...state,
         numberOfPostsByDate: action.result.numberOfPostsByDate,
       };
+    case actionTypes.SET_DIRECT_POSTING:
+      return {
+        ...state,
+        directPostingEnabled: action.profile.directPostingEnabled,
+        profileId: action.profileId,
+      };
```
hamstu: I'm not sure we need this `case` statement, since the only thing that should happen when you click the button is already handled in the middleware (opening the URL), so this code has no effect and isn't needed 😄 

Let me know if that makes sense!

lolaodelola: This was a copy & paste error on my part, thanks for spotting!

PR #433, opened 2018-11-08
[PUB-888] Add source tracking to upgrade modals

hamstu:
### Purpose

Adds source tracking to the upgrade modal so we can get a better idea of where upgrades are coming from.

https://buffer.atlassian.net/browse/PUB-888

#### Staging Deployment

### DIFF
--- a/packages/modals/middleware.js
+++ b/packages/modals/middleware.js
@@ -1,28 +1,13 @@
 import { actionTypes } from '@bufferapp/async-data-fetch';
 import { actions } from './reducer';
-
-const checkHashShowUpgradeModal = () => {
-  const hash = window.location.hash;
-  let showUpgradeModal = false;
-  let source = null;
-  if (hash.indexOf('#upgrade-to-pro') === 0) {
-    showUpgradeModal = true;
-    // #show-upgrade-modal--profile_limit => 'profile_limit'
-    source = hash.split('--')[1] || 'unknown';
-  }
-  return {
-    showUpgradeModal,
-    source,
-  };
-};
+import { shouldShowModal, getSourceFromHash } from './util/window-hash';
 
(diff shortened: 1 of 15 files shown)


packages/modals/middleware.js
```diff
@@ -1,16 +1,34 @@
...
+};
+
 export default ({ dispatch }) => next => (action) => {
   next(action);
   switch (action.type) {
-    case 'APP_INIT':
-      if (window.location.hash === '#upgrade-to-pro') {
-        dispatch(actions.showUpgradeModal());
+    case `user_${actionTypes.FETCH_SUCCESS}`: {
+      const { showUpgradeModal, source } = checkHashShowUpgradeModal();
```
josem: Maybe we could have a different name for this method? ```checkHashShowUpgradeModal``` doesn't sound very specific. Not sure if it's better to have a method like ```shouldShowUpgradeModal()``` and then another one to get the source like ```getSource()```. In that fashion we could do something like: 

```js
const shouldShowModal = () => {
  return window.location.hash.indexOf('#upgrade-to-pro') === 0);
};

const getSource = () => {
  const hash = window.location.hash;

  let source = null;
  if (hash.indexOf('#upgrade-to-pro') === 0) {
    source = hash.split('--')[1] || 'unknown';
  }

  return source;
};
```

So after that refactoring maybe it makes sense to extract the "#upgrade-to-pro" since it's repeated. 

Happy to chat more about this :)

josem: Or even "getSourceFromHash" I think 🤔

hamstu: Great one @josem! I've changed it. Let me know what you think 😄

josem: Looks great @hamstu! 🚀

PR #439, opened 2018-11-16
Add Drafts actions

anafilipadealmeida:
### Purpose
Add drafts footer actions


anafilipadealmeida:
hey @hamstu, I'd like to merge this branch since it has been opened for so long. I still want to add a couple of FETCH_SUCCESS actions to the reducer and pusher events, as well as tests, but I'd prefer to do it in smaller branches 😄

anafilipadealmeida:
👍 Here it is! I wasn't connecting it because it doesn't feel complete until the Pusher events are done, but I guess I can open more than one PR for the same card 🤦‍♀️

PR #441, opened 2018-11-18
[PUB-901] Add datepicker component for b4b analytics

albennett:
### Purpose
Add datepicker (from Analyze) to Analytics toolbar.
### Notes
I had to carry over a lot of files and folders (esp. rpc) from analyze to get DatePicker working correctly. Apologies for the large PR! 

I also was waiting on a design spec for header/tabs when I began, which is why I started with toolbar. 

The analytics package isn't getting imported anywhere, so no one should be able to see it :) 

Will need to further test datepicker when more components are being imported.

https://paper.dropbox.com/doc/Analyze-Components-in-New-Publish--ASK4cTWL0aaixNol6_hzybIMAg-Mm9abUJq9wcC7IgjFt6V0
#### Staging Deployment


package.json
```diff
@@ -110,7 +112,7 @@
   },
   "jest": {
     "transformIgnorePatterns": [
-      "/node_modules(?!/@bufferapp/performance-tracking)(?!/@bufferapp/async-data-fetch)(?!/@bufferapp/components)(?!/@bufferapp/web-components)(?!/@bufferapp/composer)(?!\/@bufferapp\/unauthorized-redirect)/"
+      "/node_modules(?!/@bufferapp/performance-tracking)(?!/@bufferapp/notifications)(?!/@bufferapp/analyze-date-picker)(?!/@bufferapp/analyze-profile-selector)(?!/@bufferapp/analyze-shared-components)(?!/@bufferapp/report-list)(?!/@bufferapp/async-data-fetch)(?!/@bufferapp/components)(?!/@bufferapp/web-components)(?!/@bufferapp/composer)(?!\/@bufferapp\/unauthorized-redirect)/"
```
albennett: Had to add these to get rpc analyze tests passing. Wish we could get use the analyze config variables in json!

packages/server/rpc/analytics/getReport/index.test.js
```diff
@@ -0,0 +1,229 @@
+/* eslint-disable import/first */
+import moment from 'moment-timezone';
```
albennett: I had to change this import from moment to moment-timezone because of failing tests. It seems moment package doesn't have all the moment.tz logic that moment-timezone has? The index.js file uses moment-timezone instead of moment. Curious whether this test is passing in analyze. 

This is the only change I made from the rpc folders I copied from analyze.

package.json
hamstu: Thanks for adding these! Agreed re: using the config. It looks like we actually can by moving this `jest` config out of the `package.json` and into a `jest.config.js` file. ([Docs here.](https://jestjs.io/docs/en/configuration.html))

packages/analytics/components/Toolbar/index.jsx
```diff
@@ -0,0 +1,21 @@
...
+import DatePicker from '@bufferapp/analyze-date-picker';
+
+const toolbarDatePicker = {
+  display: 'flex',
+  justifyContent: 'flex-end',
+  marginLeft: 'auto',
+};
+
+const Toolbar = () => (
+  <div>
```
hamstu: Since this `div` doesn't have any styling and doesn't have multiple direct children I think we can just 🗑 it. What do you think?

packages/analytics/index.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import Toolbar from './components/Toolbar';
+
+const Analytics = () => (
+  <div>
```
hamstu: Same as the previous `div` comment. 😄

packages/analytics/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@bufferapp/publish-analytics",
+  "version": "1.0.0",
+  "description": "",
```
hamstu: Might be good to put just a small description here.

packages/profile-sidebar/middleware.js
```diff
@@ -20,6 +20,18 @@ import {
   actionTypes,
 } from './reducer';
 
+// mirror format used in analyze
+const formatAnalyticsProfileObj = ({ id, service, timezone, service_username, avatarUrl }) => (
```
hamstu: Since this looks like it's _mostly_ the same data and just aliases the one field and adds the `organizationId`, I wonder if we could [add what's needed to the `profileParser`](https://github.com/bufferapp/buffer-js-publish-parsers/blob/master/src/profileParser.js) instead?

Not blocking for merge, since that's another repo, but maybe once that's done we could come back and remove this?

packages/server/webpack.config.common.js
```diff
@@ -2,6 +2,8 @@ const webpack = require('webpack');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 
 const vendor = ['react', 'react-dom', '@bufferapp/components'];
+const { analyzePackagesWhitelist, analyzeLessLoader } = require('./analyze.config.js');
```
hamstu: # 😻

packages/server/rpc/analytics/getReport/index.test.js
hamstu: > It seems moment package doesn't have all the moment.tz logic that moment-timezone has? 

Yep, exactly. The timezone data adds a huge chunk of extra bytes to the file size, so `moment` doesn't include it in the normal package.

Not sure how it's working in analyze though, maybe something in the way they've configured their tests is bypassing this. 🤔

packages/analytics/components/Toolbar/index.jsx
albennett: Agreed! I might need to add styling to the div when I start adding more to the toolbar, but I'll 🗑 it for now :)

packages/analytics/package.json
albennett: Ah yes, thank you for catching that!

packages/profile-sidebar/middleware.js
albennett: Good point! I'll merge as is for the moment, and implement to profileParser next week!

package.json
albennett: I've ran into some errors with the jest refactor. Since I'm tight on time today w/ traveling, I'm going to break this refactor into it's own PR and plan to get it merged next week.

PR #444, opened 2018-11-20
Add notifications with GET parameters

josem:
### Purpose
Add notifications with GET parameters to show in the dashboard.

The notifications are taken from the notifications.json file, inside the notifications-provider package and they can also include a substitution of a variable.  See the [index.test.js in notifications-provider](https://github.com/bufferapp/buffer-publish/blob/c09af6205607c0d07140d7371f0e16767c73686c/packages/notifications-provider/index.js) for examples.

To pass the GET parameters, we can use nt, nk and nv. See [this readme](https://github.com/bufferapp/buffer-publish/blob/c09af6205607c0d07140d7371f0e16767c73686c/packages/server/README.md).

This URL: 
https://publish.buffer.com/?nt=success&nk=example-with-variable&nv=example

Will generate this notification: 

image

### Notes
- I don't sanitize the GET param, not sure if it's needed here, what do you think? 
- I've triggered the notification in the profile sidebar to show it a little bit later so the user has more time to see it.

#### Staging Deployment


karelm:
Hey @josem this looks good to me.

I don't think you need to do any sanitization of the get parameters, because express parser middleware already does that for you, so should be fine. 
I think I'm in agreement with you that it's a good idea to delay displaying the notification, but I'm wondering whether it may be beneficial to just add a delay to display if the notifications come via query params, and then make a component at the root level of the site, as its not really related to the profile-sidebar? That being said, the profile sidebar does seem to have other logic in it, so it seems like perhaps that component is mis-labelled more than anything else really 🤷‍♂️ ?

Otherwise its looking good 👍

hamstu:
:ceilingcat: Love this @josem! 

I agree with @karelm re: location. The `profile-sidebar` is a but overloaded perhaps, and so what you might consider doing is exporting another middleware from your new `notification-provider` package and handling the dispatching there. Does that make sense?

josem:
Thank so much for the review @karelm and @hamstu! Great suggestion! :)

I've just updated the PR with the suggested change, let me know what you think please. Happy to adjust it more.

PR #446, opened 2018-11-21
[Hotfix] Fix discount percentage shown in upgrade modal

hamstu:
### Purpose

Quick fix for a bug spotted by @jwinternheimer 😄 

image

Thread: https://buffer.slack.com/archives/C151RRDL1/p1542743078521500

#### Staging Deployment


.storybook/preview-head.html
```diff
@@ -1,4 +1,4 @@
-<link href="https://fonts.googleapis.com/css?family=Open+Sans:300,400,600,700,800" rel="stylesheet">
```
hamstu: Also just fixing the font here (this is for Storybook only). 💅

PR #447, opened 2018-11-21
Update moment and buffermetrics packages

albennett:
### Purpose
Saw a `We found a potential security vulnerability in one of your dependencies.` alert in buffer-publish repo for moment package. Updated moment package versions. 
### Notes
When I ran yarn, I also got an error `@bufferapp/buffermetrics@0.6.3-beta-06: The engine "node" is incompatible with this module.` I noticed the package version was updated recently, so I updated the versions in publish and the error went away. Was there a specific reason we were using 0.6.3-beta-06?

#### Staging Deployment


hamstu:
👍 To the moment change.

Re; the buffermetrics notice. Not sure. Gonna ping @hharnisc on this one since it smells like something related to his final work on the grpc bugs we were seeing from that library. (I could be wrong on that.)  So let's wait for his input before we merge this (I can take care of it if you're away @albennett).

packages/modals/package.json
```diff
@@ -11,7 +11,7 @@
   "dependencies": {
     "@bufferapp/keywrapper": "0.2.0",
     "@bufferapp/publish-upgrade-modal": "2.0.0",
-    "@bufferapp/buffermetrics": "0.5.2"
+    "@bufferapp/buffermetrics": "0.6.3"
```
hharnisc: Can you use `0.6.3-beta-06` for this :smile: should have everything from the `0.6.3` package and use native js driver

packages/server/package.json
```diff
@@ -9,7 +9,7 @@
   "author": "hharnisc@gmail.com",
   "dependencies": {
     "@bufferapp/buffer-rpc": "0.0.4",
-    "@bufferapp/buffermetrics": "0.6.3-beta-06",
+    "@bufferapp/buffermetrics": "0.6.3",
```
hharnisc: Can you use `0.6.3-beta-06` for this :smile: should have everything from the `0.6.3` package and use native js driver

packages/store/package.json
```diff
@@ -12,7 +12,7 @@
     "@bufferapp/analyze-profile-selector": "0.65.0",
     "@bufferapp/app-sidebar": "1.9.24",
     "@bufferapp/async-data-fetch": "1.9.4",
-    "@bufferapp/buffermetrics": "0.6.3-beta-03",
+    "@bufferapp/buffermetrics": "0.6.3",
```
hharnisc: same :smile_cat:

hharnisc:
Thanks for the ping @hamstu we should be able to use `0.6.3-beta-06` for this (for now)

packages/store/package.json
albennett: Thanks! Upgrading my node version worked :)

PR #450, opened 2018-11-26
[PUB-917] Add summary-table component to analytics

albennett:
### Purpose
Add performance summary component to Analytics package

[PUB-917](https://buffer.atlassian.net/browse/PUB-917)
[Scoping Analytics doc](https://paper.dropbox.com/doc/Scoping-Analytics-for-B4B-in-New-Publish--ASky_9j9dl4PN0FzF30FuvPiAg-788eoDDmJ1QTHGzQt4vAF)

### Review
I'm currently styling it similar to Analyze. This could change after Dave reviews it. I added a css file to hide the "Add to Report" button.  

The datepicker is currently displaying an error. I'm going to add the fix with the same PR where I'm excluding datepicker styles in composer.  
#### Staging Deployment

### DIFF
--- a/package.json
+++ b/package.json
@@ -112,7 +112,7 @@
   },
   "jest": {
     "transformIgnorePatterns": [
-      "/node_modules(?!/@bufferapp/performance-tracking)(?!/@bufferapp/notifications)(?!/@bufferapp/analyze-date-picker)(?!/@bufferapp/analyze-profile-selector)(?!/@bufferapp/analyze-shared-components)(?!/@bufferapp/report-list)(?!/@bufferapp/async-data-fetch)(?!/@bufferapp/components)(?!/@bufferapp/web-components)(?!/@bufferapp/composer)(?!\/@bufferapp\/unauthorized-redirect)/"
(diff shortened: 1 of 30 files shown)


package.json
```diff
@@ -112,7 +112,7 @@
   },
   "jest": {
     "transformIgnorePatterns": [
```
albennett: created a separate PR for adding this to its own file.

packages/profile-sidebar/middleware.js
```diff
@@ -61,12 +61,10 @@ export default ({ dispatch, getState }) => next => (action) => {
           profile,
         }));
         // Dispatch different select profile for components in analyze
-        if (params.tabId === 'analytics') {
-          dispatch({
-            type: 'PROFILE_SELECTOR__SELECT_PROFILE',
-            profile: formatAnalyticsProfileObj(profile),
-          });
-        }
+        dispatch({
```
albennett: Had to remove the tabId logic, because sometimes it would get stuck in "Loading" with data. I'm assuming we don't want this dispatch to get called every time though. Happy to hear your thoughts!

hamstu: Ah interesting, I'd mark this one as something to come back to. Since this will probably end up triggering a lot of unnecessary requests while browsing the rest of Publish.

package.json
hamstu: Awesome! ✨

PR #452, opened 2018-11-27
Fixing styles for broken calendar on edit

MayaUribe:
### Purpose
Adding a class to fix the calendar style bug when editing a post on queued tab.

### Notes
I realize this isn't the best approach, it's a temporary quick one, so when we find out how to exclude the styles that are actually affecting here we should remove this line

### DIFF
--- a/packages/server/index.html
+++ b/packages/server/index.html
@@ -80,6 +80,12 @@
           transform: translateY(0);
         }
       }
+      /*
+      @todo: remove this class once analyze-shared-components is fixed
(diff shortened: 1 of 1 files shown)


MayaUribe:
@hamstu so it seems like the styling from analyze-shared-components are affecting somehow in the calendar component, I realize this isn't the best approach, but it's a temporary quick one, so when we find out how to exclude the styles that are actually affecting here we should remove this line 😄

PR #453, opened 2018-11-27
Refactor jest config into its own file

albennett:
### Purpose
Refactor jest config into own file and use analyze config file.
### Notes
Decided to break this out into it's own PR. There is not a JIRA task for it.

#### Staging Deployment

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
   "scripts": {
     "bootstrap": "yarn",
     "clean": "find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +",
-    "test": "if [ -n \"$CI\" ]; then yarn && jest --config --runInBand; else jest; fi",
-    "test:debug": "node --inspect node_modules/.bin/jest --config --runInBand",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -6,8 +6,8 @@
   "scripts": {
     "bootstrap": "yarn",
     "clean": "find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +",
-    "test": "if [ -n \"$CI\" ]; then yarn && jest --runInBand; else jest; fi",
-    "test:debug": "node --inspect node_modules/.bin/jest --runInBand",
+    "test": "if [ -n \"$CI\" ]; then yarn && jest --config --runInBand; else jest; fi",
```
hamstu: Do we need this `--config` here? It doesn't appear to be taking any argument, and I thought Jest might pick up the `jest.config.js` automatically? Could be wrong.

albennett: I thought so too when reading the docs, but couldn't get it to work properly without the --config. I'll check it again to be sure.

albennett: You're right! Just removed them. It must have had errors for different reasons before. Thanks :)

PR #456, opened 2018-11-28
[WIP] Add compare-chart component to analytics

albennett:
### Purpose
As a B4B user, I am able to view and interact with Engagement and Audience (compare-chart component) analytics.

https://buffer.atlassian.net/browse/PUB-918
#### Staging Deployment

### DIFF
--- a/analyze.config.js
+++ b/analyze.config.js
@@ -12,6 +12,7 @@ const analyzePackages = [
   'analyze-profile-selector',
   'analyze-shared-components',
   'average-table',
+  'compare-chart',
   'micro-rpc',
   'report-list',
   'summary-table',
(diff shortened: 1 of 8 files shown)


packages/profile-sidebar/reducer.js
```diff
@@ -103,7 +103,7 @@ export default (state = initialState, action) => {
 export const actions = {
   selectProfile: ({ profile }) => ({
     type: actionTypes.SELECT_PROFILE,
-    profileId: profile.id,
+    profileId: profile ? profile.id : null,
```
albennett: I'm including a quick fix to handle the error I posted in slack about connecting to fb.

hamstu: Would love to know more about why this one is happening! 🤔

PR #458, opened 2018-11-28
Remove average-table and profile-selector from analytics

albennett:
### Purpose
Remove average-table and profile-selector from analytics because of analyze package conflicts with store
#### Staging Deployment


hamstu:
This looks good to merge to me! Checked for that one user and it's not crashing! ✅ SHIP IT! 😂

PR #462, opened 2018-11-30
Update and improve Bugsnag integration ✨

hamstu:
### Purpose

Update to the latest Bugsnag library (`@bugsnag/js`) as well as adding the React library in to catch errors there. 💯 

### Notes

* There's a bit of extra work to correctly pass the git hash into a file that can be read by our server. This could be simplified by passing this into the `pre-build.sh` in our deployment pipeline.

* I'm using an alpha release of the `buffer-static-upload` that always uploads the `*.map` files. ([See my PR here](https://github.com/bufferapp/buffer-static-upload/pull/11)) — the reason for this is because the bundled JS files reference their source maps without a hash (which is added by the uploader) and so the uploader kept skipping them as unchanged. Not sure if this is the most ideal option, we may want to further improve the uploader to handle source map files in a smarter fashion 😄 

#### Staging Deployment


scripts/bugsnag-release.js
```diff
@@ -0,0 +1,30 @@
...
+    sourceControl: {
+      provider: 'github',
+      repository: 'bufferapp/buffer-publish',
+      revision: releaseData.version,
+    },
+    builderName: releaseData.author,
+  }, { /* opts */ })
+    .then(() => console.log('[bugsnag:release] SUCCESS'))
+    .catch(err => console.log('[bugsnag:release] FAIL', err.messsage));
+}
```
djfarrelly: Should we log a warning to the console if there is no version data?

packages/server/lib/bugsnag.js
```diff
@@ -0,0 +1,83 @@
+const Bugsnag = require('@bugsnag/js');
+const fs = require('fs');
+const { join } = require('path');
+
+const BUGSNAG_KEY = process.env.BUGSNAG_KEY;
+const HOSTNAME = process.env.HOSTNAME;
+
+const isProduction = process.env.NODE_ENV === 'production';
```
djfarrelly: Maybe we should change `NODE_ENV` to `"staging"` in the future - would that be cleaner?

packages/server/middleware.js
```diff
@@ -10,6 +10,7 @@ module.exports.apiError = (err, req, res, next) => { // eslint-disable-line no-u
   } else if (req.app.get('isProduction')) {
     req.app.get('bugsnag').notify(err, {
       originalUrl: req.originalUrl,
+      user: { id: req.session.global.userId },
```
djfarrelly: NOTE - With the single sign on and everyone using a global session, this value may change to be the User's global "Buffer Account" instead of their Publish user id. Just something to keep in mind! 😉

packages/server/lib/bugsnag.js
hamstu: Yeah definitely! Would probably just have to update some logic in a few other places that depend on `NODE_ENV` being `'production'`, but it's not much work and it would def. simplify my release stage logic.

packages/server/middleware.js
hamstu: Very helpful point! 

💡 Tangent thought: We're probably going to want some kind of web tool for us/advocates at that point to quickly find them in the MongoDB based on their new 'global id' 🤔

scripts/bugsnag-release.js
hamstu: Good idea! I think I may even go as far as to exit non-zero to halt the build. Since if the version file is missing then something is probably quite broken.

packages/server/middleware.js
djfarrelly: Yeah, we need to ship the buffer-admin dashboard soon @hamstu! The Core team will work on getting out the MVP version in January 😄

PR #464, opened 2018-12-03
[PUB-919] Add hourly-chart component to analytics

albennett:
### Purpose
As a B4B user, I am able to view and interact with the Hourly Engagement analytics.

https://buffer.atlassian.net/browse/PUB-919
### Review
~~I realized that the compare-chart (Metrics Breakdown) doesn't have the correct data for Total Followers. Something must be off with the rpc call.~~ Confirmed with Federico, the total followers count starts on the day you connect with analytics, so it will go from 0 to total number on that day. This is expected :) 
#### Staging Deployment


packages/analytics/analytics.css
```diff
@@ -1,3 +1,11 @@
 header > span > button {
   display: none;
 }
+
+div > section span, button span, [data-highcharts-chart] div span  {
```
albennett: I added css to overwrite some of the global styles in publish that were changing the analyze styles. I know using `important!` isn't ideal. Happy to hear suggestions

hamstu: Curious what this is fixing? Might be good to leave a comment on that. 😄

albennett: Good point! I just added some comments 😄

PR #466, opened 2018-12-03
[PUB-905] Add export-picker component to analytics package

albennett:
### Purpose
As a B4B user, I can export analytics for twitter or facebook as a CSV or as images
https://buffer.atlassian.net/browse/PUB-905
### Review
The font-size for the export picker drop-down is incorrect. However, the hourly-chart PR has a fix for this

#### Staging Deployment


packages/store/index.js
```diff
@@ -90,12 +93,16 @@ const configureStore = (initialstate) => {
...
         bufferMetricsMiddleware,
         draftsMiddleware,
+        notificationsProviderMiddleware,
         averageMiddleware,
         compareChartMiddleware,
         datePickerMiddleware,
-        notificationsProviderMiddleware,
         profileSelectorMiddleware,
         summaryTableMiddleware,
+        // Analyze -- These need to be the last middlewares in the chain
```
albennett: Looking at the analyze code, there's a note that the exports need to be last. When I didn't have them last, the download as images would get stuck while exporting.

PR #469, opened 2018-12-04
Use version of composer with stateless App component

hamstu:
### Purpose

See https://github.com/bufferapp/buffer-composer/pull/118

#### Staging Deployment


hamstu:
Thanks @albennett!

PR #471, opened 2018-12-06
Fixing issues with yarn storybook

karelm:
### Purpose

Fixing issues with storybook not compiling correctly

### DIFF
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,11 +1,19 @@
-const webpack = require('../node_modules/@storybook/core/node_modules/webpack');
-
 module.exports = {
-  plugins: [new webpack.DefinePlugin({
-    __PACKAGES__: JSON.stringify(`../packages/${process.env.PACKAGE || ''}`)
-  })],
   module: {
     rules: [
+      {
+        test: /\.less$/,
+        use: [
+          'style-loader',
+          {
+            loader: 'css-loader',
+            options: {
+              modules: true,
+            },
+          },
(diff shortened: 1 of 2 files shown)


.storybook/webpack.config.js
```diff
@@ -15,7 +23,7 @@ module.exports = {
       },
       {
         test: /\.jsx?$/,
-        exclude: /node_modules(?!\/@bufferapp\/performance-tracking)(?!\/@bufferapp\/async-data-fetch)(?!\/@bufferapp\/components)(?!\/@bufferapp\/web-components)(?!\/@bufferapp\/composer)(?!\/@bufferapp\/unauthorized-redirect)/,
+        exclude: /node_modules(?!\/@bufferapp\/*)/,
```
anafilipadealmeida: Would you be able to explain me why we've made this change to the excluded node modules @karelm? 😄

karelm: @anafilipadealmeida, this is because if there are node modules, they should already be compiled, so we shouldn't have to compile them ourselves, we should be consuming the output of someones work in its final state

anafilipadealmeida: 👍 It makes sense! I was just wondering why we had such a specific rule for some of them before!

karelm: I imagine its because before, we probably went and excluded all the rules that were causing problems in the project, while I decided (dont know whether wrongly or rightly) to excludes all bufferapp modules in totallity... hopefully it wont be a mistake 😅

PR #474, opened 2018-12-10
[PUB-994] Add profile-header component to Analytics toolbar

albennett:
### Purpose
Add profile-header component to Analytics toolbar
https://buffer.atlassian.net/browse/PUB-994
#### Staging Deployment

### DIFF
--- a/packages/profile-page/components/ProfilePage/index.jsx
+++ b/packages/profile-page/components/ProfilePage/index.jsx
@@ -41,9 +41,8 @@ const loadingAnimationStyle = {
 };
 
 const tabContentStyle = {
-  maxWidth: '54rem',
+  maxWidth: '864px',
 };
-
 
 const TabContent = ({ tabId, profileId, childTabId }) => {
   switch (tabId) {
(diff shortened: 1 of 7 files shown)


packages/profile-page/components/ProfilePage/index.jsx
```diff
@@ -41,7 +41,7 @@ const loadingAnimationStyle = {
 };
 
 const tabContentStyle = {
-  maxWidth: '49rem',
+  maxWidth: '54rem',
```
albennett: Dave approved this change. The reason we need to add it is that the width of the container was affecting the posts-table component from displaying correctly.

hamstu: 👍 Awesome work getting some design insight here @albennett! Do you think we could switch this to use `px` though? I find that so much easier to reason about in my head instead of `rem`s. (For additional context, it looks like we'll be using `px` almost exclusively in the new design system, so this would be a good one for consistency.)

I could be off on the math – so please double-check haha – but I think this would be `(16 * 54)px` or `864px` - since `1rem = 16px`; though maybe that just applies to the font sizes... let me know!

albennett: Sure can, I'll change it to `px`. Thanks for the suggestion, @hamstu 👨‍🎤

PR #475, opened 2018-12-10
[PUB-993]Add notification to display when analytics are not supported

albennett:
### Purpose
Add notification to display when analytics are not supported
https://buffer.atlassian.net/browse/PUB-993
#### Staging Deployment

### DIFF
--- a/packages/analytics/components/Notification/story.jsx
+++ b/packages/analytics/components/Notification/story.jsx
@@ -1,0 +1,10 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { checkA11y } from 'storybook-addon-a11y';
+import Notification from './index';
+
+storiesOf('Analytics Notification', module)
+  .addDecorator(checkA11y)
+  .add('should display notification if analytics are not supported', () => (
+    <Notification />
+));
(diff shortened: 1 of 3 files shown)


packages/analytics/components/Notification/story.jsx
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { checkA11y } from 'storybook-addon-a11y';
+import Notification from './index';
+
+storiesOf('Analytics Notification', module)
+  .addDecorator(checkA11y)
+  .add('should display notification if analytics are not supported', () => (
```
hamstu: This is a super small one, but I think this story might actually be better named 'not supported notification', or something like that. 

The current wording would fit better as the name of a jest test (e.g., `it('should display...', () => { })` that tests the logic of actually displaying, whereas this is a very simple rendering of the component (and it always renders, so it's not really dependant on anything being supported as the naming suggests) — again really more of a nitpick here, but I hope that makes sense! 😅

albennett: Yes, makes sense! I'll change it, thanks @hamstu 😄

PR #478, opened 2018-12-13
Add welcome modal for existing free user rollout

hamstu:
### Purpose

image

### Notes

* Also made the code that checks if the user can see new publish more generic with the 'hasNewPublish' field (which comes from https://github.com/bufferapp/buffer-web/pull/15732/files#diff-cbc48474f17890e24eba1543d58c8078R5354)

#### Staging Deployment

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -2,7 +2,7 @@ import 'babel-polyfill';
 import { configure } from '@storybook/react';
 
 // automatically import all story.js files
-const req = require.context('../packages/', true, /components\/.*\/*story\.jsx$/);
+const req = require.context(__PACKAGES__, true, /components\/.*\/*story\.jsx$/)
 
 function loadStories() {
   req.keys().forEach(req);
--- a/.storybook/webpack.config.js
+++ b/.storybook/webpack.config.js
@@ -1,4 +1,9 @@
+const webpack = require('../node_modules/@storybook/core/node_modules/webpack');
+
 module.exports = {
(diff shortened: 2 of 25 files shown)


.storybook/config.js
```diff
@@ -2,7 +2,7 @@ import 'babel-polyfill';
 import { configure } from '@storybook/react';
 
 // automatically import all story.js files
-const req = require.context('../packages/', true, /components\/.*\/*story\.jsx$/);
+const req = require.context(__PACKAGES__, true, /components\/.*\/*story\.jsx$/)
```
hamstu: @karelm I'm bringing this back as it's needed to do per-package storybook runs. It doesn't crash for me locally, so I think the `*` fix you added is enough and we can bring this back. Let me know if this makes sense!

.storybook/webpack.config.js
```diff
@@ -1,4 +1,9 @@
+const webpack = require('../node_modules/@storybook/core/node_modules/webpack');
+
 module.exports = {
+  plugins: [new webpack.DefinePlugin({
+    __PACKAGES__: JSON.stringify(`../packages/${process.env.PACKAGE || ''}`)
+  })],
```
hamstu: Same comment as storybook config 😄

suprasanna:
Hey @hamstu - feel free to deploy whenever feels ready 😄. If we can make that copy change to the modal that we chatted about in the Paper before we do, even better. If you shoot me a quick note once live, I can update advocates!

PR #486, opened 2018-12-18
[PUB-1011] Task/disable actions queue contributor

anafilipadealmeida:
### Purpose
Contributor should not see actions in posts (delete, edit, and so on). Only managers can perform actions in the queue's posts

### Notes
Task: [PUB-1011](https://buffer.atlassian.net/browse/PUB-1011)

### DIFF
--- a/packages/queue/index.js
+++ b/packages/queue/index.js
@@ -6,15 +6,26 @@ import { actions } from './reducer';
 
 import QueuedPosts from './components/QueuedPosts';
 
-const formatPostLists = (posts) => {
+const formatPostLists = (profile, posts) => {
+  const isManager = profile.isManager;
   const orderedPosts = Object.values(posts).sort((a, b) => a.due_at - b.due_at);
   let lastHeader = null;
   return orderedPosts.reduce((acc, post, index) => {
     if (lastHeader !== post.day) {
       lastHeader = post.day;
-      acc.push({ queueItemType: 'header', text: post.day, id: `header-${index}` });
+      acc.push({
+        queueItemType: 'header',
+        text: post.day,
+        id: `header-${index}`,
+        isManager,
+      });
     }
-    acc.push({ queueItemType: 'post', index, ...post });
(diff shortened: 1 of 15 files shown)


packages/queue/index.js
```diff
@@ -26,11 +37,15 @@ export default connect(
     const profileData = state.profileSidebar.profiles.find(p => p.id === ownProps.profileId);
     if (profileQueuePosts && profileData) {
       return {
+        isManager: state.profileSidebar.selectedProfile.isManager,
```
hamstu: Is this one needed? I don't see it being used in the `QueuedPosts` component.

packages/queue/index.js
```diff
@@ -26,11 +37,15 @@ export default connect(
...
         loading: profileQueuePosts.loading,
         loadingMore: profileQueuePosts.loadingMore,
         moreToLoad: profileQueuePosts.moreToLoad,
         page: profileQueuePosts.page,
-        postLists: formatPostLists(profileQueuePosts.posts),
+        postLists: formatPostLists(
+          state.profileSidebar.selectedProfile,
```
hamstu: Perhaps since this method only needs to know `isManager` on the profile you can pass `isManager` directly? You could also access it in a slightly shorter way with `profileData.isManager` I believe.

anafilipadealmeida: True! Great suggestion 👍

anafilipadealmeida:
@hamstu Also added the `isManager` prop to determine when to display the share next button in the sent tab. Afterwards I went to the storybook to make the changes in the SentPosts story and it took me a bit to realize it's not broken, just displaying the loading mode on purpose. 🤦‍♀️ I believe in the future we could benefit more if we had a proper example of a sent page in stories, but for now I'm gonna leave it like that.

hamstu:
Good catch on the storybook there @anafilipadealmeida! Yeah, we should definitely make that more usable.

PR #488, opened 2018-12-19
[PUB-946] Redirect user to queue if they do not have access to routes

albennett:
### Purpose
Redirect user to queue if they do not have access to routes
https://buffer.atlassian.net/browse/PUB-946

Example: For a free user, they will not see the Drafts tab in the navigation. However, if they visit the drafts route, they will still have access to it. 
### Notes
I added the logic to the tabs package since we were already checking permissions there. It could also make sense to add the logic to the profile-page package.
#### Staging Deployment

### DIFF
--- a/packages/tabs/components/TabNavigation/index.jsx
+++ b/packages/tabs/components/TabNavigation/index.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Redirect } from 'react-router';
 import PropTypes from 'prop-types';
 import {
   Tabs,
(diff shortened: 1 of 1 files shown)


anafilipadealmeida:
navigation
Hey @albennett, I'll take a deeper look at it in a couple of hours, but noticed that when trying to navigate to a different account that doesn't have the tab we are in, it selects the first social account we have by default.

albennett:
Great point, @anafilipadealmeida! I'll add to the publish-formatters package later today to allow passing the profile id when redirecting to queue 😄

PR #492, opened 2019-01-02
Handle forEach error if orderedPosts is not an array

albennett:
### Purpose
Handle error for forEach if orderedPosts is not an array
https://app.bugsnag.com/buffer/buffer-publish/errors/5c2cccf7851927001c792b0d?filters[event.since][0]=30d&filters[error.status][0]=open
### Notes
Noticed this error in bugsnag after merging the imageClick PR for sent posts this morning. `TypeError · e.forEach is not a function` I wasn't able to replicate the error while impersonating the users, but I'm assuming orderedPosts must be undefined or not an array.  
#### Staging Deployment

### DIFF
--- a/packages/sent/index.js
+++ b/packages/sent/index.js
@@ -8,7 +8,9 @@ const formatPostLists = (posts) => {
   const postLists = [];
   let day;
   let newList;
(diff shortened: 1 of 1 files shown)


hamstu:
Thanks for the quick bugfix on this one @albennett!

PR #497, opened 2019-01-07
Normalize general settings styles

anafilipadealmeida:
### Purpose
Design tweaks to general settings tab

### DIFF
--- a/packages/general-settings/components/GoogleAnalytics/index.jsx
+++ b/packages/general-settings/components/GoogleAnalytics/index.jsx
@@ -1,20 +1,23 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import {
-  Text, Toggle, Divider,
-  Button, Input, Card,
+  Text, Toggle, Button, Input, Card,
 } from '@bufferapp/components';
 
 const enableGoogleAnalyticsStyle = {
-  display: 'flex',
-  flex: 1,
-  flexDirection: 'column',
   marginBottom: '0.5rem',
 };
 
 const generalWrapperStyle = {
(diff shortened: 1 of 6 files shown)


packages/general-settings/components/GoogleAnalytics/index.jsx
```diff
@@ -44,116 +46,126 @@ const GoogleAnalytics = ({
...
-          </div>
-          <div style={inputStyle}>
             <Text
-              weight={'medium'}
-              color={'black'}
+              size={'small'}
+              weight={'bold'}
             >
-              Campaign Medium
+              &nbsp;(This will override any existing UTM parameters)
```
karelm: Hey @anafilipadealmeida, just wondering why theres a non-breaking space at the start of this sentence? Mainly out of curiosity really 😄

anafilipadealmeida: Great question @karelm ! I found this to be the simplest way of adding the space between the components. I thought about a padding or adding another property to the parent, but didn't think it was worth it 😅Were you thinking about something else?

karelm: Nope, just saw that there was a space there, and there wasn't one before, so thought I'd ask incase it was an accident / just curious really 👍 straight forward way to add an extra space when needed, looks perfect to me 💯

anafilipadealmeida: Yes, I changed the two Text components to be displayed in a row instead of a column, that's why the space was needed now 😄

PR #500, opened 2019-01-09
Upgrade Composer to v2.6.2

hamstu:


albennett:
Closing as the composer version in production is higher than in PR

PR #502, opened 2019-01-09
Upgrade Composer to v2.6.5

hamstu:
See: https://github.com/bufferapp/buffer-composer/pull/132


hamstu:
Closing in lieu of a newer version to come with a needed fix.

PR #507, opened 2019-01-16
Add Instagram Direct Posting Modal

anafilipadealmeida:
### Purpose
Add IG Direct Posting Modal for new IG Authentication flow

### Notes
Works under the feature flip `new_ig_authentication`


anafilipadealmeida:
captura de ecra 2019-01-16 as 14 10 33
Hey @hamstu , if I try to import the Button from @bufferapp/ui I get this error. Am I doing something wrong?

hamstu:
Haven't seen that before @anafilipadealmeida! But then I think only Reply has really started using the components so far. What happens if you add `yarn add immutability-helper` to publish?

hamstu:
Hey @anafilipadealmeida! I've added a commit here to use a newer/fixed version of the UI components. It should compile an work now.

⚠️ **Important note:** The new `<Button>` component is not a drop in replacement for our current Button. Be sure to refer to the docs to see the new usage: https://bufferapp.github.io/ui/#/ui/ui/button – one important difference is that you pass the text as a `label` prop, instead of as one of the children of the component.

PR #509, opened 2019-01-17
[PUB-1039] Changing calendar URL to use new URL

MayaUribe:
### Purpose
Changing calendar URL to use new generic URL (content_only)


karelm:
thought this needed to be added here too :shipit:

PR #510, opened 2019-01-22
[PUB-1046] Adding Billing in the Settings Menu in New Publish.

MayaUribe:
### Purpose
Added Billing into Settings menu.

### Notes
I added the Billing as a sub menu within Settings, this can be changed if necessary.


karelm:
Great stuff here @MayaUribe 👍

PR #513, opened 2019-01-23
[PUB-1027] Clearing default scheduled time when sharing a post again (re-buffer).

MayaUribe:
### Purpose
Clearing default scheduled date and time when sharing a post again (re-buffer), because currently it marks as is set in a past date.

### DIFF
--- a/packages/sent/reducer.js
+++ b/packages/sent/reducer.js
@@ -199,13 +199,17 @@ export default (state = initialState, action) => {
 };
 
 export const actions = {
-  handleShareAgainClick: ({ post, profileId }) => ({
-    type: actionTypes.OPEN_COMPOSER,
-    updateId: post.id,
-    editMode: false,
-    post,
-    profileId,
-  }),
+  handleShareAgainClick: ({ post, profileId }) => {
+    post.scheduled_at = null;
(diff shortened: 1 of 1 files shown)


packages/sent/reducer.js
```diff
@@ -199,13 +199,17 @@ export default (state = initialState, action) => {
...
-  handleShareAgainClick: ({ post, profileId }) => ({
-    type: actionTypes.OPEN_COMPOSER,
-    updateId: post.id,
-    editMode: false,
-    post,
-    profileId,
-  }),
+  handleShareAgainClick: ({ post, profileId }) => {
+    post.scheduled_at = null;
+    post.due_at = null;
```
karelm: can i suggest that instead of changing the value of scheduled_at and due_at, that instead perhaps the spread operator be used here.

so 
```
    return {
    post,	      type: actionTypes.OPEN_COMPOSER,
    profileId,	      updateId: post.id,
  }),	      editMode: false,
      post: {
    ...post,
    scheduled_at: null,
    due_at: null,
    },
      profileId,
    };    
```
that way we arent modifying the original object, so no side-effects can be caused here? what do you think to that?

PR #522, opened 2019-02-01
[PUB-800] Adding rpc endpoint and wiring it up to Google Analytics form.

MayaUribe:
### Purpose
Adding rpc endpoint and wiring it up to Google Analytics form.

### DIFF
--- a/packages/general-settings/middleware.js
+++ b/packages/general-settings/middleware.js
@@ -1,8 +1,9 @@
 import { getURL } from '@bufferapp/publish-formatters';
 import { actionTypes as profileActionTypes } from '@bufferapp/publish-profile-sidebar';
+import { actions as notificationActions } from '@bufferapp/notifications';
 import {
   actions as asyncDataFetch,
-  actions as dataFetchActions,
+  actions as dataFetchActions, actionTypes as dataFetchActionTypes,
 } from '@bufferapp/async-data-fetch';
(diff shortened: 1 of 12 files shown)


packages/general-settings/middleware.js
```diff
@@ -1,8 +1,9 @@
 import { getURL } from '@bufferapp/publish-formatters';
 import { actionTypes as profileActionTypes } from '@bufferapp/publish-profile-sidebar';
+import { actions as notificationActions } from '@bufferapp/notifications';
 import {
   actions as asyncDataFetch,
-  actions as dataFetchActions,
+  actions as dataFetchActions, actionTypes as dataFetchActionTypes,
```
karelm: @MayaUribe, would you be okay with putting this on the next line at all?

packages/general-settings/reducer.js
```diff
@@ -71,6 +89,31 @@ export default (state = initialState, action) => {
         ...state,
         googleAnalyticsEnabled: action.result.isEnabled,
       };
+    case actionTypes.SAVE_GA_CUSTOM_FORM:
+      return {
+        ...state,
```
karelm: @MayaUribe, I'm just checking here, but do you want to set the UTM settings after the save action happens, or is it enough that they have already been set in the previous set_utm_<thing> calls?

PR #524, opened 2019-02-01
[PUB-1104] Add IG data tracking

anafilipadealmeida:
### Purpose
Add Instagram action tracking

### Notes
[PUB-1104](https://buffer.atlassian.net/browse/PUB-1104)

### Review
Hey @hamstu, I realized instead of tracking an action twice when we want custom metadata (like with the Upgrade Modal), we could just create an exception inside the function `createBufferMetricsMiddleware`.  How do you feel about it? We could even list them all in another file if in the future it gets too long 😄

### DIFF
--- a/packages/store/index.js
+++ b/packages/store/index.js
@@ -48,10 +48,19 @@ export const history = createHistory();
 
 const bufferMetricsMiddleware = createBufferMetricsMiddleware({
   application: 'PUBLISH',
-  metadata: state => ({
-    userId: state.appSidebar.user.id,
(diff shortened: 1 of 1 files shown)


hamstu:
This feels like a great way to do it @anafilipadealmeida! In the cases where there's already an action being dispatched then this is a nice way to add extra data. 

I know one or two of my upgrade modal tracking pieces could use this, but I think one of them is tracking something that isn't as clearly named in its action, so it helps to manually track there – but I think this is a great pattern going forward to collect more metadata. Thanks!

PR #527, opened 2019-02-05
Move /health-check before the validateSessionMiddleware

esclapes:
### Purpose

There is a fair amount of unnecessary requests to the `buffer-login` services generated from the publish `health-check` route.

This route was registered after the `validateSessionMiddleware` which redirects to `login.buffer.com` if the session is not valid, which is always the case for the healthcheck.


hamstu:
Thanks for dropping in to fix this one @esclapes! 🙏

PR #528, opened 2019-02-05
[PUB-1058] Format post error messages

albennett:
### Purpose
Format post error messages by stacking the error into its own bar and adding the longer messages with link steps. 
https://buffer.atlassian.net/browse/PUB-1058
### Notes
PR for backend: https://github.com/bufferapp/buffer-web/pull/15848
### Review
Wasn't quite sure the best way to parse html string with react. https://buffer.slack.com/archives/C151RRDL1/p1549470020433500
screen shot 2019-02-06 at 6 57 57 pm

#### Staging Deployment


packages/shared-components/PostErrorBanner/index.jsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { Text } from '@bufferapp/components';
+import { white } from '@bufferapp/components/style/color';
+
+const errorBannerWrapper = dragging => ({
+  backgroundColor: '#E0364F', // TODO: add to buffer-components
```
albennett: Is this a color with the new design system? Should I not add it to buffer-components since we're switching to the new design system soon?

packages/shared-components/PostErrorBanner/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+
+const createMarkup = error => (
+  { __html: error }
+);
+
+// Need to add <a> style because can't directly get to <a> in error html string
+const PostErrorBanner = ({ dragging, error }) => (
+  <div id="errorBanner" style={errorBannerWrapper(dragging)}>
+    <style>{`
+      #errorBanner a { color: ${white}; }
```
hamstu: Ah yeah, I see how this one is tricky. Feels great to put in some direct CSS to help with this for now! 👍

packages/shared-components/PostErrorBanner/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+    `}
+    </style>
+    <Text
+      size={'mini'}
+      weight={'medium'}
+      color={'white'}
+    >
+      <div dangerouslySetInnerHTML={createMarkup(error)} />
```
hamstu: Living on the edge! 💯 

(In all seriousness though, this is the best way to insert HTML and since it comes from our backend we can trust it, so no risks here. It's only `dangerous` if it's rendering something from the user.)

packages/shared-components/PostErrorBanner/index.jsx
```diff
@@ -0,0 +1,45 @@
...
+  backgroundColor: '#E0364F', // TODO: add to buffer-components
+  padding: '8px 16px',
+  color: white,
+  display: 'flex',
+  alignItems: 'center',
+  marginLeft: '-1px',
+  marginRight: '-1px',
+  lineHeight: '20px',
+  borderRadius: '4px 4px 0 0',
+  opacity: dragging ? 0 : 1,
```
hamstu: Good catch in making sure this doesn't show when the post turns into 'placeholder' mode!

packages/shared-components/PostErrorBanner/index.jsx
hamstu: If you got it from Tyler then it's definitely in BDS. But don't worry about trying to pull or put it anywhere central for now; not really worth the effort until we better scope out our BDS migration plan.

PR #539, opened 2019-02-08
PUB-1111 | Updating sidebar to support back to classic buffer menu option

karelm:
### Purpose

Created back to buffer option in side menu

image

this launches the modal:

image


packages/sidebar/README.md
```diff
@@ -0,0 +1,116 @@
+# @bufferapp/publish-example
```
josem: Should we update/remove this readme?

packages/sidebar/components/Sidebar/story.jsx
```diff
@@ -0,0 +1,98 @@
...
+  subscribe: () => {},
+  dispatch: () => {},
+  getState: () => ({...state}),
+});
+
+const fakeUser = {
+  loading: false,
+  id: '1234',
+  name: 'Hamish Macpherson',
+  email: 'hamstu@gmail.com',
```
josem: This is already in different places in the Git history so not too important at this point but since this is a public repo, we might want to use a fake email and name like test@example.com before merging it to master.

karelm:
@jose can I get a second review over this when you have a moment please?

josem:
Looks good to me @karelm! 💯

packages/app-switcher/middleware.js
```diff
@@ -12,9 +12,7 @@ export default ({ dispatch }) => next => (action) => {
       }));
       break;
     case `sendFeedback_${actionTypes.FETCH_SUCCESS}`:
-      setTimeout(() => {
-        window.location = getURL.getClassicBufferURL();
-      }, 1000);
+      window.location = getURL.getClassicBufferURL();
```
josem: In this case @karelm, it'd be interesting to add the /back_publish route we've prepared here: https://github.com/bufferapp/buffer-web/blob/1a0d1f3eec01d5e369d47ed924535a91db437799/app/config/routes.php#L300. So we can correctly set the features and everything for the user.

karelm: That's a great catch, thanks Jose I'll get that sorted 🙂

karelm:
@twanlass are you okay giving the design elements here a once over for me? you will need the 'Paid Users in New Publish' flag enabled to see it in the sidebar menu, but hopefully you should be able to see it. Let me know if not 😄

twanlass:
Heya @karelm 👋 

One small design note here (so sorry for not being more clear 🙈): the menu item /  `<li>` shouldn't have a dark background by default, all menu items have hover state. [[GIF](http://hi.buffer.com/da427c2abd5d)]

If adding a hover state to the menu items is a challenge at the moment feel free to omit the background color change altogether. [[REF](http://hi.buffer.com/5deab25e0100)]

karelm:
Thanks @twanlass I'll look at adjusting that, sorry I thought you meant that you wanted it that colour at all times, I'll see what I can do. I did think it looked strange, that's why I said about the space between it and the menu item below. Sorry, I misunderstood that and should have asked for further clarification 🙂 thanks for reviewing this for me 🤗

twanlass:
No worries at all @karelm – totally on me 🙌

PR #541, opened 2019-02-08
[PUB-1052] Adding a modal for steal profiles warning

MayaUribe:
### Purpose
Adding a modal warning users they are attempting to connect a profile that has already being connected to a different Buffer Org.

### DIFF
--- a/packages/modals/util/showModal.js
+++ b/packages/modals/util/showModal.js
@@ -10,6 +10,9 @@ const getShowModalKey = () => {
   return '';
 };
 
+export const shouldShowStealProfileModal = () =>
+  window.location.hash.match(/#show-cant-transfer-profile\/?$/i);
+
 export const shouldShowWelcomeModal = () =>
   getShowModalKey() === 'welcome-modal-1';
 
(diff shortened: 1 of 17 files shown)


packages/modals/util/showModal.js
```diff
@@ -10,6 +10,9 @@ const getShowModalKey = () => {
   return '';
 };
 
+export const shouldShowStealProfileModal = () =>
+  window.location.hash.match(/#show-cant-transfer-profile\/?$/i);
```
hamstu: Hey @MayaUribe! Just :ceilingcatting: on this one, but I wonder if we could use the same functionality as the other methods here instead of checking the hash (`#`) value, I say this since I used this method originally with `AppModals`, but found it was not reliable (sometimes it just wouldn't work, I think React/react-router was mixing it up. 🤷‍♂️)

The way to do it like the others would be to snure we send someone to NP with the the right query params; e.g., `https://publish.buffer.com?mk=cant-transfer-profile`. The only gotcha is that this doesn't yet support passing along a variable. You can see in the code I split at the `--` character for the upgrade modal to pull out the source, so you could do something like that, or maybe we can add that functionality to the Publish node server (packages/server/index.js). I think it would be pretty quick to add support for an extra &mv=<blabla> param that we could use here.

Let me know if this makes sense! I'm happy to hop on a quick pair call if it helps here. 😄

packages/modals/util/showModal.js
```diff
@@ -24,3 +27,10 @@ export const getSourceFromKey = () => {
   const key = getShowModalKey();
   return key.split('--')[1] || 'unknown';
 };
+
+export const getURLParameter = (name) => {
```
hamstu: ☝️ Same as my note above, I think we could avoid this tricky RegEx'in by using the `?mk=modal-name` param and perhaps adding support for a variable in the server. 😄

packages/modals/util/showModal.js
MayaUribe: Great suggestion @hamstu! I was actually thinking of a better way to do that part, and this comment was in the right moment! thank you a lot, I will add that support to send a value 🥇

PR #548, opened 2019-02-11
Add `dd-agent` for better monitoring

hamstu:
### Purpose

Following some unexpected downtime we've decided to add the DD APM for better monitoring in New Publish.

### Notes

@erickhun Let me know if this looks good to you!
@kiriappeee / @erickhun I'm also wondering if either of you know a way we can know whether the `env` is staging vs production? I'm not sure if we have an env variable exposed to the publish server for this. (See below image)

image

#### Staging Deployment

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -33,6 +33,10 @@ extraSpecs:
     type: RollingUpdate
   minReadySeconds: 10
 env:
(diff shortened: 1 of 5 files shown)


hamstu:
Looking good on the DD side – going to merge this one in!

PR #549, opened 2019-02-11
[PUB-1115] Allow a user to drag and reorder their sidebar profiles

albennett:
### Purpose
Give the option for users to drag and reorder their sidebar profiles
https://buffer.atlassian.net/browse/PUB-1115
### Notes
Refactored locked and unlocked profiles into one list so that drag and drop would work for all. 
### Review
@hamstu Currently getting a storybook error when running tests for `Could not find the drag and drop manager in the context of DragSource(ProfileDragWrapper)`. Will look more into it. Curious if you ran into this error when you were working on queue drag and drop
#### Staging Deployment

### DIFF
--- a/packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
+++ b/packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
@@ -1,0 +1,110 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+
+import flow from 'lodash.flow';
+import { DragSource, DropTarget, DropTargetConnector } from 'react-dnd';
+import ProfileListItem from '../ProfileListItem';
+
+const profileSource = {
+  beginDrag(props, monitor, component) {
+    return {
+      id: props.id,
+      index: props.index,
+      width: component.containerNode.offsetWidth,
+    };
+  },
+};
+
+const profileTarget = {
+  drop(props, monitor, component) {
+    component.decoratedComponentInstance.containerNode.blur();
+    props.onDropProfile({ commit: true });
+  },
+  hover(props, monitor, component) {
+    const { index: dragIndex } = monitor.getItem();
+    const { index: hoverIndex, onDropProfile } = props;
+
+    // Don't replace profile with itself...
+    if (dragIndex === hoverIndex) {
+      return;
+    }
+
+    // Determine rectangle on screen
+    const node = component.decoratedComponentInstance.containerNode;
+    const hoverBoundingRect = node.getBoundingClientRect();
+
+    // Get vertical middle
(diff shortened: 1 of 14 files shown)


packages/profile-sidebar/components/ProfileSidebar/index.jsx
```diff
@@ -37,39 +35,12 @@ const profileListStyle = {
...
-        <div style={{ padding: '.5rem .25rem' }}>
-          {translations.lockedListTooltip1 +
-            profileLimit +
-            translations.lockedListTooltip2}
-        </div>
-      </IconArrowPopover>
-    </div>
-  </div>
-);
+const DefaultFallbackType = <Text size={'large'}>Free</Text>;
```
albennett: I had to resolve some conflicts with master. This is a line already in master.

packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
```diff
@@ -0,0 +1,107 @@
...
+    if (dragIndex === hoverIndex) {
+      return;
+    }
+
+    // Determine rectangle on screen
+    const node = component.decoratedComponentInstance.containerNode;
+    const hoverBoundingRect = node.getBoundingClientRect();
+
+    // Get vertical middle
+    const hoverThird = (hoverBoundingRect.bottom - hoverBoundingRect.top) / 3;
```
karelm: get vertical middle... but divide by 3, just checking thats the expected approach there

karelm:
Interestingly, If i have a locked social account I can drag it out and move it round with my other accounts... might be worth considering: https://www.useloom.com/share/319d863a9c354434af6fdf618d80913e

packages/profile-sidebar/components/ProfileSidebar/index.jsx
karelm: I checked in the file, and I cant find any usages of this variable anywhere... so you may be able to remove it/may need to check if another piece of logic got removed aswell!? (perhaps)

albennett:
Hey Karël! Thanks so much for reviewing this. Regarding dragging a locked account, this is correct in behavior. When a paid user downgrades to free, they often realize that some of their accounts are locked and want to change which ones they have access to. Currently, advocates are suggesting they reorder the profiles to change which ones they get access to. We probably want to revisit this solution in the future though :)

packages/profile-sidebar/components/ProfileSidebar/index.jsx
albennett: Ah, you're right! It's not in master, so it must have been removed recently. It seems I accidentally kept it around in the merge. 🙈 I'll remove it.

PR #562, opened 2019-02-15
[PUB-1123] Add "Learn about Buffer for Business" CTA in header for Pro

hamstu:
### Purpose

[PUB-1123](https://buffer.atlassian.net/browse/PUB-1123)

### Notes

@karelm I've made the logic for tracking actions manually in publish a little more simple with a new `@bufferapp/publish-data-tracking` package. There's a `trackAction` method that will automatically set the application to 'PUBLISH' and the user ID of the current user. I also added support for callbacks, as I noticed you added that to the buffermetrics library recently!

I know you're working on this: https://github.com/bufferapp/buffer-publish/pull/539 – so maybe once this is merged in you could refactor the tracking call you have. 😄 

#### Staging Deployment


anafilipadealmeida:
Ceiling cat on this one @karelm and @hamstu : I feel a bit uncomfortable about us having duplicated tracking. I believe maybe having actionTypes with more descriptive names, like `actionTypes.PUBLISH_UPGRADE_something` and make a better use of custom metadata fields in the `getMetricsMetadata` function would be enough for us to avoid tracking things twice (once manually and another automatically). But I would love to hear your opinion on that 😄

hamstu:
Thank you both for the review and comments, I really appreciate it! (Ceiling cats are always welcome on my PRs Ana!) 🙏 

@karelm:
> I think the only thing that concerns me is that we are duplicating some of the buffermetrics logic

 Ah, do you mean the wrapping of the call to `buffermetrics.trackAction`? If so, I can see what you mean, but I think it does work well since a) it removes the need to pass the `application` and the `{ userId }` metadata every time and b) it allows us to spy on the tracked events so we can surface what's being tracked in the console. Let me know if I'm off in my understanding of what you're saying though!

One thought I had would be that we could maybe update the `buffermetrics/client`  directly to support something like this:

```js
import { createTrackAction } from '@bufferapp/buffermetrics/client';

const trackAction = createTrackAction({
	application: 'PUBLISH',
	getUserId: () => window._user && window._user.id,
	onBeforeSend: (event) => logTrackingToConsole(event),
});

// Use it elsewhere
trackAction({ { location: 'sidebar', action: 'clicked_cta', metadata: { foo: 'bar' } }, callbacks });
```

Probably not something for now, but maybe a future refactor? 🤔 

---

@anafilipadealmeida 
> I feel a bit uncomfortable about us having duplicated tracking.

Thanks for sharing that @anafilipadealmeida! Yep, we do – as you mentioned - have some duped (or well, very similar) from tracking from the upgrade modal. If you run `showTracking()` in the console and then open it the upgrade modal you will see something like this:

```
[buffermetrics] publish modals show_upgrade_to_pro {userId: "52dffdf286b3e9b10700000f", source: "app_sidebar"}
[buffermetrics] publish modals show_upgrade_modal {userId: "52dffdf286b3e9b10700000f", profileId: "5c61c38eea5527524b3301e3"}
```

That first one is from my manual tracking, and the second is automatic. **I completely agree we should just have track this as a single event.** Right now I know that the data team is only looking at the first one `show_upgrade_to_pro`, and so while I didn't get to it in this yet, I plan on cleaning it up in the future! Thanks for bringing it up 😄 

There is one event here that _does_ need to be tracked manually

anafilipadealmeida:
@hamstu Sounds great! How do you feel about writing this info in a README file or leave a note somewhere, so everybody on the team gets the full context? 😄 Someone else might unintentionally track things twice without realizing why!

hamstu:
Yep! Planning to share this new tracking methodology along with notes like that once this is merged. 🙏

PR #564, opened 2019-02-15
Upgrading library to fix undefined variable issue

karelm:
### Purpose
Upgrading library to fix undefined variable issue


hamstu:
Thanks for the speedy fix on this one! 🔥

PR #568, opened 2019-02-19
[PUB-1168] Stylings for past reminders.

MayaUribe:
### Purpose
Update story for past reminders and added styles for the header and subtitle.


MayaUribe:
Hey there @anafilipadealmeida! would love for you to take a look at this one, I updated the story, but I know it will need an adjustment later on, when you deploy the Send to Mobile part :)

packages/past-reminders/components/PastRemindersPosts/index.jsx
```diff
@@ -17,6 +17,16 @@ const headerStyle = {
   width: '100%',
 };
 
+const titleStyle = {
+  color: '#000',
```
anafilipadealmeida: Hey @MayaUribe, we don't need the color here, since we already have <Text color={'black'}>. The display:block is also not needed, as this is a div.

packages/past-reminders/components/PastRemindersPosts/index.jsx
```diff
@@ -17,6 +17,16 @@ const headerStyle = {
...
+  marginBottom: '8px',
+  display: 'block',
+};
+
+const subTitleStyle = {
```
anafilipadealmeida: I would personally remove this fontSize. The version 3.1.0 of buffer-components is making the size={'mini'} equivalent to 14px, so once the version is updated in publish this Text component would already turn into 14px.

packages/past-reminders/components/PastRemindersPosts/index.jsx
MayaUribe: Woops, I forgot to remove that line!

packages/past-reminders/components/PastRemindersPosts/index.jsx
MayaUribe: Again, I was playing with Tyler styles, but I changed those I haven't commit those changes, will in a few minutes.

packages/past-reminders/components/PastRemindersPosts/index.jsx
```diff
@@ -12,11 +12,20 @@ import {
 } from '@bufferapp/components';
 import ComposerPopover from '@bufferapp/publish-composer-popover';
 
+import {
+  fontFamily,
```
anafilipadealmeida: @MayaUribe Do we need this?

MayaUribe: No we don't! removing it! :)

PR #569, opened 2019-02-19
Adjusting Send to mobile text.

MayaUribe:
### Purpose
Adjusting Send to mobile text.


anafilipadealmeida:
Hey @MayaUribe, how does Tyler feel about the buttons not having the same width? Is that a problem?

MayaUribe:
> Hey @MayaUribe, how does Tyler feel about the buttons not having the same width? Is that a problem?

I just saw your comment here @anafilipadealmeida, no actually I just committed a change that makes those buttons have the same width.

PR #573, opened 2019-02-20
JIRA:PUB-1172 | Informing users that they are actively trialling a plan

karelm:
### Purpose

Show users in new publish a banner Informing users that they are actively trialling a plan

http://hi.buffer.com/a4b8ed82bced

http://hi.buffer.com/204b016db13e

vs what used to be there:

http://hi.buffer.com/204b016db13e

#### Staging Deployment

### DIFF
--- a/packages/cta-banner/components/BillingUpgradeCTABanner/index.jsx
+++ b/packages/cta-banner/components/BillingUpgradeCTABanner/index.jsx
@@ -1,0 +1,159 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+import { Text } from '@bufferapp/components';
+import FeatureLoader from '@bufferapp/product-features';
+import { insertIntoTranslation } from '@bufferapp/publish-i18n';
+
+const textColor = 'white';
+
+const styling = {
+  backgroundColor: '#1F35B3',
+  color: textColor,
+  padding: '5px',
+  textAlign: 'center',
+};
+
+const buttonStyle = {
+  color: textColor,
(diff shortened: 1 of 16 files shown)


karelm:
Hey @hamstu, 

great questions there, and I'm writing up tests around more of the logic at present, as I had the same concerns, just wanted to get something out asap, and then polish up the testing logic around this work... also, great shout on the name of the function, it definitely started as a straight forward text logic... but I think you might be right there actually, it would be better to abstract the `* typeof s === 'string' ? s : s.map((e, i) => <Fragment key={String(i)}>{e}</Fragment>` logic into something more re-usable and maybe apply the outcome of that to the resultant function (not sure that makes sense as a sentence, but essentially I think what you said was right)

Thanks for highlighting these things, please watch this space as I'm working on it at present, and hopefully will have something wrapped up for you in this area soon.

packages/cta-banner/components/BillingUpgradeCTABanner/index.jsx
```diff
@@ -0,0 +1,156 @@
...
+  );
+
+  if (!trial.hasCardDetails) {
+    const planTrial = [
+      { replaceString: '{plan}', replaceWith: currentPlan({ color: textColor, weight: 'bold', size: 'mini' }) },
+    ];
+    return (
+      <div style={styling}>
+        <Text weight="bold" color={textColor} size="mini">
+          <TranslationReplacer
```
hamstu: 😍 Love it!

PR #580, opened 2019-02-21
[PUB-1195] Preventing Publish from crashing on failing updates

MayaUribe:
### Purpose
Adding a Boundary error message, to prevent publish from crashing on updates with errors.
Related to https://buffer.atlassian.net/browse/PUB-1159


packages/error-boundary/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@bufferapp/publish-error-boundary",
+  "version": "2.0.0",
+  "description": "Allows to display error within Publish",
+  "main": "index.js",
+  "scripts": {
+    "lint": "eslint . --ignore-pattern coverage node_modules"
+  },
+  "author": "hharnisc@gmail.com",
```
karelm: @MayaUribe would you be able to fix up this package.json please?

packages/error-boundary/components/ErrorBoundary/index.jsx
```diff
@@ -0,0 +1,104 @@
...
+  <Button
+    secondary
+    onClick={() => { this.onRefresh(); }}
+  >
+    <div style={buttonWrapper}>
+      <RefreshIcon color={'curiousBlue'} size={{ width: '14px' }} />
+      <span style={buttonText}>Refresh</span>
+    </div>
+  </Button>
+  */
```
karelm: would you be able to remove this commented out section please?

packages/error-boundary/package.json
MayaUribe: Hey there @karelm I actually ended up deleting this package and creating a fallback component where the ErrorBoundary already takes place.

packages/error-boundary/components/ErrorBoundary/index.jsx
MayaUribe: Hey there @karelm sorry about that, I was doing some tests there, but I already cleaned up the code.

MayaUribe:
> This looks great @MayaUribe! LGTM! 👌
> 
> The only tweak I might suggest would be to import the CSS in the JS like
> 
> ```js-jsx
> import style from './errorBoundary.css`;` 
> ```
> 
> Then you can apply it to the divs/elements like:
> 
> ```js-jsx
> <div style={style.nameOfClass}`> ... </div>
> ```
> 
> Then we could remove the `global()`s in the CSS. The class names would also have to be converted to `camelCase` or you could access them with `style['name-of-class']`. Does that make sense? I say this only since globals can be tricky and cause conflicts sometimes :)

Awesome!! great idea, will update that for sure, makes perfect sense

hamstu:
Thank you @MayaUribe! 🙏

PR #582, opened 2019-02-22
PUB-1172 | Adjusting flexbox to ensure that the whitespace gaps are correct

karelm:
### Purpose

Making adjustments to components to ensure that when in a trial, the whitespace being shown at the top and bottom are correctly spaced

#### Staging Deployment


karelm:
Yeah, I kind of needed to, because the flexbox enforced the entire height available, but sometimes that meant it needed to be shared with another element, so 100vh was making it 100vh+other div and causing an overflow in whitespace 👍

PR #590, opened 2019-02-25
Change Babel config to make bundles compatible with IE11

josem:
### Notes
Added plugins to make bundled code compatible with Internet Explorer 11.

#### Staging Deployment
This repo is CI/CD enabled and staging deployment should be available at:
https://arrow-functions-ie11-publish.dev.buffer.com/


djfarrelly:
I'm curious @josem, does this target the dependencies too like `micro-rpc-client`. That has arrow functions as well and could be causing other issues if it's not transpiled: https://app.bugsnag.com/buffer/buffer-publish/errors/5c70617455368f001b1ed996?filters%5Bevent.since%5D%5B%5D=1d&filters%5Berror.status%5D%5B%5D=open&event_id=5c7440e5003594212fc90000

karelm:
Hey @djfarrelly it looks like you are correct that the micro-rpc-client and others weren't being transpiled correctly, so I've split them out and also some of our modules were also not being tranpiled correctly. So I've modified it, but I'm more concerned for the styling issues that have resulted, and the size of the bundle above just the "does it run" problem

this is the experience on ie11 after making it run
http://hi.buffer.com/798071551eee

josem:
Thanks so much @karelm! 🙌 

Yeah Dan, I've left an update on JIRA https://buffer.atlassian.net/browse/PUB-1119 and we won't probably support IE 11 for the time being since it might involve too much work at the moment for the small percentage of users that we have. We'll show a message instead (Super also suggested this).

djfarrelly:
Thanks for the updates @karelm and @josem - I added a quick note in Jira instead, but that looks rough. I'm cool with the idea of locking it down for now and maybe coming back to it in the future if IE11 usage is high enough :)

karelm:
@josem can this PR be closed + this branch deleted now?

josem:
Yes @karelm! Thanks for the nudge! 💯

PR #597, opened 2019-02-27
[PUB-1174] Welcome B4B Trial Modal

hamstu:
### Purpose

https://buffer.atlassian.net/browse/PUB-1174

image

#### Staging Deployment


packages/welcome-b4b-trial-modal/components/WelcomeB4BTrialModal/index.jsx
```diff
@@ -0,0 +1,108 @@
...
+  }, {
+    success: hideModal,
+    error: hideModal,
+  });
+};
+
+const WelcomeB4BTrialModal = ({
+  translations,
+  hideModal,
+}) => (<div style={{ position: 'fixed', zIndex: '3000' }}>
```
karelm: how come an explicit zIndex is needed here? isnt this taken care of by Popover component? (just wondering as I'm not sure 😄)

packages/web/index.jsx
```diff
@@ -17,7 +16,7 @@ store.dispatch({
 const ErrorBoundary = getErrorBoundary();
 
 render(
-  <ErrorBoundary FallbackComponent={BoundaryFallback}>
+  <ErrorBoundary>
```
karelm: Is this accidentally removed from here?

packages/web/components/ErrorBoundary/index.jsx
```diff
@@ -30,7 +32,11 @@ if (window._bugsnagConfig) {
 
 const getErrorBoundary = () => {
   if (BugsnagErrorBoundary) {
-    return BugsnagErrorBoundary;
+    return ({ children }) => ( // eslint-disable-line
+      <BugsnagErrorBoundary FallbackComponent={BoundaryFallback}>
```
karelm: I'm intrigued whether this would be better to be kept clean without the BoundaryFallback, because otherwise we cant put in multiple ErrorBoundary's around components?

karelm:
I've got a couple of questions in there, but overall this feels great to me 👍 💯 :shipit:

packages/welcome-b4b-trial-modal/components/WelcomeB4BTrialModal/index.jsx
hamstu: Ah, a holdover from copying the modal from another package. I don't know TBH! 😅  Will have to take a gander next time I'm working on a modal. Thanks for calling it out. 👍

packages/web/components/ErrorBoundary/index.jsx
hamstu: Good one! Let's table that for now since this change was only done to fix things locally. (Locally the current code tries to add the BoundaryFallback to a React.Fragment (wich throws a React error). This is from the way we have the conditional Bugsnag setup – and so there's no behavior change here; just a local bugfix.)

PR #601, opened 2019-02-28
Show message for Internet Explorer 11 and lower

josem:
### Purpose
Add a very simple message for IE11 and lower since they're not supported for now in New Publish: http://hi.buffer.com/e8752647df72.

### Notes
- I've implemented it using a simple approach with vanilla JS since I couldn't use React.

#### Staging Deployment


twanlass:
Added a small update to the copy @josem. Included your style adjustment suggestion as well @hamstu 👍🚀

PR #607, opened 2019-02-28
[PUB-1247] Contributors should not change Posting Schedule settings

anafilipadealmeida:
### Purpose
Contributors should not change Posting Schedule settings

### Notes
[PUB-1247](https://buffer.atlassian.net/browse/PUB-1247)


packages/shared-components/ScheduleTable/ScheduleTableCell/index.jsx
```diff
@@ -77,7 +79,7 @@ const TableCellContents = ({
           fontSize={'small'}
         />
       </div>
-      {RemoveButton(time, onRemoveTimeClick, buttonStyle)}
+      {RemoveButton(time, onRemoveTimeClick, buttonStyle, disabled)}
```
karelm: you could also call this like a component and make it more clear that its a react component you are rendering `<RemoveButton time={time} ... />` but it might require you to change the `RemoveButton = (time,...)` to `RemoveButton = ({time, ...`

karelm:
Just one small comment, but overall this looks great! Feel free to ship :shipit: 💪 💯

packages/shared-components/ScheduleTable/ScheduleTableCell/index.jsx
anafilipadealmeida: Thanks 🙌 Changed it, and the code readability improved a lot!

PR #626, opened 2019-03-06
[PUB-1242] Show calendar tab to team members

anafilipadealmeida:
### Purpose
Team Members should see the Calendar Tab. Used `isBusinessAccount` prop to flag profiles from team members (includes all Managers and Contributors of team accounts).

### Notes
[PUB-1242](https://buffer.atlassian.net/browse/PUB-1242)
Me and @karelm had talked about what to do in those cases where we have a FeatureLoader and we also want to make the feature available for team members. The solution would be to put the Team Member condition in the Fallback, which in this case would be something like this:
captura de ecra 2019-03-06 as 13 57 48
However, after confirming with Super that the Calendar tab is going to be available for all Pro users and up, and looking at how the validation on the other tabs was done in the code, I decided to remove the FeatureLoader, to keep consistency and readability and to optimize the code.

### DIFF
--- a/packages/shared-components/Tabs/index.jsx
+++ b/packages/shared-components/Tabs/index.jsx
@@ -14,7 +14,7 @@ const Tabs = ({
     if (!tab || !tab.props.tabId) return tab;
     return React.cloneElement(tab, {
       selected: selectedTabId === tab.props.tabId,
-      onClick: onTabClick,
+      onClick: tab.props.onClick || onTabClick,
       secondary,
     });
   })}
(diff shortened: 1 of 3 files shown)


packages/shared-components/Tabs/index.jsx
```diff
@@ -14,7 +14,7 @@ const Tabs = ({
     if (!tab || !tab.props.tabId) return tab;
     return React.cloneElement(tab, {
       selected: selectedTabId === tab.props.tabId,
-      onClick: onTabClick,
+      onClick: tab.props.onClick || onTabClick,
```
anafilipadealmeida: When I removed the `<FeatureLoader>` from the Calendar tab on the TabNavigation, the `onClick={() => openCalendarWindow(profileId)`stopped working. I discovered it was because of this piece of code. The FeatureLoader was unintentionally a child of `<Tabs>, so this onClick, instead of overwriting the child `<Tab>`, was in this case being added to `FeatureLoader`

PR #627, opened 2019-03-07
Updating grpc command to download binary and do verbose output

karelm:
### Purpose

Updating grpc command to download binary


albennett:
Looks great, thanks for deploying this! I created a branch for this change yesterday but paused when I was unsure whether it fixed my issue. 😄

karelm:
fair enough, I think at least this does fix the grpc binary running, so however you look at it, its probably a little better than where we are now, but if you need to sort more things out there @albennett please feel free to 👍

PR #629, opened 2019-03-07
Removes mini-cal feature

twanlass:
### Purpose
- Deprecate the mini calendar in Publish

#### Staging Deployment


twanlass:
Heya @hamstu – I'd love your eyes on this. We're looking to deprecate the mini calendar. After looking at usage and FullStory sessions it's clear this feature is causing more confusion than value by a landslide. 

I've removed the component in a few obvious places, though I've left a few references to `numberOfPostsByDate` in the reducer and tests in case this prop / data is used elsewhere? If not, it probably feels good to remove from there as well? 

Thanks in advance 🙏

packages/queue/index.js
```diff
@@ -62,7 +62,6 @@ export default connect(
         editMode: state.queue.editMode,
         editingPostId: state.queue.editingPostId,
         showCalendar: profileQueuePosts.showCalendar,
-        numberOfPostsByDate: profileQueuePosts.numberOfPostsByDate,
```
anafilipadealmeida: Hey @twanlass, ceiling cat in here 😄You should also delete the `onMiniCalendarMonthChange`:
https://github.com/bufferapp/buffer-publish/pull/629/files#diff-bc022e8a62326894e41126f80506e3e2R162

We only use `getNumberOfPosts` for the MiniCalendar. I would probably think if in the near future there's any planned feature that would benefit from it, and if not I would just remove everything. I can take another look at this tomorrow if you'd like, as I'm familiar with this part of the code 😄

twanlass: @anafilipadealmeida you're awesome! I'd love your 👀 on this tomorrow. I'll defer to do you on what else should be removed 🙇 🌮 🌮 🌮 🌮

twanlass:
Awesome stuff @anafilipadealmeida 🙌. Deploying Monday feels great 🚀

hamstu:
Late reply, but thanks so much for helping out here @anafilipadealmeida! 🌮 ❤️

PR #651, opened 2019-03-22
Improve error boundary / Blank screen crashes in UI

karelm:
### Purpose

When users have retweets which have has their original tweet deleted, we sometimes have a "null" tweet, and this causes the entire UI to crash.

To address this issue, we should build more resilience into the UI, and prevent the tweet from crashing the entire UI when this occurs.

I've built some additional resilience into our UI here:
http://hi.buffer.com/35d3bb092725

that will show just a blank tab, instead of a full blank screen, and allow some parts of the UI to continue to function.

#### Staging Deployment


twanlass:
Heya @karelm – is there an easy way to trigger this `ErrorBoundary`?

karelm:
You should be able to retweet a tweet, and then delete the original tweet @twanlass and that should evoke the bug... Or just ask advocacy, because its quite a common bug to see a fully blank screen sadly. Hope that helps, gl

karelm:
@twanlass since i've added a whole bunch of code to prevent the ui from crashing... you can force the ui to crash by adding a 
```throw new Error('blah');```
here: https://github.com/bufferapp/buffer-publish/blob/2a7d3e7574e505c0d4129ad3a9e877fd696b83b8/packages/shared-components/PostDragWrapper/index.jsx#L152 to crash all posts

and you can add 
```throw new Error('blah');```
here: https://github.com/bufferapp/buffer-publish/blob/2a7d3e7574e505c0d4129ad3a9e877fd696b83b8/packages/queue/components/QueueItems/index.jsx to crash the queue tab

twanlass:
Heya @karelm – wasn't quite able to crash the actual post items 🤔. Is `shared-components/PostDragWrapper/index.jsx` used with the new day view?

karelm:
Hey @twanlass sorry that wasnt working for you, if you throw the error here instead, while using a TextPost (so a tweet with no add-ons / images) it should break: https://github.com/bufferapp/buffer-publish/blob/49bc3a69f231a707c96efd5c6c7ac04bf31fb138/packages/shared-components/TextPost/index.jsx#L59

hope that helps, and thanks for sprucing up the error message 🙌

PR #655, opened 2019-03-26
Allow paid user to view all sent posts

albennett:
### Purpose
Allow paid or team member to view all of their sent posts. Currently, a user is only able to see sent posts from the last 30 days. 
https://buffer.atlassian.net/browse/PUB-1277
#### Staging Deployment


packages/sent/middleware.js
```diff
@@ -10,6 +11,10 @@ export default ({ dispatch }) => next => (action) => { // eslint-disable-line no
         args: {
           profileId: action.profile.id,
           isFetchingMore: false,
+          since: getState().appSidebar.user.is_free_user &&
+            !getState().profileSidebar.selectedProfile.business ?
+            null :
+            moment().subtract(10, 'years').unix(), // get all posts in the last ten years
```
albennett: In classic, there's a 30 day, 90 day or custom option for viewing Sent Posts. Since we don't have that option with New Publish, I added a high number to ensure we're getting all of their posts.

packages/profile-page/components/ProfilePage/index.jsx
```diff
@@ -136,13 +139,18 @@ const ProfilePage = ({
...
   page,
 }) => {
   const isPostsTab = ['queue', 'drafts', 'awaitingApproval', 'pendingApproval', 'pastReminders'].includes(tabId);
+  const isChildSentTab = tabId === 'analytics' && childTabId !== 'overview';
   const handleScroll = (o) => {
     const reachedBottom = o.scrollHeight - o.scrollTop === o.clientHeight;
-    if (reachedBottom && moreToLoad && isPostsTab && !loadingMore) {
-      onLoadMore({ profileId, page, tabId });
+    const since = isBusinessAccount || !features.isFreeUser() ?
+      moment().subtract(10, 'years').unix() : null;
```
hamstu: I don't think we need to pass `since` here. [Looking at the backend it appears this will default to `0`](https://github.com/bufferapp/buffer-web/blob/1eccee551a9552d04ab66bf0d8b9fdbce7d48a12/api/controllers/profiles_controller.php#L1548) which should have the same effect. Does that make sense? Perhaps you can run a quick test to see if that still works.

packages/profile-page/components/ProfilePage/index.jsx
```diff
@@ -136,13 +139,18 @@ const ProfilePage = ({
   onLoadMore,
   loadingMore,
   moreToLoad,
+  isBusinessAccount,
+  features,
   page,
 }) => {
   const isPostsTab = ['queue', 'drafts', 'awaitingApproval', 'pendingApproval', 'pastReminders'].includes(tabId);
+  const isChildSentTab = tabId === 'analytics' && childTabId !== 'overview';
```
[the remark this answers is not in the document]
hamstu: And maybe as an extra tweak here – nitpick, really! – we could keep the same variable and just add the extra condition to it:

```js
const isPostsTab = ['queue', 'drafts', 'awaitingApproval', 'pendingApproval', 'pastReminders'].includes(tabId) ||
  (tabId === 'analytics' && childTabId === 'sent');
```

Then we can still just check `isPostsTab` later. What do you think?

packages/sent/middleware.js
hamstu: See my previous comment — would love to know if we can just leave out passing `since` here 😄

packages/profile-page/components/ProfilePage/index.jsx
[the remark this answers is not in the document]
albennett: Hey! Great point - sometimes a child tab isn't needed with Sent posts because the Sent component shows on default under Analytics which is why I chose to explicitly check for 'overview' as a childTab. I'll still refactor with the `overview` & can leave a comment in the code :)

packages/profile-page/components/ProfilePage/index.jsx
albennett: Ah, great point. I noticed the total wasn't getting returned as the correct number which is why I thought it needed `since` but now I see that's intentional. 🗑

PR #665, opened 2019-04-05
Implementing instagram first comment

karelm:
#### Staging Deployment


packages/instgram-first-comment-modal/components/InstagramFirstCommentModal/index.jsx
```diff
@@ -0,0 +1,326 @@
...
+                  </BDSButton>
+                </div>
+              </div>
+            </div>
+          </Popover>
+        </div>
+      );
+    }
+    return (
+      <div style={{ position: 'fixed', zIndex: '3000', borderRadius: '4px', boxShadow: '0px 1px 4px rgba(0, 0, 0, 0.16)' }}>
```
MayaUribe: Is not that important, but most of these Popovers have a very similar structure, so I think it might be useful to have a generic component we could re-use here, what do you think?

packages/instgram-first-comment-modal/index.js
```diff
@@ -0,0 +1,46 @@
...
+      }(document, 'script', 'facebook-jssdk'));
+    },
+    canRequestMorePermission: (profileId) => {
+      const popup = window.open(`https://${getBaseURL()}/oauth/instagram/${profileId}/reconnect/auth`);
+
+      const interval = setInterval(() => {
+        if (popup.closed) {
+          clearInterval(interval);
+          dispatch(dataFetchActions.fetch({
+            name: 'single_profile',
```
MayaUribe: 🙌

packages/instgram-first-comment-modal/components/InstagramFirstCommentModal/index.jsx
karelm: Thanks @MayaUribe I'll circle back for engineering work for later 😄

packages/client-access/README.md
```diff
@@ -0,0 +1,3 @@
+# @bufferapp/client-access
+
+A component to expose client access information to the front end for a particular profile
```
hamstu: I'm coming in with very little context here, so take this with a grain of salt, but I'm afraid I don't understand what 'client access' means. Is there any chance we could clarify the purpose of this package, maybe put a small example here? Just for those passing through later.

packages/client-access/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@bufferapp/client-access",
+  "version": "2.0.0",
+  "description": "An example UI package",
```
hamstu: nitpick: has the example description

packages/client-access/reducer.js
```diff
@@ -0,0 +1,21 @@
+import { actionTypes as asyncDataFetchActionTypes } from '@bufferapp/async-data-fetch';
+import keyWrapper from '@bufferapp/keywrapper';
+
+export const actionTypes = {};
```
hamstu: I'd vote to nix this line since it's not needed.

packages/client-access/reducer.js
```diff
@@ -0,0 +1,21 @@
...
+
+const initialState = {
```
hamstu: You could nix this too and just put `export default (state = {}, action) => {` later.

packages/instgram-first-comment-modal/components/InstagramFirstCommentModal/index.jsx
```diff
@@ -0,0 +1,320 @@
...
+        inform user we need them to log into "profile name" to grant permissions for IG
+        -> click button + trigger auth flow
+        */
+        this.userNeedsSignin();
+      }
+    }
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('updateIFFirstCommentState', this.updateState);
```
hamstu: nitpick: was `updateIFFirstCommentState` meant to be `updateIGFirstCommentState`?

packages/instgram-first-comment-modal/components/InstagramFirstCommentModal/index.jsx
```diff
@@ -0,0 +1,320 @@
...
+        getLoginStatus();
+      };
+    } else {
+      getLoginStatus();
+    }
+    /* eslint-enable no-undef */
+  }
+
+  sendStateUpdate(message) {
+    window.dispatchEvent(new CustomEvent('updateIFFirstCommentState', {
```
hamstu: Haven't tested, but this might fail in IE 11 according to https://caniuse.com/#search=CustomEvent.

packages/instgram-first-comment-modal/components/InstagramFirstCommentModal/story.jsx
```diff
@@ -0,0 +1,21 @@
...
+import translations from '@bufferapp/publish-i18n/translations/en-us.json';
+
+import IGFirstCommentModal from './index';
+import PropTypes from 'prop-types';
+
+storiesOf('IGFirstCommentModal', module)
+  .addDecorator(checkA11y)
+  .add('should show welcome modal', () => (
+    <IGFirstCommentModal
+      translations={translations['welcome-paid-modal']}
```
hamstu: I think that may be the wrong translations key?

packages/instgram-first-comment-modal/components/InstagramFirstCommentModal/index.jsx
```diff
@@ -0,0 +1,320 @@
...
+import {
+  Popover,
+  Text,
+  LoadingAnimation
+} from '@bufferapp/components';
+import { TranslationReplacer } from '@bufferapp/publish-i18n';
+
+
+import {
+  BDSButton,
```
hamstu: We can use the real one now, from `@bufferapp/ui`, Thread incoming on this though, so no worries here. :D

packages/instgram-first-comment-modal/components/InstagramFirstCommentModal/index.jsx
```diff
@@ -0,0 +1,320 @@
...
+    }
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('updateIFFirstCommentState', this.updateState);
+  }
+
+  getFBLoginStatus () {
+    /* eslint-disable no-undef */
+    window.addEventListener('updateIFFirstCommentState', this.updateState);
```
hamstu: I'm curious, why the roundabout window event listening here instead of just directly calling `this.updateState` later?

packages/instgram-first-comment-modal/index.js
```diff
@@ -0,0 +1,46 @@
...
+      (function (d, s, id) {
+        if (d.getElementById(id)) return;
+        const fjs = d.getElementsByTagName(s)[0];
+        const js = d.createElement(s);
+        js.id = id;
+        js.src = '//connect.facebook.net/en_US/sdk.js';
+        fjs.parentNode.insertBefore(js, fjs);
+      }(document, 'script', 'facebook-jssdk'));
+    },
+    canRequestMorePermission: (profileId) => {
```
hamstu: nitpick: with this name it feels like this method would return a boolean, since it opens a popup perhaps a clearer name?

packages/instgram-first-comment-modal/util/index.jsx
```diff
@@ -0,0 +1,3 @@
+export const getBaseURL = () => {
+  return window.location.hostname === 'publish.local.buffer.com' ? 'local.buffer.com' : 'buffer.com';
```
hamstu: We do this logic in a lot of places! 

I know we've tucked some of it in that external [formatters package](https://github.com/bufferapp/buffer-js-publish-formatters), but it's annoying to have to work on that package outside publish, bump the version, update, etc. and so I end up doing this same thing too (co-locating the logic with the pkg that needs it.) 

I'm thinking we should have a bring the formatters back _into_ publish, or create a separate new utils one for stuff like this (just gotta watch it doesn't become the kitchen sink utils package!) 

PS. /thinking out loud here, not requesting changes right now!

packages/profile-sidebar/middleware.js
```diff
@@ -30,6 +30,11 @@ export default ({ dispatch, getState }) => next => (action) => {
       }));
       break;
     }
+    case 'single_profile':
+      dispatch(dataFetchActions.fetch({
+        name: 'single_profile',
```
hamstu: Just for consistency could we rename this to `singleProfile`, as all the other rpc endpoints are camelCase 😄

packages/client-access/README.md
karelm: client access tokens... sorry, will clarify that here. Thanks for pointing this out out Hamish 👍

karelm:
Hey @hamstu, I've checked the CustomEvent stuff, and it turns out... we don't support IE11, so its all good 😄thanks for raising that, and I've checked it in the browser, and its fine in edge, so hopefully should all be fine 🙌

packages/instgram-first-comment-modal/components/InstagramFirstCommentModal/index.jsx
karelm: I need to wait until Facebook has initialised before calling the this.updateState method. 
By wrapping it in an event listener we prevent the events from firing unless they are ready, and if someone cancels the modal dialog half way through facebook initializing / some event happening, then we unbind the event listener, and the event just fires and doesn't trigger a potential memory leak.

Don't know if what I mean is clear here... but essentially its because FB is async, and the modal may have been closed, so we want to prevent a potential memory leak

packages/instgram-first-comment-modal/util/index.jsx
karelm: I've moved it into formatters for now, just because theres a getURL package and it looked aligned with this, but agree we need some better solution in future

packages/instgram-first-comment-modal/components/InstagramFirstCommentModal/index.jsx
hamstu: Ah I see, I see! Thanks for that explanation – also makes more sense after seein the flow in your demo today! ✨

hamstu:
Sounds great @karelm! I had that thought too, wasn't sure if we had IE 11 in our support list — glad to hear that's not a blocker :)

PR #666, opened 2019-04-12
Update app sidebar with new 1.11.1 version

josem:
### Purpose
Updated package with this change: https://github.com/bufferapp/buffer-js-app-sidebar/blob/master/CHANGELOG.md#1111---2019-04-12. 

#### Staging Deployment


josem:
Hey @hamstu! I'm merging this.

The app is working fine but I think we need to update Jest to fix the problem in the tests. I've tried to go down that rabbit hole but it includes too much stuff so I think we can work on it once we're back from the retreat.

PR #667, opened 2019-04-12
Updated jest config so the test pass again

josem:
### Purpose
The problem was that jest was trying to execute the tests in node_modules. I think this PR can solve that. 

#### Staging Deployment

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -16,5 +16,8 @@ module.exports = {
   globals: {
     __PACKAGES__: '../packages',
   },
+  testPathIgnorePatterns: [
+    'package/sidebar/node_modules',
+  ],
   testURL: 'https://publish.local.buffer.com',
 };


jest.config.js
```diff
@@ -16,5 +16,8 @@ module.exports = {
   globals: {
     __PACKAGES__: '../packages',
   },
+  testPathIgnorePatterns: [
+    'package/sidebar/node_modules',
```
hamstu: Maybe we could put `'package/**/node_modules'` here?

hamstu:
Hey @josem, curious if this fixes the issue? I see you have `package` instead of `packages` (with the `s`), wouldn't that last one be the right way? 🤔

josem:
Yeah I think we were trying to run the tests from node_modules as well 🤔 . 

This seemed to do the trick but happy to chat more about a more solid solution.

PR #673, opened 2019-04-20
Add instagram advanced analytics

albennett:
#### Staging Deployment

### DIFF
--- a/packages/analytics/components/AnalyticsList/index.jsx
+++ b/packages/analytics/components/AnalyticsList/index.jsx
@@ -11,7 +11,6 @@ import { trackAction } from '@bufferapp/publish-data-tracking';
 import LockedProfileNotification from '@bufferapp/publish-locked-profile-notification';
 import getErrorBoundary from '@bufferapp/publish-web/components/ErrorBoundary';
 
-
 import Toolbar from '../Toolbar';
 import Notification from '../Notification';
 import ProfileHeader from '../ProfileHeader';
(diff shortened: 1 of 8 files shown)


packages/analytics/components/AnalyticsList/index.jsx
```diff
@@ -68,11 +68,17 @@ const AnalyticsList = ({
...
-          <SummaryTable />
+          {!isInstagramBusiness &&
+            <SummaryTable />
+          }
           <CompareChart />
-          <HourlyChart />
-          <AverageTable />
-          <PostsTable />
+          {!isInstagramBusiness &&
+            <div>
```
hamstu: To avoid an extra div here you could use `React.Fragment`.

packages/analytics/index.js
```diff
@@ -6,8 +6,10 @@ export default connect(
     profile: state.profileSidebar.selectedProfile,
     isLockedProfile: state.profileSidebar.isLockedProfile,
     isBusinessAccount: state.profileSidebar.selectedProfile.business,
+    isInstagramBusiness: state.profileSidebar.selectedProfile.isInstagramBusiness,
```
hamstu: 👌

packages/analytics/components/AnalyticsList/index.jsx
chayjay: Not 100% sure what you mean here, can you elaborate?

hamstu: For transparency, I shared with @chayjay in a DM: 

Yeah, great one @char! Thanks for asking. `React.Fragment` is a way of wrapping components (since you have to have a wrapper for multiple components when you're coding the structure) but it doesn't get outputted in the HTML.

PR #675, opened 2019-04-23
Grid - Adding validation to check if the user has business account.

MayaUribe:
### Purpose
Adding validation to check if the user has business account.

#### Staging Deployment

### DIFF
--- a/packages/tabs/index.js
+++ b/packages/tabs/index.js
@@ -10,7 +10,6 @@ import TabNavigation from './components/TabNavigation';
 // default export = container
 export default connect(
   (state, ownProps) => ({
-    isBusinessUser: state.appSidebar.user.is_business_user,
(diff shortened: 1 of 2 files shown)


packages/tabs/index.js
```diff
@@ -10,6 +10,7 @@ import TabNavigation from './components/TabNavigation';
 // default export = container
 export default connect(
   (state, ownProps) => ({
+    isBusinessUser: state.appSidebar.user.is_business_user,
```
MayaUribe: Hey @josem I'm not really sure why, but I added this line, because `state.profileSidebar.selectedProfile.business` is always returning true, so I added this one to check on the user specifically, what do you think?

josem: Hmm, interesting! I don't understand why it's always returning true to be honest, I'll investigate a little bit from my side 👍

PR #678, opened 2019-04-24
[PUB-1360] 🐲 Dragon Drop v2

hamstu:
### Purpose

See the [Jira card](https://buffer.atlassian.net/browse/PUB-1350).

Update the implementation of our post drag and drop. We now only support dragging posts into empty slots. This simplifies things very nicely!

### Notes
 
* Day View (and therefore also drag and drop) is now available to all plans.
* The version of React has been updated (also in the composer) to 16.8.3 since we needed a newer version to work with the updated version of `react-dnd`.
* This depends on the new API endpoint in https://github.com/bufferapp/buffer-web/pull/16090

#### Staging Deployment


packages/shared-components/PostEmptySlot/index.jsx
```diff
@@ -61,7 +68,10 @@ const getHoverMsg = (service) => {
         <span style={iconStyle}>
           <IconComponent />
         </span>
-        <span> { service === 'twitter' ? 'Schedule a Tweet' : 'Schedule a Post' } </span>
+        <span>
+          {' '}
```
karelm: how come this is needed here?

packages/shared-components/PostEmptySlot/index.jsx
```diff
@@ -85,16 +95,16 @@ class PostEmptySlot extends Component {
   }
 
   render() {
-    const { service, time, onClick } = this.props;
-    return (
+    const { service, time, onClick, focus } = this.props;
+    return ( // eslint-disable-next-line
```
karelm: do you need to disable the next line here?

hamstu: Ah yeah, it's just to hide the warning about `<div>`'s not being allowed to have event handlers; it should be a button for accessibility, but iirc that messed up a few styles, so leaving as this for now! But it would be great to improve for a11y in the future.

packages/shared-components/PostEmptySlot/index.jsx
hamstu: Oh I think this came from the _prettier_ auto formatting I ran on this file! Good catch. Should be able to go back to what was there before.

PR #685, opened 2019-04-29
[PUB-1408] Remove Business upgrade path from Pro trialists

chayjay:
### Purpose
Remove the 'Learn about Buffer for Business' option when users are trialing Pro Publish. 
#### Staging Deployment

### DIFF
--- a/packages/tabs/index.js
+++ b/packages/tabs/index.js
@@ -14,6 +14,7 @@ export default connect(
     isManager: state.profileSidebar.selectedProfile.isManager,
     selectedTabId: ownProps.tabId,
     selectedChildTabId: ownProps.childTabId,
+    onProTrial: state.appSidebar.user.trial.onTrial,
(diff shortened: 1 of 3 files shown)


packages/tabs/index.js
```diff
@@ -14,6 +14,7 @@ export default connect(
     isManager: state.profileSidebar.selectedProfile.isManager,
     selectedTabId: ownProps.tabId,
     selectedChildTabId: ownProps.childTabId,
+    onProTrial: state.appSidebar.user.trial.onTrial,
```
albennett: Hey @chayjay are we sure this is specifically for pro trial, or could it also be for business trial?

chayjay: Hey, @albennett this isn't specifically for Pro trials, it's for business trials as well. I have PR out right now to fix: https://github.com/bufferapp/buffer-publish/pull/688/files#diff-c8e7e161381d571dd8c7bdaa2fa9985fR17

PR #691, opened 2019-04-30
More fixes for infinite scroll

hamstu:
### Purpose

Ignore whitespace in the PR as there is some formatting in there

#### Staging Deployment


packages/profile-page/index.js
```diff
@@ -6,62 +6,71 @@ import { actions as dataFetchActions } from '@bufferapp/async-data-fetch';
...
-        default:
-          requestName = 'queued';
+export default hot(
+  connect(
+    (state, ownProps) => {
+      const { tabId, profileId, childTabId } =
+        getProfilePageParams({ path: ownProps.history.location.pathname }) ||
+        {};
+      // With analytics, the reducer state name doesnt match the tabId
+      const reducerName =
```
albennett: 💅

hamstu: Fun fact, I just used the 'Prettier' extension in VS Code to format this file! Just something I've been experimenting with. It's not set to auto-format on save or anything, but it's nice in a pinch when the file is getting complex. :)

packages/queue/middleware.js
```diff
@@ -91,19 +91,6 @@ export default ({ dispatch, getState }) => next => (action) => {
         notificationType: 'success',
         message: 'Okay, we\'ve deleted that post!',
       }));
-      /**
```
albennett: Do we no longer need to re-fetch because of the react hooks?

hamstu: We **do** need to re-fetch, but I moved the re-fetching further down in the middleware in a previous PR. Removing it here was part of resolving a merge conflict where it came back, I think.

Let me know if that makes sense! :)

PR #696, opened 2019-05-03
[PUB-1397] Integrate AppShell from `@bufferapp/ui`

hamstu:
### Purpose

image

#### Staging Deployment


packages/preferences/components/Preferences/index.jsx
```diff
@@ -1,8 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { Tabs, Tab } from '@bufferapp/publish-shared-components';
-import { Link, ArrowLeftIcon, Text } from '@bufferapp/components';
```
anafilipadealmeida: I wonder if it would be better to exclude the changes on this file from the AppShell and create a dedicated PR, since it's also affecting the scrolling. Also, I've deployed today replacing some components in this file, so it would be better to pull the latest master and see how they work together (I guess the "Back to Dashboard" button may need a tweak or two to get along with those changes

packages/preferences/components/General/index.jsx
```diff
@@ -1,7 +1,6 @@
 import React, { Fragment } from 'react';
 import { Divider } from '@bufferapp/components';
 import CloseAccount from '@bufferapp/close-account';
-import EditEmail from '@bufferapp/edit-email';
```
anafilipadealmeida: @hamstu How you would feel about creating an Engineer jira card to remove this component from the code, as well as the unused related modals?

hamstu: Great idea @anafilipadealmeida! Will do.

PR #697, opened 2019-05-03
[PUB-1382] Fix infinite loading on sent posts page

hamstu:
#### Staging Deployment


chayjay:
> @chayjay Are we keeping infinite load for sent or replacing it with a button (as it appears we have here)?

@hamstu we're keeping infinite loading, and we're adding a Load More button in case infinite loading doesn't work for some folx. Not sure why but it wasn't working for some cases of people using Windows 10.

PR #702, opened 2019-05-08
[PUB-1404] Add new properties and events for Appcues integration

hamstu:
### Purpose

As per @ecksteinmk's [comment on the Jira card](https://buffer.atlassian.net/browse/PUB-1404?focusedCommentId=25407&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25407) I've added the following for Appcues:

1. A new `orgTeamMemberUsage` property that contains the number of users in the organization. **This includes the owner.** We'll use this to know if they've added a team member (i.e., when the count is `>=` 2). This depends on https://github.com/bufferapp/buffer-web/pull/16143 being merged.
2. A new Appcues event when they add a new post in the composer: `Created Post`.
3. A new Appcues event when they view the analytics tab. `Viewed Analytics Tab`.

@karelm does this all look good to you?

#### Staging Deployment


karelm:
Hey @hamstu this looks good to me. I left a comment on the buffer-web pr about making this property an org user instead of team member, which might have implications here. But otherwise I'm happy with this work 🙂 💪🚀

What do you think on how I've separated the third party stuff into is own module? I thought it was useful to isolate it, hopefully it will prevent it from leaking into the rest of our codebase

PR #706, opened 2019-05-09
Fix sidebar drag and drop

anafilipadealmeida:
### Purpose
Fix sidebar drag and drop after the `react-dnd` library update.


packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
```diff
@@ -65,43 +75,49 @@ const profileTarget = {
...
 };
 
-class ProfileDragWrapper extends Component {
-
-  render() {
-    const {
-      connectDragSource,
-      connectDropTarget,
-    } = this.props;
-    const profileProps = this.props;
```
anafilipadealmeida: @albennett, it's quite risky to do `const profileProps = this.props`, in those cases it's better to use `const profileProps = {...this.props}`. Here's a great article about copying Objects: https://scotch.io/bar-talk/copying-objects-in-javascript and a simple example I put together: https://jsbin.com/davezuhaza/edit?js,console 🙂

packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
```diff
@@ -65,43 +75,49 @@ const profileTarget = {
...
+    connectDragSource,
+    connectDropTarget,
+    ...profileProps
+  }, ref) => {
+    const elementRef = useRef(null);
+    connectDragSource(elementRef);
+    connectDropTarget(elementRef);
+    useImperativeHandle(ref, () => ({
+      getNode: () => elementRef.current,
+    }));
```
karelm: Hey @anafilipadealmeida,

I don't quite know if this is the case or not here, but since I didn't know about useImperativeHandle I has a read of the docs for the Hooks Api and saw that they suggest you use forwardRef when you use that useImperativeHandle function: https://reactjs.org/docs/hooks-reference.html#useimperativehandle

I'm not sure about useImperativeHandle / forwardRef as I've not used them before, but thought I'd mention it to you incase a) this isn't necessary here, or b) this is something that is missing from this piece of code

If you could check whether thats needed or not, it would be appreciated 😀

packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
```diff
@@ -65,43 +75,49 @@ const profileTarget = {
...
-          <ProfileListItem
-            {...profileProps}
-          />
-        </div>,
-      ),
+const ProfileDragWrapper = React.forwardRef(
+  ({
+    connectDragSource,
+    connectDropTarget,
+    ...profileProps
```
karelm: does profileProps need to contain connectDragSource or connectDropTarget? as in the old implementation it contained those values, but in your version they are no longer being passed on

packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
karelm: Ignore me!!!! I found you were using React.forwardRef instead or spreading it out from React with the other props 😅 

Would you be able to make it use the spread version instead for consistency please?

packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
anafilipadealmeida: Nop 🙂 As long as we're passing an elementRef to the connectDragSource and connectDropTarget functions it works fine. Those functions accept a react component, an DOM element or a ref object. This approach keeps as profileProps only the ones needed for the ProfileListItem.

packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
```diff
@@ -47,12 +53,16 @@ const profileTarget = {
     // When dragging upwards, only move when the cursor is above 50%
 
     // Dragging downwards
-    if (dragIndex < hoverIndex && hoverClientY < hoverThird) {
+    // if (dragIndex < hoverIndex && hoverClientY < hoverThird) {
```
albennett: 🗑

packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
albennett: Haha I did the exact same thing @karelm with useImperativeHandle and was about to add the same comment! 😂

packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
anafilipadealmeida: Great catch! With so many comments on the code, I totally missed it 😅

packages/profile-sidebar/components/ProfileDragWrapper/index.jsx
anafilipadealmeida: Thank you so much for all the input here! I'm gonna make the decision to deploy as it is right now, and leave myself a task to revisit it later on the next cycle 😄. I still have a task depending on this change, and since there are other people also working on the sidebar (for instance the search profile feature), I prefer not to hold it much longer!

PR #713, opened 2019-05-14
[PUB-1390] Prevent users from unlocking team accounts by dragging

anafilipadealmeida:
### Purpose
Prevent users from unlocking blocked team accounts by dragging them.

### Notes
[PUB-1390](https://buffer.atlassian.net/browse/PUB-1390)


packages/profile-sidebar/reducer.js
```diff
@@ -43,14 +43,54 @@ const moveProfileInArray = (arr, from, to) => {
...
+  changing only 1 property, i.e., adding the profile to only 1 array */
+  const {
+    enabledProfiles,
+    lockedProfiles,
+    teamProfiles,
+  } = reorderedProfiles.reduce(
+    (acc, cur) => {
+      /* If the service_type is business (team) and not profile,
+      it goes to the teamProfiles array. */
+      if (cur.service_type === 'business') {
```
anafilipadealmeida: We need to send a prop through the API that flags a profile that has team members. I'm using the `service_type === 'business'` at this point as an example, but has to be replaced by the real prop

hamstu: Perhaps the organization role they have on the profile could be used here? https://github.com/bufferapp/buffer-web/blob/234c20b238a19b168e9080d79dc298e9ff4ac502/shared/models/profile_model.php#L2878

anafilipadealmeida: The organization role does not provide a unique value to profiles with team members. José put this branch in place: add-has-team-members-profile but it's still not quite returning the right prop, I'll have to dig a bit more!

hamstu: Thanks for that context @anafilipadealmeida!

Would it make sense to just check if the profile is _not owned by the current user_? This would mean that it's a 'team profile', right?

PR #717, opened 2019-05-14
Instagram first comment modal for New Publish users

karelm:
### Purpose

[PUB-1365](https://buffer.atlassian.net/browse/PUB-1365) -  Implementing Welcome to Instagram first comment modal

Users should see this modal if they have the first_comment feature and don't have the readmessage `paid_users_to_new_publish_ig_first_comment`

or can be triggered via the ?mk=instagram-new-first-comment-user-modal query param

the readmessage is automatically triggered when they see this, so dismiss / not it will disappear after first viewing to ensure its not overbearing on the end user

### DIFF
--- a/packages/modals/middleware.js
+++ b/packages/modals/middleware.js
@@ -15,6 +15,7 @@ import {
   getShowModalValue,
   resetShowModalKey,
   shouldShowInstagramFirstCommentModal,
+  shouldShowInstagramNewFirstCommentUserModal,
 } from './util/showModal';
 
 export default ({ dispatch, getState }) => next => (action) => {
@@ -31,6 +32,9 @@ export default ({ dispatch, getState }) => next => (action) => {
       }
       if (shouldShowInstagramFirstCommentModal()) {
         dispatch(actions.showInstagramFirstCommentModal());
+      }
+      if (shouldShowInstagramNewFirstCommentUserModal()) {
+        dispatch(actions.showInstagramNewFirstCommentUserModal());
       }
       if (shouldShowWelcomeModal()) {
         dispatch(actions.showWelcomeModal());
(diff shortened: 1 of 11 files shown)


packages/modals/middleware.js
```diff
@@ -32,6 +33,9 @@ export default ({ dispatch, getState }) => next => (action) => {
       if (shouldShowInstagramFirstCommentModal()) {
         dispatch(actions.showInstagramFirstCommentModal());
       }
+      if (shouldShowInstagramNewFirstCommentUserModal()) {
```
albennett: When working on the trial expired modals, I realized we don't necessarily need to add these modals to the `APP_INIT` case. The `shouldShowWelcomeModal` etc are getting sent as params from the `NewPublishRollout.php` file and we're matching the params as `getShowModalKey` in the utils folder. (The exception is the `shouldShowUpgradeModal` which is getting triggered in Org Admin).

Since we're deciding whether to show the IG modal on user data load instead of params, we don't really need to check in the `APP_INIT`. It's def not hurting anything though!😀

packages/modals/util/showModal.js
```diff
@@ -29,6 +29,9 @@ export const shouldShowStealProfileModal = () =>
 export const shouldShowInstagramFirstCommentModal = () =>
   getShowModalKey() === 'instagram-first-comment-modal';
 
+export const shouldShowInstagramNewFirstCommentUserModal = () =>
```
albennett: Same as above. Not sure if this is needed since we're getting modalKey as a query in server/index.js https://github.com/bufferapp/buffer-publish/blob/master/packages/server/index.js#L267

packages/modals/middleware.js
karelm: Thats a great one @albennett I felt that it would be worth being able to trigger the modal via query string here so it can be tested and easily reviewed. I know it adds a bit more code, but felt the utility was worth the additional code. You do make a great point though 🤔 I will think about this a little more because its a lot of template code aswell, so we can likely adjust this somehow too 👍

karelm: Thanks @albennett for raising this, I did run some checks and I think that for testing purposes this is probably okay to leave here as it gives us a way to trigger the modal from the query string and doesn't add a major overhead. 

I think we could probably simplify things here and not run multiple expressions, but I'm okay with this for the time being. If you have strong feelings about leaving it here, I'm also happy to remove it 😄

PR #720, opened 2019-05-15
[PUB-1463] Update ui components in preferences

anafilipadealmeida:
### Purpose
Update Preferences' button and text components to BDS. Why? It's a good place to start this migration to the new design system, and it’s gonna have more relevance with the easier access through the new AppShell.

### Notes
[PUB-1463](https://buffer.atlassian.net/browse/PUB-1463)
The modals used in the Preferences page were still not updated: a couple of them are related to changing passwords and emails. There's been a discussion about removing some of those https://threads.com/34366424587, and right now some of the modals are not even being used (the buttons to edit email and password are redirecting to core).

### DIFF
--- a/packages/change-password/components/ChangePassword/index.jsx
+++ b/packages/change-password/components/ChangePassword/index.jsx
@@ -1,24 +1,29 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Text } from '@bufferapp/components';
-import { Button } from '@bufferapp/ui';
+import { Button, Text } from '@bufferapp/ui';
 import { Row } from '@bufferapp/publish-shared-components';
 import Modal from '../Modal';
 
+const rowStyle = {
+  display: 'flex',
+  justifyContent: 'space-between',
+  alignItems: 'center',
+};
+
 const ChangePassword = ({ showModal, onRequestOpenModal, onRequestCloseModal, onSubmit }) => (
-  <Row>
+  <div style={rowStyle}>
     <div
       style={{
         marginRight: '1rem',
       }}
     >
-      <Text color={'black'} size={'mini'}>Password</Text>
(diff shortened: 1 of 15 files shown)


packages/change-password/components/ChangePassword/index.jsx
```diff
@@ -1,24 +1,29 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Text } from '@bufferapp/components';
-import { Button } from '@bufferapp/ui';
+import { Button, Text } from '@bufferapp/ui';
 import { Row } from '@bufferapp/publish-shared-components';
 import Modal from '../Modal';
 
+const rowStyle = {
```
karelm: seeing as this Row config appears to be repeating, is there value in creating a component to do this consistently instead of writing it over + over again?

anafilipadealmeida: Great question! There is one `<Row>` component that we were using before here, but it doesn't quite fit right now. The only reason I decided not to create a component here is because once we implement the new Card component (yet to build) in the Preferences, it will replace the need for this extra style.

karelm: okay, no worries just thought I'd ask if it made sense, thats fine by me 😄

anafilipadealmeida: It's great that you asked! We might even need to implement some kind of row component again! The Preferences don't have a clear design spec (and have elements that belong to core and will probably be moved), which makes every decision seem a bit temporary 😅

anafilipadealmeida: Changed my mind @karelm 😂 I've confirmed that the `<Row>` component, unlike what I thought, it's not being used at all, so I'm gonna use it instead and change its css!

PR #721, opened 2019-05-15
Updating the css hash to use the context of the root directory to run tests against and create hashes

karelm:
### Purpose

When webpack runs and hashes get created for css files, we can provide an extra context parameter to define the root that the hash is based on, to provide greater consistency.


anafilipadealmeida:
Any reason why this wasn't happening before @karelm?

karelm:
> Any reason why this wasn't happening before @karelm?

there appears to be 2 versions of css-loader being used, and it looks like its having some side-effects. but i'm not sure when that started happening :( @anafilipadealmeida

PR #745, opened 2019-05-24
Using musl instead of glibc for grpc binary

karelm:
### Purpose

Using MUSL instead of glibc for grpc binary


hamstu:
👍 I remember changing this for myself once before. 

Hopefully this isn't whack-a-mole and just starts causing issues for others. Let's keep a close eye on it after merge, perhaps let the team know to tell us if they see any grpc problems coming up after?

karelm:
@hamstu the image we use for docker is based on alpine, and alpine changed over to using musl for greater cross compatibility, the problem last time was that the team was using the old docker container, but I think by now everyone is likely on the newer container version... so this issue shouldn't happen anymore, but I will let the team know before I merge this one in. Thanks for raising that 😀

PR #773, opened 2019-06-10
[PUB-1534] Remove beta-redirect package and references

albennett:
## Description
Remove beta-redirect package and references.
https://buffer.atlassian.net/browse/PUB-1534

## Context & Notes
Now that everyone can publish, there is no need to check and send back to classic. This may also help with the redirect loop issue.

@hamstu -- I had a few questions since I saw you added the JIRA card 😄 Do you know if this will affect users who have enterprise accounts (I didn't think we were switching them to NP yet?). If someone has the rolled back feature, would they be allowed in NP now?

## Checklist

#### Staging Deployment


hamstu:
Thanks @albennett! Sorry for the delayed reply. I think we can remove this regardless of enterprise. They should all be in classic, and even if they did slip into NP they could switch back. Re: rolled back features - customers who have that won't be rolled out or automatically redirected to NP, but it doesn't prevent them from hitting the URL directly or switching (I _think_ we still show the switcher, not sure.)

Let me know if that feels ok to you!

PR #775, opened 2019-06-11
[PUB-1402] Making the queue the default view if it's not a valid tab

MayaUribe:
## Description
Currently, we only have a validation for a user to be able to access a tab, in the Tab itself, and not in the Tab Content, which leads to certain issues, for example, if the user changes a profile, is capable of seeing a "new" set of valid tabs, but still see the content even if it's invalid (examples of this behavior are grid and past reminders tabs)

**Example Gif**
Screen Recording 2019-06-12 at 01 14 PM

With these changes, we are validating both the Tab and the Content in a single file, which will make it easier for us in the future if we decide to add a new tab.

## Context & Notes
When a user has an Instagram profile selected and is wether in Past Reminders or Grid tab, when switching profile to a non-instagram account they should see the queue by default instead.

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/profile-page/components/ProfilePage/index.jsx
+++ b/packages/profile-page/components/ProfilePage/index.jsx
@@ -67,7 +67,7 @@ const tabContentStyle = {
  * @param isFreeUser
  * @param onChangeTab method to change tab
  */
-const handleTabChange = (
+const verifyTab = (
   tabId,
   profileId,
   selectedProfile,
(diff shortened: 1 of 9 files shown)


anafilipadealmeida:
@MayaUribe are you thinking about doing a similar thing for other tabs, like Drafts and Pending Approval? I'm wondering if there would be a way to solve this on a higher level, like in the routes package, so we wouldn't have to replicate the same validation we are already doing on the TabNavigation component 🤔

MayaUribe:
> @MayaUribe are you thinking about doing a similar thing for other tabs, like Drafts and Pending Approval? I'm wondering if there would be a way to solve this on a higher level, like in the routes package, so we wouldn't have to replicate the same validation we are already doing on the TabNavigation component 🤔

Thank you for the comment here @anafilipadealmeida! I started making changes on the route package, but I didn't want to mess with the rest, but I think I might do it because there are many cases and maybe with a bit of refactor we have all cases covered in a single file 🤔

PR #781, opened 2019-06-18
[PUB-1546] Display start trial modal on instagram first comment toggle

albennett:
## Description
Display the start trial modal when a free user toggles ig first comment. Start trial in the background, with the composer and current state still open. 

To get the trial starting without a refresh, we had to add logic to send data from publish to the composer to update user data.

If a user is free and not able to start a trial, we are hiding the ig toggle for now. It was trickier to upgrade without a refresh, since we're hitting different endpoints & entering stripe cc data (more context in JIRA card). Adding this as a non-blocker separate task.

## Context & Notes
This change is needed to improve the instagram first comment upgrade paths
Jira card: https://buffer.atlassian.net/browse/PUB-1546
Related ig first comment modal PR: https://github.com/bufferapp/buffer-publish/pull/779

https://www.loom.com/share/76b5a7e65a6246079ef6c5d0fd2e6725
## Checklist

#### Staging Deployment


packages/composer/composer/components/Composer.jsx
```diff
@@ -137,6 +141,35 @@ class Composer extends React.Component {
     ComposerActionCreators.toggleAttachment(this.props.draft.id, AttachmentTypes.RETWEET);
   }
 
+  onToggleComment = (e, commentEnabled) => {
```
karelm: this name seems unrelated to whats happening here now @albennett can this be changed to something more appropriate? (also, glad to see you found the trigger interaction work 👍)

packages/composer/composer/stores/AppStore.js
```diff
@@ -806,6 +808,16 @@ const markAppAsNotLoaded = () => {
   state.appState.isLoaded = false;
 };
 
+// There's a CORS error when trying to reset the s3UploadSignature. Leaving signature
+// as is until refresh
+const resetUserData = (userData) => {
```
karelm: does this function already exist elsewhere? If so can we export a version of this function and import it in those places?

albennett: Hey @karelm thanks for leaving some comments! There is a method `loadInitialUserData` that gets called on init. I ended up not using this method for two reasons: 
1. There's a CORS error when resetting the s3 signature 
2. The way the method is setting the data causes the buffer loading image to appear briefly, which isn't ideal for this case. 

I'll take a look though to see if there's a better way to refactor! 😄

packages/composer/composer/components/Composer.jsx
albennett: Hmm, all the actions (show modal or show input) happen on toggle comment, so I could rename `handleActionOnToggleComment

karelm: Okay, thats fair enough, maybe dont rename it 😄 I think this is whats confused me: 
```
          action: 'SHOW_IG_FIRST_COMMENT_PRO_TRIAL_MODAL',
```
It seems like a specific instagram modal is being shown where perhaps we should do this 
```
if(canStartProTrial || showProUpgradeModal) {
    AppActionCreators.triggerInteraction({
        message: {
          action: 'SHOW_COMMENT_UPGRADE_MODAL',
          canStartProTrial,
          showProUpgradeModal,
        },
      });
} else  {
```
then the intention is clear that we are showing an upgrade modal or triggering the first comment code.
I hope this makes sense, if not please let me know. Also, this is just a suggestion, so no change necessary unless you think it appropriate. I think the modal action name left me a bit unsure, and now I get that its just launching modals for different circumstances. So perhaps a comment will suffice 😀

packages/composer/composer/components/Composer.jsx
```diff
@@ -951,7 +977,7 @@ class Composer extends React.Component {
                 />
               </div>
               <div
-                style={{ display: 'flex', alignItems: 'center', marginLeft: '9px' }}
+                style={{ display: 'flex', alignItems: 'center', marginLeft: '8px' }}
```
karelm: would it make more sense extracted this into the css file?

packages/composer/composer/components/Composer.jsx
```diff
@@ -939,6 +960,11 @@ class Composer extends React.Component {
         {shouldDisplayFirstCommentSection(draft.commentEnabled) &&
         <div>
           <div className={styles.toggleCommentContainer}>
+            { (this.props.isOnProTrial || this.props.isFreeUser) &&
+              <div style={{ marginRight: '8px' }}>
```
karelm: would it make more sense extracted this into the css file?

albennett:
Hey @josem this does not include the upgrade if user cannot start trial. It looks like backend changes need to be made to not cause a refresh. There's some comments in the JIRA that give more context. Currently adding it as a separate task so it's not blocking. Hiding the toggle for those users for now.

PR #782, opened 2019-06-18
Fix issue when resetting drafts in composer

josem:
Fix issue reported on Slack by Karel: https://buffer.slack.com/archives/C151RRDL1/p1560849082052000

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/composer/composer/stores/ComposerStore.js
+++ b/packages/composer/composer/stores/ComposerStore.js
@@ -117,12 +117,9 @@ const softResetState = () => {
   // The only data to preserve is the enabled state of each draft
   newState.drafts = newState.drafts.map((newDraft) => {
     const { isEnabled, editorState } = state.drafts.find(({ id }) => id === newDraft.id);
-
-    return {
-      ...newDraft,
-      isEnabled,
-      editorState: resetEditorContents(editorState),
-    };
+    newDraft.isEnabled = isEnabled;
+    newDraft.editorState = resetEditorContents(editorState);
+    return newDraft;
   });
 
   state = newState;


packages/composer/composer/stores/ComposerStore.js
```diff
@@ -117,12 +117,9 @@ const softResetState = () => {
   // The only data to preserve is the enabled state of each draft
   newState.drafts = newState.drafts.map((newDraft) => {
     const { isEnabled, editorState } = state.drafts.find(({ id }) => id === newDraft.id);
-
-    return {
-      ...newDraft,
-      isEnabled,
-      editorState: resetEditorContents(editorState),
-    };
+    newDraft.isEnabled = isEnabled;
```
karelm: Just curious here @josem but why dpes the prop spreading need to be removed here?

josem: So I kept it but instead of generating a generic object I return the draft object now so I can still call the validation methods in the Draft class if that makes sense.

karelm: I've been working in this area, and was wondering whether a better approach to this is to create a `new Draft()` instead? I'm not sure it matters too much, but thought I'd raise the question :)

josem: Hmm, yeah, I think it'd respect more the original intention there. I'll work on that change 👍, thanks so much!

karelm: Definitely no urgency at all just a small suggestion really :)

josem: Ah sorry, nvm, I think in the getInitialState method we already grab new Draft objects so newDraft is a new Draft object here.

PR #800, opened 2019-06-27
[PUB-1609] Path for new business trialists' empty dashboard

chayjay:
## Description

Adding new path for business trialists to hit empty dashboard when they have no social accounts connected.

## Context & Notes
This change is required to ensure Business Trialists have a smooth on-boarding experience.
JIRA: https://buffer.atlassian.net/browse/PUB-1609?atlOrigin=eyJpIjoiMDdhMjhjZGRmMTJmNDI1MmJmYzFhZjEwNDI5NDdjMTAiLCJwIjoiaiJ9

http://hi.buffer.com/6e1662861041
## Checklist

#### Staging Deployment


packages/profile-sidebar/middleware.js
```diff
@@ -70,6 +70,9 @@ export default ({ dispatch, getState }) => next => (action) => {
         path,
       });
       const profiles = getState().profileSidebar.profiles;
+      const { productFeatures: { planName } } = getState();
+      const onTrial = getState().appSidebar.user.trial.onTrial;
+      const isOnBusinessTrial = planName === 'business' && onTrial;
```
anafilipadealmeida: Would it make sense to move the isOnBusinessTrial logic to the userParser and use just the result coming from the user request? Something like `isOnBusinessTrial: isOnBusinessPlan(userData.trial_plan)` on userParser.js

chayjay: @anafilipadealmeida, I think that makes total sense! It'll allow me to clean up the logic in other places too. I'll work on it right away ☺

chayjay:
Hey @anafilipadealmeida, would love to circle back to this with you tomorrow - I think the logic here will need to be changed to address a race condition I keep running into. I can only get the new path to render in like 1 in 10 cases. I believe this is because isOnBusinessTrial is being evaluated as undefined most of the time due to the user data not being available at the time of evaluation.

anafilipadealmeida:
Hey @chayjay, great catch! 
You probably don't have any problems in the modals/middleware, as you only need the prop once the user request was successful, after `user_${dataFetchActionTypes.FETCH_SUCCESS}`
However, in the profileSidebar, the first user request was still not made. I assume it is because the first user request is made on the appSidebar, an external package that is rendered in the Sidebar component. Even though the user request is made on APP_INIT, it seems to be the last of the initial requests to be made.

anafilipadealmeida:
@hamstu, summing up: we need to have access to the user data in the profile-sidebar middleware, to know if the user is on a Business Trial. However sometimes the profiles request was happening before the user request. Our solution: instead of making the profiles request on `APP_INIT`, to do it after `user_${dataFetchActionTypes.FETCH_SUCCESS}`. It seems to be working, and we are assuming there's not a situation where the user request is not done on APP_INIT and we would still need the profiles request to happen. However we'd love to hear your thoughts!

MayaUribe:
@anafilipadealmeida @chayjay that solution you came up with makes a lot of sense to me, I think that as long as we are making the `fetch->user` request on `APP_INIT`, this is looking good to me 😄

chayjay:
@hamstu thank you so much!! I made changes based on your recommendations. How does it look now?

hamstu:
Looks great @chayjay! 🙌

MayaUribe:
Hey there @chayjay @anafilipadealmeida I deployed my changes with the Tooltip and merged them into this PR 😄, so I believe this is our last step to finish up this task 🚀

PR #815, opened 2019-07-12
[PUB-1390] Change logic drag locked team profiles

anafilipadealmeida:
## Description
Prevent users from unlocking blocked team accounts by dragging them.

## Context & Notes
If a profile is locked and the user is not its owner (it means the owner downgraded the plan), by dragging the profile it will not get unlocked. These types of profiles will never be able to be moved to the top of unlocked profiles, because we have a profile order in the sidebar: first unlocked, then locked profiles.

[PUB-1390](https://buffer.atlassian.net/browse/PUB-1390)

locked

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/profile-sidebar/reducer.js
+++ b/packages/profile-sidebar/reducer.js
@@ -46,14 +46,53 @@ const moveProfileInArray = (arr, from, to) => {
 const handleProfileDropped = (profiles, action) => {
   const { profileLimit, hoverIndex, dragIndex } = action;
   const reorderedProfiles = moveProfileInArray(profiles, dragIndex, hoverIndex);
-  // add profiles after limit as disabled
-  if (reorderedProfiles.length > profileLimit) {
-    reorderedProfiles.map((profile, index) => {
-      profile.disabled = index >= profileLimit;
-      return profile;
-    });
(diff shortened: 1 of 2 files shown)


MayaUribe:
This looks great @anafilipadealmeida, I just had an idea that I think we could consider to add sometime in the future, we currently have 3 types of profiles: enabled, locked and team, and it's pretty easy to differentiate the first two, however, I think it'd be a good thing to have some icon or something of that sort so the user can see which profiles are a "team profile" and maybe even from which team those are from, this is just a random idea 😄

anafilipadealmeida:
Hey @MayaUribe! In reality we only have 2 types of profiles: locked and unlocked. For this specific functionality, what we are saying is that team members profiles, when dragged, keep their state (if they are locked they stay locked, if they are unlocked they stay unlocked). However, I agree with you that on the sidebar itself users cannot distinguish between profiles they own and not. A couple of months ago I had a similar conversation w/ Joel and Super, but we realized that this problem would be solved by switching between accounts. But in the meanwhile, maybe we can bring it up to Super and Lauren again, in case there's a simple design solution for it

PR #832, opened 2019-07-18
Avoid sending something different than a number as a page for pagination

josem:
## Description
Fix https://buffer.atlassian.net/browse/PUB-1662.

## Notes
Since there's a lot of indirection on where the page param comes from, I thought it's easier and more effective to be more defensive about this value so if we have something different than a number we set it to 1.

#### Staging Deployment

### DIFF
--- a/packages/server/rpc/sentPosts/index.js
+++ b/packages/server/rpc/sentPosts/index.js
@@ -11,7 +11,7 @@ module.exports = method(
       page = 1;
     }
 
-    rp({
+    return rp({
       uri: `${process.env.API_ADDR}/1/profiles/${profileId}/updates/sent.json`,
       method: 'GET',
       strictSSL: !(process.env.NODE_ENV === 'development'),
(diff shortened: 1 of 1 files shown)


packages/server/rpc/sentPosts/index.js
```diff
@@ -6,8 +6,12 @@ const { buildPostMap } = require('./../../formatters/src');
 module.exports = method(
   'sentPosts',
   'fetch sent posts',
-  ({ profileId, page }, { session }) =>
-    rp({
+  ({ profileId, page }, { session }) => {
+    if (typeof page !== 'number' || isNaN(page)) {
```
karelm: @josem I think something else to consider here is why is page ever set to anything different than a number, and if it is, can we log it to an error log somewhere?

josem: Yeah sorry, I agree, but in the time we gave to this task that's all I could do. I couldn't find where it comes from since apparently is dynamically called and the callback is passed as an attribute in different places. 

I'll add a comment to the ticket with that for future investigation 👍

karelm: no worries, just more something to raise for future investigation really 👍 Great work getting this fixed up though

PR #837, opened 2019-07-22
[PUB-1599] Add tracking for posts and drafts

albennett:
## Description
Add segment tracking for publish specific events mentioned in this google doc: https://docs.google.com/spreadsheets/d/1nPkN-5_qCtXgbqWEtWnMm8n_vwDZrA6bhMbn7065MZU/edit#gid=50773407

Specific Tracking includes:
Draft Approved & Submitted & Deleted
Post Created

## Context & Notes
For now, we've decided to implement this on the frontend. We need to call `init` and pass in global account data to track correctly with segment. 
https://github.com/bufferapp/buffer-dev-config/pull/90

JIRA: https://buffer.atlassian.net/browse/PUB-1599
Currently using a similar analytics-middleware package as Analyze is using. 

## Checklist

#### Staging Deployment


packages/globalAccount/middleware.js
```diff
@@ -0,0 +1,27 @@
...
+
+  switch (action.type) {
+    case 'APP_INIT':
+      dispatch(asyncDataFetch.fetch({
+        name: 'globalAccount',
+      }));
+      break;
+    case `globalAccount_${dataFetchActionTypes.FETCH_SUCCESS}`:
+      dispatch(analyticsActions.init(state.globalAccount._id, {
+          // organizationId: state.globalAccount.organizationId,
```
albennett: Hey @djfarrelly , I noticed the returned globalAccount object doesn't include an organizationId or a name. I'm getting the name from the appSidebar, is that ok? Do we need an organizationId?

packages/analytics-middleware/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@bufferapp/analytics-middleware",
```
karelm: should this be named publish-analytics-middleware?

packages/composer/composer/utils/DataImportUtils.js
```diff
@@ -358,6 +360,8 @@ const DataImportUtils = {
         commentEnabled: meta.commentEnabled,
         commentText: meta.commentText || null,
         shopgridLink: meta.shopgridLink || null,
+        tabId: meta.tabId || null,
```
karelm: do we need the childTabId too? I'm not sure, just asking really

packages/composer/composer/utils/TrackingUtils.js
```diff
@@ -0,0 +1,39 @@
+const getComposerSource = ({ tabId, emptySlotMode }) => {
+  let sourceName = null;
+  switch (tabId) {
+    case 'queue':
+      sourceName = emptySlotMode ? 'queue_list' : 'queue';
```
karelm: Hey @albennett, I'm just wondering, why does emptySlotMode change the sourceName here? surely the source is still the same, just what we are displaying is different. Its still the queue in both these scenarios, it's just that one is empty, and one has items?  If you would be able to let me know more, I'd really appreciate it 😀

packages/composer/composer/utils/WebAPIUtils.js
```diff
@@ -79,18 +79,29 @@ const WebAPIUtils = {
...
 
       return API.post(endpoint, formattedData)
         .catch(() => ({
           success: false,
           message: `Whoops, Buffer's servers couldn't be reached to save the
             update, sorry about that. Would you be up for trying again?`,
         }))
-        .then((response) => Object.assign(response, { serviceName }));
+        .then((response) => {
+          if (!updatePost) {
```
karelm: Is there a better place for this logic to live than here? It seems strange that trackSegmentAction is happening in the then of the webAPIUtils file, but I don't really know if theres anywhere more appropriate, but thought i'd ask at least 😄

packages/analytics-middleware/package.json
albennett: Hey :) Possibly! I mainly carried over the analytics-middleware package from analyze as we'll prob try to share the same version in the future. (We decided to temp not as we figure out what each team needs) In the meantime, I could add publish to make it more specific.

packages/composer/composer/utils/TrackingUtils.js
albennett: Great question! I can add a comment in the code to make this more clear. emptySlotMode is a bool in the composer that tells whether a user opened up the composer from an empty slot or from the main input. I think we needed this in the past because the button text is diff depending on the two.

With the tracking source, we want to know whether a user created a post from opening up the composer in the empty slot or from the main input. The source names come from this doc: https://docs.google.com/spreadsheets/d/1nPkN-5_qCtXgbqWEtWnMm8n_vwDZrA6bhMbn7065MZU/edit#gid=50773407

packages/composer/composer/utils/WebAPIUtils.js
albennett: Great question! I agree, it's a bit strange. I chose this because it's the closest to the actual event and with omnibox, it could return more than one response. I'm def open to suggestions though if you think something else makes more sense. 😀

Some added context: We may refactor this in the near future to track in the backend instead of the frontend. We initially decided to track publish specific events in the frontend, mainly because of time. However, post created is tricky regarding calendar & extension.

packages/drafts/middleware.js
```diff
@@ -53,18 +69,32 @@ moves from approval tab to drafts if needsApproval false)
...
-    case `changeDraftStatus_${dataFetchActionTypes.FETCH_SUCCESS}`:
+    }
+    case `changeDraftStatus_${dataFetchActionTypes.FETCH_SUCCESS}`: {
       dispatch(notificationActions.createNotification({
         notificationType: 'success',
         message: 'We\'ve successfully moved this draft!',
       }));
+      const draft = action.result ? action.result.draft : {};
+      /* this is also called when a draft in approval is moved back into drafts, which
+       is why we need to check the needs_approval bool */
```
djfarrelly: 👍 Useful context!

packages/analytics-middleware/middleware.js
```diff
@@ -0,0 +1,33 @@
...
+      if (window.analytics) {
+        window.analytics.identify(action.userId, action.payload);
+      }
+      break;
+    case actionTypes.TRACK_EVENT:
+      if (window.analytics) {
+        window.analytics.track(
+          action.eventName,
+          Object.assign({
+            product: window.PRODUCT_TRACKING_KEY,
```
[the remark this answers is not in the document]
albennett: Hey :) We are also including it in our server index file https://github.com/bufferapp/buffer-publish/blob/master/packages/server/index.js#L167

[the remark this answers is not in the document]
djfarrelly: Oh - it's already in master! Sorry I didn't catch this Amy Lee :)

packages/composer/composer/utils/TrackingUtils.js
```diff
@@ -0,0 +1,39 @@
...
+      break;
+  }
+  return sourceName;
+};
+
+const getSegmentMetadata = ({ post = {}, profile = {}, formattedData = {}, composerSource }) => ({
+  channelId: post.profile_id,
+  channelNetwork: post.profile_service,
+  channelType: profile.serviceType,
+  client: post.client ? post.client.website : null,
```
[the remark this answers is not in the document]
albennett: Hey Dan, great questions! From the tracking google doc, the client definition is "this accounts for 3rd party apps posting to Buffer".  

The post client is an object we get back with the response update. It’s an obj with url, name, id, website, description properties. I believe most of the logic is coming from the `client_model` Locally, the name property is printing as "Buffer Publish Dev”. Do you think returning the client name makes sense for this or are you thinking of something else?

packages/composer/composer/utils/TrackingUtils.js
```diff
@@ -0,0 +1,39 @@
...
+  composerSource,
+  hasFirstComment: !!post.comment_text,
+  hasLocation: !!post.service_geolocation_id,
+  hasShopGridLink: !!formattedData.link,
+  isDraft: !!post.commentText,
```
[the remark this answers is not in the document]
albennett: ah, no that is not correct for `isDraft`. I missed this 🙃, thanks for catching it.

packages/composer/composer/utils/TrackingUtils.js
```diff
@@ -0,0 +1,39 @@
...
+  mediaType: post.type,
+  postId: post.id,
+  product: 'publish',
+  shareDate: post.due_at,
```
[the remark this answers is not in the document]
albennett: We are getting back a unix timestamp for `due_at`. For example: `1564005125`

With drafts, it will return null for `due_at`. Great question!

packages/globalAccount/middleware.js
albennett: Hmm with Publish, a user can technically be associated with more than one organizationId if they're a team member. We currently have orgIds on the profile level, but not the user level.

albennett: We could possibly send an array instead of a string and include all orgIds associated with a publish user?

packages/analytics-middleware/middleware.js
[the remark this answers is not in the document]
albennett: No worries at all! 😀

packages/composer/composer/utils/TrackingUtils.js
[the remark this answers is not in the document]
djfarrelly: Yeah, returning the client name sounds like a good choice to me @albennett :) I think this is one of those properties that we're not totally sure how we'll need to use it yet, but the string client name seems like it will be the easiest to use in Mixpanel for now!

packages/analytics-middleware/actions.js
```diff
@@ -0,0 +1,35 @@
+import keyWrapper from '@bufferapp/keywrapper';
+
+export const actionTypes = keyWrapper('ANALYTICS', {
+  INIT: 'INIT',
+  TRACK_EVENT: 'TRACK_EVENT',
+  /* We need to include PUBLISH in the name since we are including analyze packages
+  that have a PAGE_CHANGE actionType. This causes the action to be called from the analyze code */ 
+  PUBLISH_PAGE_CHANGE: 'PUBLISH_PAGE_CHANGE',
```
karelm: @albennett this is fine having the `PUBLISH_` prefix, it just looks like a sign we might just need this as a global middleware in future I guess, or that we would benefit from namespacing our reducers, middleware, etc...
Does that same situation not apply to the INIT and TRACK_EVENT actions types aswell?

karelm: If so, would it not be simpler to change the keyWrapper namespace from ANALYTICS to PUBLISH_ANALYTICS and you wouldn't need to change the PUBLISH_PAGE_CHANGE name? just some food for thought

albennett: Great point. The TRACK_EVENT and INIT are not getting triggered. I think it has to do with how we explicitly have to trigger the profile selector logic for analyze in the profileSidebar middleware. The pageChange logic is happening in this action in analyze. 
` dispatch({
   type: 'PROFILE_SELECTOR__SELECT_PROFILE',
    profile: formatAnalyticsProfileObj(selectedProfile),
});`
I agree, much simpler & safer changing the keyWrapper name.

PR #849, opened 2019-07-26
Redirect user to login if missing valid access token

albennett:
## Description
Redirect user to login if missing valid access token

## Checklist

#### Staging Deployment

### DIFF
--- a/helm/buffer-publish.yaml
+++ b/helm/buffer-publish.yaml
@@ -59,6 +59,16 @@ env:
     value: https://k8s-api-php71.buffer.com
   - name: ANALYZE_API_ADDR
     value: https://analyze-api.buffer.com
+  - name: CLIENT_ID
+    valueFrom:
+      secretKeyRef:
+        key: client-id
+        name: buffer-publish-keys
+  - name: CLIENT_SECRET
+    valueFrom:
+      secretKeyRef:
+        key: client-secret
+        name: buffer-publish-keys
   - name: BUGSNAG_KEY
     valueFrom:
       secretKeyRef:
--- a/packages/server/middlewares/verifyAccessToken.js
+++ b/packages/server/middlewares/verifyAccessToken.js
@@ -23,8 +23,8 @@ function isAccessTokenValid(req) {
       client_secret: process.env.CLIENT_SECRET,
     },
   })
-    .then(r => r)
-    .catch(r => r);
+    .then(r => true)
+    .catch(r => false);
 }
 
 module.exports = async (req, res, next) => {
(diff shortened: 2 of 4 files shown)


helm/buffer-publish.yaml
```diff
@@ -59,6 +59,16 @@ env:
     value: https://k8s-api-php71.buffer.com
   - name: ANALYZE_API_ADDR
     value: https://analyze-api.buffer.com
+  - name: CLIENT_ID
+    valueFrom:
+      secretKeyRef:
+        key: client-id
+        name: buffer-publish-keys
```
karelm: is this in the secrets collection?

packages/server/middlewares/verifyAccessToken.js
```diff
@@ -0,0 +1,41 @@
+const rp = require('request-promise');
+
+function redirect (res, url) {
+  res.redirect(307, url);
+}
+
+function isRequestingApp(req) {
+  return req.method === 'GET' && req.accepts('text/html');
```
djfarrelly: Ah to make sure it's not a fetch/xhr right I'm guessing?

packages/server/middlewares/verifyAccessToken.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+function isAccessTokenValid(req) {
+  if (!req.session.publish.accessToken) {
+    return true;
```
djfarrelly: Should this be false @albennett if the token isn't there?

packages/server/middlewares/verifyAccessToken.js
federicoweber: That's right @djfarrelly, this basically constrains it to the first app load, as it's quite an expensive check to perform for all requests ;)

packages/server/middlewares/verifyAccessToken.js
federicoweber: Hey @djfarrelly, this is set to `true`, to limit it to only invalid access tokens. In this way, if we don't have one the middleware will just invoke `next`, and let the rest of the app logic kick in.

packages/server/middlewares/verifyAccessToken.js
```diff
@@ -0,0 +1,41 @@
...
+    .catch(r => false);
+}
+
+module.exports = async (req, res, next) => {
+  // As this is a quite expensive check to perform, and we want the redirect to happen
+  // before the app is rendered, we are processing it only on the first request.
+  if (isRequestingApp(req)) {
+    const isValid = await isAccessTokenValid(req);
+    if (!isValid) {
+      return redirect(res, 'https://login.buffer.com/login?redirect=https://publish.buffer.com?skipImpersonation=true');
```
federicoweber: Hey @albennett, we need this cause there could be some accounts for which the login service will erroneously think they are in impersonation mode, and won't re-issue the token. That was the problem I war experiencing Friday with the account redirect 😉.

albennett: Thanks for adding this @federicoweber! Glad you figured out the account redirect issue 🙌

helm/buffer-publish.yaml
albennett: Hey @karelm great question. We checked that this was present in the secrets collection :)

PR #859, opened 2019-07-31
Use user.json endpoint and handle non-production env

philippemiguet:
I will be deprecating the convert_access_token.json endpoint b/c we don't really need it anymore in the login service. It is too complex for what we really use it for, so I will be replacing it with a smaller endpoint.

So, let's user `user.json` here, it will serve the same purpose :) 

Buffer web PR: https://github.com/bufferapp/buffer-web/pull/16386

### DIFF
--- a/packages/server/middlewares/verifyAccessToken.js
+++ b/packages/server/middlewares/verifyAccessToken.js
@@ -20,13 +20,14 @@ function isAccessTokenValid(req) {
   }
 
   return rp({
-    uri: `${process.env.API_ADDR}/1/user/convert_access_token.json`,
-    method: 'POST',
+    uri: `${process.env.API_ADDR}/1/user.json`,
+    method: 'GET',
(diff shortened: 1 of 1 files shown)


packages/server/middlewares/verifyAccessToken.js
```diff
@@ -20,13 +20,14 @@ function isAccessTokenValid(req) {
...
-    method: 'POST',
+    uri: `${process.env.API_ADDR}/1/user.json`,
+    method: 'GET',
     strictSSL: !(process.env.NODE_ENV === 'development' || process.env.NODE_ENV === 'test'),
     qs: {
-      access_token: req.session.publish.accessToken,
       client_id: process.env.CLIENT_ID,
       client_secret: process.env.CLIENT_SECRET,
+      access_token: req.session.publish.accessToken,
+      simplified_response: true,
```
karelm: 👍

philippemiguet:
Thanks Karel! 🤗

msanroman:
Great stuff!

PR #892, opened 2019-08-20
CORE-768 stripe SCA required changes to RPC

esclapes:
This is part of the PSD2 requirements for Strong Customer Authentication via Stripe payments. **It requires front end changes before it can be merged**

## Description

Changes to the RPC endpoints are included in this PR:

* new PRC method `createSetupIntent` to be called before the user can add/update the credit card
* update `switchPlan` RPC endpoint to use a payment method id instead of the card token used before

## Context & Notes

* JIRA: https://buffer.atlassian.net/browse/CORE-768
* General paper with context is [here](https://paper.dropbox.com/doc/CORE-642-PSD2-Technical-Breakdown--AjMKrvwtz3r4dRmB658IFsV9Ag-jUXdDkVIhFJRWPOALLAut)

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/server/rpc/createSetupIntent/index.js
+++ b/packages/server/rpc/createSetupIntent/index.js
@@ -6,7 +6,7 @@ const strictSSL = !(process.env.NODE_ENV === 'development');
 module.exports = method(
   'createSetupIntent',
   'create a setup intent in Stripe',
-  ({ session }) => rp({
+  (args, { session }) => rp({
     uri: `${process.env.API_ADDR}/1/billing/create-setup-intent.json`,
     method: 'GET',
     strictSSL,
--- a/packages/server/rpc/createSetupIntent/index.js
+++ b/packages/server/rpc/createSetupIntent/index.js
@@ -8,10 +8,12 @@ module.exports = method(
   'create a setup intent in Stripe',
   ({ session }) => rp({
     uri: `${process.env.API_ADDR}/1/billing/create-setup-intent.json`,
-    method: 'POST',
+    method: 'GET',
     strictSSL,
     qs: {
       access_token: session.publish.accessToken,
+      client_id: process.env.CLIENT_ID,
+      client_secret: process.env.CLIENT_SECRET,
     },
   })
     .then(result => JSON.parse(result))
--- a/packages/server/rpc/createSetupIntent/index.js
+++ b/packages/server/rpc/createSetupIntent/index.js
@@ -8,7 +8,7 @@ module.exports = method(
   'create a setup intent in Stripe',
   (args, { session }) => rp({
     uri: `${process.env.API_ADDR}/1/billing/create-setup-intent.json`,
-    method: 'GET',
+    method: 'POST',
     strictSSL,
     qs: {
       access_token: session.publish.accessToken,
--- a/packages/server/rpc/createSetupIntent/index.js
+++ b/packages/server/rpc/createSetupIntent/index.js
@@ -6,22 +6,19 @@ const strictSSL = !(process.env.NODE_ENV === 'development');
 module.exports = method(
   'createSetupIntent',
   'create a setup intent in Stripe',
-  ({ session }) => {
-    console.log('ARGS', arguments)
-    return rp({
-      uri: `${process.env.API_ADDR}/1/billing/create-setup-intent.json`,
-      method: 'POST',
-      strictSSL,
-      qs: {
-        access_token: session.publish.accessToken,
-      },
-    })
-      .then(result => JSON.parse(result))
-      .catch((err) => {
-        if (err.error) {
-          const { error } = JSON.parse(err.error);
-          throw createError({ message: error });
-        }
-      })
-    }
+  ({ session }) => rp({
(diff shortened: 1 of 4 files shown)


esclapes:
@oscoda pinging you on this PR as your changes to the front end will require the new endpoints and changes in this PR

cc @chayjay as you have recently being updating the `switchPlan` modal.

packages/server/rpc/createSetupIntent/index.js
```diff
@@ -0,0 +1,27 @@
+const { method, createError } = require('@bufferapp/buffer-rpc');
+const rp = require('request-promise');
+
+const strictSSL = !(process.env.NODE_ENV === 'development');
+
+module.exports = method(
+  'createSetupIntent',
+  'create a setup intent in Stripe',
+  ({ session }) => {
+    console.log('ARGS', arguments)
```
josem: Just a tiny thing but do we need this log in production?

packages/server/rpc/switchPlan/index.js
```diff
@@ -27,7 +27,7 @@ const getCtaFromSource = source =>
 module.exports = method(
   'switchPlan',
   'switch user plan',
-  async ({ cycle, token, source, plan }, { session }) => {
+  async ({ cycle, paymentMethodId, source, plan }, { session }) => {
```
josem: This change requires another change [here ](https://github.com/bufferapp/buffer-publish/blob/9b2a8151a05802cdbec55e67fb9b9f3fe74312a8/packages/stripe/middleware.js#L38-L46) I think since we're passing the token and now it seems we need to pass the paymentMethodId. I wonder howe we could get that from that middleware. Happy to jump on a call to solve that.

packages/server/rpc/createSetupIntent/index.js
esclapes: Not at all! :sweat_smile: 

That should not be there, indeed!

packages/server/rpc/switchPlan/index.js
esclapes: Hey @josem, you are spot on! This branch won't be merged until @oscoda get's the front end and middleware part ready. We will also change the interaction with Stripe in the form too, so this is just the ground work.

josem: Ah sorry! I missed that! 

In that case all good! :)

packages/server/rpc/createSetupIntent/index.js
```diff
@@ -0,0 +1,24 @@
...
+
+module.exports = method(
+  'createSetupIntent',
+  'create a setup intent in Stripe',
+  ({ session }) => rp({
+    uri: `${process.env.API_ADDR}/1/billing/create-setup-intent.json`,
+    method: 'POST',
+    strictSSL,
+    qs: {
+      access_token: session.publish.accessToken,
```
philippemiguet: @esclapes, I don't think we use the `access_token` though we require the client id and secret: https://github.com/bufferapp/buffer-web/blob/fcb0eafea3820302178c42ef7042d62d20950a83/api/controllers/billing_controller.php#L287

Have you tested this RPC endpoint is working properly?

esclapes: Good catch @philippemiguet !

It seems we do not heck credentials when in local development  :man_shrugging: 
 so the endpoint was working fine on my tests. I will update this and use postman to call the dev branch environment to double-check

https://github.com/bufferapp/buffer-web/blob/6e60e9874f0ed2b5d6d7f206bfb463bb83fdae83/api/core/MY_Controller.php#L551

packages/server/rpc/createSetupIntent/index.js
```diff
@@ -0,0 +1,26 @@
+const { method, createError } = require('@bufferapp/buffer-rpc');
+const rp = require('request-promise');
+
+const strictSSL = !(process.env.NODE_ENV === 'development');
+
+module.exports = method(
+  'createSetupIntent',
+  'create a setup intent in Stripe',
+  ({ session }) => rp({
```
philippemiguet: @esclapes, looking at other RPC endpoint, it seems that `req` is the second argument, so `session` may be undefined since you do not declare a first argument. 

Dace is having an error with his PR and I think this is coming from here:

esclapes: Hey @philippemiguet, you are right.

I had to dig into this too while I was implementing this endpoint because I was having the opposite problem when testing with Postman. 

The number of arguments is not always the same because [`req` becomes the first argument when we pass no `args`](https://github.com/bufferapp/buffer-rpc/blob/master/src/rpc.js#L26) (or an empty array) on the rpc call.

However, I see that `async-data-fetch` [sends and empty object](https://github.com/bufferapp/buffer-js-async-data-fetch/blob/master/src/middleware.js#L15) instead when we call it with no `args`, so indeed for Dace it won't be there.

@oscoda I am sorry about this, I was not aware of the default `args` value on `async-data-fetch`. I will add a commit to fix it.

esclapes:
@philippemiguet I belive this has been merged into another branch, right? Should we close it?

PR #893, opened 2019-08-20
[PUB-1701] Draft tabs content loading

MayaUribe:
## Description

Fixing the way we load drafts, awaiting approval and pending approval tabs.

## Context & Notes

We are currently experiencing a bug in which users can't load more drafts when reaching to the bottom of the page, the reason was the way we were fetching and displaying the drafts in our UI.

The solution was to load the drafts data on tab change, instead on profile change only, and also "reseting" the drafts state when this change takes place, so that we can see the actual content for the selected tab.

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/drafts/middleware.js
+++ b/packages/drafts/middleware.js
@@ -1,5 +1,6 @@
-import { actionTypes as profileActionTypes } from '@bufferapp/publish-profile-sidebar';
+import { actionTypes as tabsActionTypes } from '@bufferapp/publish-tabs';
 import { actions as analyticsActions } from '@bufferapp/publish-analytics-middleware';
+import { getProfilePageParams } from '@bufferapp/publish-routes';
 
 import {
   actions as dataFetchActions,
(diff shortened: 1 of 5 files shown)


packages/drafts/middleware.js
```diff
@@ -20,16 +21,24 @@ const getTrackingData = ({ post = {}, channel = {} }) => ({
 export default ({ dispatch, getState }) => next => (action) => {
   next(action);
   const state = getState();
+  const path = getState().router.location.pathname;
+  const { tabId } = getProfilePageParams({ path }) || {};
+  const needsApproval = tabId === 'awaitingApproval' || tabId === 'pendingApproval';
+  const isDraft = ['awaitingApproval', 'pendingApproval', 'drafts'].indexOf(action.tabId) !== -1;
 
   switch (action.type) {
-    case profileActionTypes.SELECT_PROFILE:
```
karelm: Hey @MayaUribe, do we not need to do this when selecting between different profiles too?

MayaUribe: Not really @karelm, because we are currently calling the `selectTab` action in `onProfileClick` [[ref](https://github.com/bufferapp/buffer-publish/blob/e12ef25fad52894e05870733c2a36d0f420403f3/packages/profile-sidebar/index.js#L32)], so the drafts are also updated on changing either profile or tab.

I hope that explanation makes sense 😄

karelm: Cool, then all looks good to me 🙂

PR #900, opened 2019-08-21
Extracting UploadZone component and isolating logic into new package

karelm:
## Description

Extracting the UploadZone component into its own function, and refactoring callbacks to be passed into function.

## Context & Notes

This is a WIP with the functions extracted and some of the callbacks refactored
Still need to isolate the callbacks properly and pass them in, plus figure out how to run the component in isolation and test it better.

We have a dependency on `@bufferapp/publish-composer` for the string utils, so still need to refactor those out

## Checklist

#### Staging Deployment


packages/UploadZone/components/UploadZone/index.jsx
```diff
@@ -0,0 +1,213 @@
...
+      if (!uploadFormatsConfig.has(fileFormat)) {
+        invalidFormatFilesCount++;
+        return false;
+      }
+
+      const uploadFormatConfig = uploadFormatsConfig.get(fileFormat);
+      if (file.size > uploadFormatConfig.maxSize) {
+        const formattedMaxSize = getHumanReadableSize(uploadFormatConfig.maxSize);
+        queueError({
+          message: `We can't upload "${preventXss(file.name)}" because it's too large: we can
```
albennett: I wonder if we should add these error messages to the translations package.

packages/UploadZone/components/UploadZone/uploadZone.css
```diff
@@ -0,0 +1,22 @@
+.transparentClickZone {
```
albennett: it would be neat to refactor these with inline styles.

packages/composer/composer/action-creators/ComposerActionCreators.js
```diff
@@ -504,103 +505,164 @@ const ComposerActionCreators = {
         ModalActionCreators.closeModal('overlay');
       });
   },
-
-  uploadDraftFile: (id, file, uploadType) => {
+  notifiers: {
```
albennett: 🙌

albennett:
Great job with pulling the UploadZone logic into its own package. 🙌 So neat to see how these changes are improving the composer logic, too! Def. worth it. I left a comment or two, though I know this is WIP. 😄 

With the UploadZone, will the UX be the box with the progress indicator similar to the current composer? We may want to start thinking through on how we're going to change the UX to a button.

packages/UploadZone/components/UploadZone/index.jsx
karelm: we can add dynamic messages to the translations package 😄 will look into addressing this one 👍

packages/UploadZone/components/UploadZone/uploadZone.css
karelm: Yes, it was something I was definitely looking into doing based on Ana's thread (https://threads.com/34368866906) 👍
I thought I'd get things extracted first + cleaned up then once I was happy I was going to refactor them to be styled components 😄

packages/composer/composer/action-creators/ComposerActionCreators.js
karelm: I was looking through this all, and realised how similar most of these calls to handleViewAction were too, they basically all expect an id, uploaderInstance, a url, height, and width, with a few others expecting a couple extra / additional properties, so I think theres potential to reduce this down to a more pure functional approach and just use a single uploadedCallback notifier and a trackAction notifier, but will give this some more thought later 👍

packages/composer/composer/components/LinkAttachmentThumbnailEditor.jsx
```diff
@@ -81,6 +84,14 @@ class LinkAttachmentThumbnailEditor extends React.Component {
           />}
 
         <UploadZone
+          mixedMediaUnsupportedCallback={FileUploader.throwMixedMediaTypesError}
```
albennett: Great job on isolating this 🙌

PR #903, opened 2019-08-23
[PUB-1905] Adding single slot per day and load more queue.

MayaUribe:
## Description

Adding single slot per day and load more queue.

## Context & Notes

Adding a slot per day in the Stories tab queue, so it's available at all times to create new content.

## Screenshots
Image 2019-08-23 at 12 22 16 PM

## Checklist

#### Staging Deployment


packages/shared-components/QueueItems/index.jsx
```diff
@@ -90,7 +117,7 @@ const renderPost = ({
 
   const defaultStyle = {
     default: {
-      marginBottom: '2rem',
+      marginBottom: !isStory ? '2rem' : '8px',
```
karelm: Would it make more sense to not have this as !isStory to make the logic more straight forward to understand? So just have `isStory ? '8px' : '2rem'` instead? Just a small suggestion 🙂

packages/shared-components/QueueItems/index.jsx
```diff
@@ -196,26 +223,62 @@ const renderDraft = ({
...
+            <span style={headerTextDateStyle}>{date}</span>
+          </React.Fragment>
+        ) : <span style={headerTextDayOfWeekStyle}>{text}</span>
+      }
+    </div>
   </div>
 );
 
+const renderSlot = ({ id, slot, profileService }, onEmptySlotClick) => (
+  <PostEmptySlot
```
karelm: It might be worth changing the name of this component from PostEmptySlot to maybe QueueEmptySlot?

packages/shared-components/QueueItems/index.jsx
```diff
@@ -196,26 +223,62 @@ const renderDraft = ({
...
 
 const QueueItems = (props) => {
-  const { items, type, ...propsForPosts } = props;
+  const { items, type, onEmptySlotClick, ...propsForPosts } = props;
   const itemList = items.map((item, index) => {
     const { queueItemType, ...rest } = item;
     if (queueItemType === 'post') {
-      return type === 'drafts' ? renderDraft({ draft: rest, ...propsForPosts }) : renderPost({ post: rest, index, ...propsForPosts });
+      switch (type) {
+        case 'drafts':
```
karelm: I'm not sure there are drafts for the stories queue, but I could be wrong 😅

packages/shared-components/QueueItems/index.jsx
MayaUribe: I literally just made that change 😄

packages/shared-components/QueueItems/index.jsx
MayaUribe: I think it makes sense to change it @karelm but for the time being I think I'm just going to leave it as it is, since I'm reusing the one we already have, so I'd have to hake several changes in several components, not sure if that makes sense to you 😄

packages/shared-components/QueueItems/index.jsx
MayaUribe: @karelm no there's aren't :) that QueueItems component is loading posts data for all tabs, except the queue, which is why I added the `stories` check.

PR #905, opened 2019-08-27
[CORE-770] Update credit card component to use Stripe's Setup Intent and SCA

oscoda:
## Description

Adds a new Stripe Elements payment flow with SCA.

### Todo:

~I have a pending PR for `ui` that will allow the hiding of the footer button supplied by the `Modal` component. That should be approved and merged in before this PR as this will rely on those changes.~

## Context & Notes

[JIRA CORE-770](https://buffer.atlassian.net/browse/CORE-770)
More details on how to test it in this Thread: https://threads.com/34370297404

Screen Recording 2019-09-03 at 10 34 PM

## Checklist

#### Staging Deployment

[CORE-770]: https://buffer.atlassian.net/browse/CORE-770


packages/credit-card-form/components/CreditCardForm/index.jsx
```diff
@@ -0,0 +1,28 @@
...
+  };
+
+  componentDidMount() {
+    const { getSetupIntent } = this.props;
+    getSetupIntent();
+  }
+
+  render() {
+    return (
+      <StripeProvider apiKey="pk_test_TYooMQauvdEDq54NiTphI7jx">
```
philippemiguet: @oscoda, I'm sure that's the right testing key. We need to use the one that is probably passed from the server to the frontend :)

packages/server/index.js
```diff
@@ -125,11 +125,7 @@ const showModalScript = (key, val) => {
   `;
 };
 
-const stripeScript = `<script src="https://js.stripe.com/v2/"></script>
-<script type="text/javascript">
-    Stripe.setPublishableKey('${stripePublishableKey}');
```
philippemiguet: We need a way to pass this key to the frontend to be used with the StripeProvider

packages/stripe/middleware.js
```diff
@@ -35,18 +35,17 @@ export default ({ dispatch, getState }) => next => (action) => {
...
-            },
-          }));
+          //  Upgrade plan
         }
       });
       break;
+    case actionTypes.SETUP_INTENT_REQUEST:
+      dispatch(
+        asyncDataFetchActions.fetch({
+          name: 'createSetupIntent',
```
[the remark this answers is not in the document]
philippemiguet: Hey @oscoda I left a comment in Slack, my intuition is that the issue is coming from Edu's PR here: https://github.com/bufferapp/buffer-publish/pull/892#pullrequestreview-280120633. I asked him to merge his buffer-web PR and fix his buffer-publish PR, it should unblock you :)

packages/stripe/middleware.js
```diff
@@ -26,7 +26,7 @@ export default ({ dispatch, getState }) => next => (action) => {
   const errorMessages = getState().i18n.translations.stripe;
   switch (action.type) {
     case actionTypes.CREDIT_CARD_VALIDATING:
-      Stripe.createToken(action.card, (status, response) => {
+      Stripe.handleCardSetup(action.card, (status, response) => {
```
philippemiguet: @oscoda, I know this is still a WIP, but since I'm in there, here you will need to pass the Setup Intent Client Secret instead of the card :)

oscoda:
Hi @philippemiguet @esclapes, I added some updates to this PR. I think it is close, so would like to get some feedback when you feel you are available to give it a look. There are a few small things that still need cleaning up (updating tests, properly displaying error and success states, slight styles, etc.), but happy to get to those tomorrow. 

A couple of things to note:

1.  The modal from the UI library comes with a default button that you can pass a custom callback. This presents an issue because the Stripe Elements credit card form is wrapped in a special Stripe context (called `<Elements>`) and because the button lives outside of that context, it would require a lot of hacking to try and submit a child form from a parent's submit button, prevent the default behavior, and also get access to the context of where that event is fired. This implementation adds a button within the Stripe Elements credit card form so that it's all kept together. Currently there is no way to **not** render this default modal button. I would like to ask the BDS team about updating this so we can maybe pass a `hideButton` prop to disable rendering it. 

Screen Shot 2019-08-30 at 12 42 31 AM

2. I seem to be getting an error from the RPC endpoint that handles `switchPlan`. I can't quite figure out what the issue is, but I have a feeling it's something really simple that I'm not doing correctly. It should hit that endpoint after a card has successfully been validated / authenticated:

Screen Shot 2019-08-30 at 12 57 54 AM

Screen Shot 2019-08-30 at 12 33 44 AM

As always - thank you!  🙏

packages/credit-card-form/components/CreditCardForm/styles.js
```diff
@@ -0,0 +1,17 @@
+import styled from 'styled-components';
+
+export const InputWrapper = styled.div`
+  border: 1px solid #E0E0E0;
```
anafilipadealmeida: we usually `import { grayLight } from '@bufferapp/ui/style/colors';` and use the color variable 🙂

packages/credit-card-form/components/CreditCardForm/styles.js
```diff
@@ -0,0 +1,17 @@
...
+  border-radius: 4px;
```
anafilipadealmeida: The same for border-radius `import { borderRadius } from '@bufferapp/ui/style/borders';`

packages/credit-card-form/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "@bufferapp/credit-card-form",
+  "version": "1.0.0",
```
anafilipadealmeida: Our own packages usually start at "2.0.0" and are named with "publish" to be easier to identify: "@bufferapp/publish-credit-card-form"

packages/credit-card-form/components/CreditCardForm/form.jsx
```diff
@@ -0,0 +1,63 @@
+import React, { Component } from "react";
+import { CardElement, injectStripe } from "react-stripe-elements";
+import { Button } from "@bufferapp/ui";
+import { ButtonWrapper, InputWrapper } from "./styles";
+
+class CreditCardForm extends Component {
+  constructor () {
+    super();
+
+    this.handleSubmit = this.handleSubmit.bind(this);
```
karelm: handleSubmit is already bound on line 13, so the entire constructor can be removed really

packages/credit-card-form/components/CreditCardForm/index.jsx
```diff
@@ -0,0 +1,40 @@
...
+    createSetupIntentRequest();
+  }
+
+  render() {
+    const { stripePublishableKey, stripe } = this.props;
+
+    return (
+      <StripeProvider stripe={stripe}>
+        <Elements>
+          <CreditCardForm {...this.props} />
```
karelm: it would be nicer to be explicit about props here instead of spreading all props... if someone adds a className or something similar to the parent, theres a chance that gets send through and changes the styling associated, aswell as other interesting potential side-effects of doing this.

If you can just create explicit props here, it would make it clearer + more simple to understand. Even if this is just dont using spread props, e.g.

```
const { creditCardFormProp1, creditCardFromProp2, creditCardFormProp3 } = this.props;

...

<CreditCardForm {...{ creditCardFormProp1, creditCardFromProp2, creditCardFormProp3 }} />
```

let me know how this feels to you

packages/credit-card-form/components/CreditCardForm/index.jsx
```diff
@@ -0,0 +1,40 @@
...
+  componentDidMount() {
+    const { createSetupIntentRequest } = this.props;
+    createSetupIntentRequest();
+  }
+
+  render() {
+    const { stripePublishableKey, stripe } = this.props;
+
+    return (
+      <StripeProvider stripe={stripe}>
```
karelm: I had to change this prop from apiKey to be stripe instead + pass in a null value for the sake of testing.
It would be really great if we could lean into the async approach here a bit more closely / some of the advanced techniques discussed https://github.com/stripe/react-stripe-elements#advanced-integrations

If we implement it that way, we can then make this explicitly testable, and also handle things if stripe isn't loaded for some reason. Also, switching stripe to load async allows us to not block loading for the rest of the site. Stripe is only really needed when we are doing something related to Monetary functions on the site, so it would be really great to not even have to load it until we reach an appropriate code-path really.

packages/credit-card-form/reducer.js
```diff
@@ -0,0 +1,12 @@
+export default (state = { stripePublishableKey: null, stripe: null }, action) => {
+  switch (action.type) {
+    case 'APP_INIT':
+      return {
+        ...state,
+        stripe: window._stripe,
```
karelm: this has been introduced to make the component testable, as this can then be injected separately into the component

packages/server/index.js
```diff
@@ -125,10 +125,13 @@ const showModalScript = (key, val) => {
...
-</script>
+const stripeScript = `
+  <script src="https://js.stripe.com/v3/"></script>
+  <script type="text/javascript">
+    const stripe = Stripe('${stripePublishableKey}');
+    window._stripe = stripe;
```
karelm: this line has been introduced to make the component testable, as this can then be injected separately into the component later. We can probably do something a little better than this using the advanced techniques suggestions for stripe (https://github.com/stripe/react-stripe-elements#advanced-integrations)

packages/server/rpc/createSetupIntent/index.js
```diff
@@ -0,0 +1,26 @@
+const { method, createError } = require('@bufferapp/buffer-rpc');
+const rp = require('request-promise');
+
+const strictSSL = !(process.env.NODE_ENV === 'development');
+
+module.exports = method(
+  'createSetupIntent',
```
karelm: createSetupIntent doesn't really clearly explain what this is doing... could be rename it to something else? perhaps setupPaymentProviderIntent or something more clear?

packages/switch-plan-modal/components/SwitchPlanModal/index.jsx
```diff
@@ -218,68 +129,14 @@ class SwitchPlanModal extends React.Component {
...
-                />
-              </div>
-            </div>
+            <StripeCreditCardForm
+              buttonLabel={
+                validating ? translations.validating : getButtonText({ plan, translations })
+              }
+              closeButtonLabel={translations.close}
+              closeAction={this.onSecondaryAction}
+              {...this.props}
```
karelm: I think this is another one of those ones where it would be nicer to be explicit about the props being handed down, if buttonLabel/closeButtonLabel/closeAction exist in the props, they will be overridden by the `{...this.props}` call, which could cause some strange side-effects if someone doesn't notice.

If you could explicitly declare the props for the component and pass them in I'd really appreciate it 👍

PR #913, opened 2019-09-02
Normalize Posts/ Drafts/ Stories Footer

anafilipadealmeida:
## Description
Create a fully Presentational Component to render the Footer, no matter what type of card (either Story, Post or Draft)

## Context & Notes

- With the creation of Stories, we would be adding the same logic for the third time. This presentational component allows us to have all the styles in one place (making the consistency easier to maintain) and it allows us to adapt the existing code without major changes.

- There's also no need for so much prop drilling: some props being passed just to show delete buttons are replaced by the use of useState (react hooks)

- Allows us to have the entire Footer with BDS

- This PR includes the adaptation of PostFooter. 

**Next Septs (different PR, after deploying this):** 
- Do the same with Drafts Footer
- Remove PostFooterButtons and PostFooterDelete from codebase
- Remove delete props (`onCancelConfirmClick` and `onDeleteClick`) from several Post and Drafts components

https://hi.buffer.com/2bfbaf079ea2

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/shared-components/CardFooter/index.jsx
+++ b/packages/shared-components/CardFooter/index.jsx
@@ -1,0 +1,218 @@
+import React, { Fragment, useState } from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { borderWidth } from '@bufferapp/components/style/border';
+import { offWhite } from '@bufferapp/components/style/color';
+import {
+  Link,
+} from '@bufferapp/components';
+
+import { Text, Button } from '@bufferapp/ui';
+import { grayLight, grayDarker, red } from '@bufferapp/ui/style/colors';
+import { fontWeightBold } from '@bufferapp/ui/style/fonts';
+
+const ALLOWED_COLORS = ['shuttleGray', 'red'];
+
+const igCommentIconWrapper = {
+  display: 'flex',
+  alignItems: 'center',
+  paddingLeft: '0.5rem',
+  marginLeft: '0.5rem',
+  borderLeft: '1px solid #cfd4d6',
+};
+const renderCommentIcon = () => (
+  <span style={igCommentIconWrapper} title="Post includes a comment">
+    <svg
+      width="16"
+      height="16"
+      viewBox="0 0 16 16"
+      fill="none"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        fillRule="evenodd"
+        clipRule="evenodd"
+        fill="#59626a"
+        d="M1 3a1 1 0 0 1 1-1h12a1 1 0 0 1 1 1v7a1 1 0 0 1-1 1H6c-.746 0-2.886 1.672-4.13 2.52-.457.31-.87.032-.87-.52V3z"
+      />
+    </svg>
+  </span>
+);
+
+const CardFooterWrapper = styled.div`
+  display: flex;
+  padding: 0.5rem 1rem;
+  opacity: ${({ isDragging }) => (isDragging ? 0 : 1)};
+`;
+
+const CardFooterWrapperWithBorder = styled(CardFooterWrapper)`
+  padding: 12px 16px;
+  background-color: ${offWhite};
+  borderTop: ${borderWidth} solid ${grayLight};
+`;
+
+const CardFooterContent = styled.div`
+  display: flex;
+  flex-grow: 1;
+  align-items: center;
+`;
+
+const VerticalDivider = styled.span`
+  margin: 0 20px 0 12px;
+  height: 32px;
+  border-left: 1px solid ${grayLight};
+  border-right: 1px solid ${grayLight};
+`;
+
+const IconWrapper = styled.span`
+  margin-right: 0.5rem;
+  display: flex;
+  align-items: center;
+`;
+
+const Message = styled(Text)`
+  font-size: 12px;
+`;
+
+const PreMessage = styled(Text)`
+  font-size: 12px;
+  font-weight: ${fontWeightBold};
+`;
+
+const EditButton = styled(Button)`
+  margin-right: 8px;
+`;
+
+const RequeueButton = styled(Button)`
+  margin-left: 8px;
+`;
+
+const ButtonWrapper = styled.span`
+  display: flex;
+`;
+
+const CardFooter = ({
+  hideButtons,
+  icon,
+  preMessage,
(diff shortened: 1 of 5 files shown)


packages/shared-components/CardFooter/index.jsx
```diff
@@ -0,0 +1,214 @@
...
+import MessageIcon from '@bufferapp/ui/Icon/Icons/Message';
+import { Text, Button } from '@bufferapp/ui';
+import {
+  grayLight, grayDarker, red, grayDark,
+} from '@bufferapp/ui/style/colors';
+import { fontWeightBold } from '@bufferapp/ui/style/fonts';
+
+const ALLOWED_COLORS = [grayDark, red];
+
+const IgCommentIconWrapper = styled.span`
```
karelm: is this explicitly an Instagram Comment Icon Wrapper, as I imagine it's just a comment icon wrapper. Would it better suit this to be renamed to remove the Instagram Reference? (just a small suggestion)

packages/shared-components/CardFooter/index.jsx
```diff
@@ -0,0 +1,214 @@
...
+                  &nbsp;
+                </PreMessage>
+              )}
+              <Message type="p" textColor={textColor}>{ message }</Message>
+            </Fragment>
+          )
+        }
+        {hasFirstComment
+          && (
+            renderCommentIcon()
```
karelm: Can this not just be a <CommentIcon /> instead of a renderCommentIcon()?

packages/shared-components/CardFooter/index.jsx
```diff
@@ -0,0 +1,214 @@
...
+  onSubmitClick,
+  submitLabel,
+  onRequeueClick,
+  hasFirstComment,
+  isPerformingAction,
+  actionMessage,
+  isDragging,
+  disableBorder,
+}) => {
+  const hideAllButtons = (hideButtons || isPerformingAction) || messageLink;
```
karelm: I'm not sure you need the brackets around this, but if i'm missing something do let me know 😄

packages/shared-components/CardFooter/index.jsx
```diff
@@ -0,0 +1,214 @@
...
+  const [isConfirmingDelete, setConfirmingDelete] = useState(false);
+  const textColor = ALLOWED_COLORS.includes(messageColor) ? messageColor : grayDarker;
+  const WrapperComponent = disableBorder ? CardFooterWrapper : CardFooterWrapperWithBorder;
```
karelm: what does disabledBorder mean here? just wondered if its contextually clear, or whether we can use a better named variable?

packages/shared-components/CardFooter/index.jsx
```diff
@@ -0,0 +1,214 @@
...
+        }
+      </CardFooterContent>
+      {!hideAllButtons
+        && (
+          <ButtonWrapper>
+            {onDeleteClick
+              && (
+                <Fragment>
+                  {!isConfirmingDelete
+                    ? <Button type="text" label="Delete" size="small" onClick={() => setConfirmingDelete(true)} />
```
karelm: would it be cleaner if this entire isConfirmingDelete section was extracted out into its own component at all?

packages/shared-components/PostFooter/index.jsx
```diff
@@ -1,145 +1,26 @@
...
 const renderIcon = (hasError, isSent, isCustomScheduled, isInstagramReminder, isPastReminder) => {
   if ((!hasError && !isCustomScheduled && !isInstagramReminder) || isPastReminder) return;
 
-  return (<div style={postActionDetailsIconStyle}>
-    {isInstagramReminder && !hasError ? <CircleInstReminderIcon color={'instagram'} /> : null}
-    {isCustomScheduled && !hasError && !isInstagramReminder ? <ClockIcon color={isSent ? 'shuttleGray' : 'outerSpace'} /> : null}
-  </div>);
+  return (
+    <div>
+      {isInstagramReminder && !hasError ? <CircleInstReminderIcon color="instagram" /> : null}
```
karelm: would be really nice to isolate this to a reminder component, that happens to know its an instagram reminder, vs another networks reminder.
e.g.
```
// with reminder set to { isReminder: true, type: 'instagram' }
reminder.isReminder ? <Reminder type={reminder.type} />
```

More of a future suggestion to remove explicit social networks from the main code paths, I think that would give us the best future reusability 👍

packages/shared-components/CardFooter/index.jsx
anafilipadealmeida: You're not missing anything 😅 I started by having a more complex condition with negation and it was need, with this simplification it's no longer needed so I'll remove it!

packages/shared-components/CardFooter/index.jsx
anafilipadealmeida: I can maybe call it hideTopBorder? 🤔

karelm: that might well make it a little bit clearer... but maybe focus on what triggers that condition? Is it a condition of the component... like, nested component vs not? or something that might make its usage clearer than the state of the border instead? Just thought I'd bring it up, but feel free to leave things alone :)

packages/shared-components/CardFooter/index.jsx
anafilipadealmeida: I would personally prefer to keep it here, as it is quite small and only used here. We have so much granularity that often it gets harder to track things

packages/shared-components/PostFooter/index.jsx
anafilipadealmeida: I like the idea of doing it in the future 👍

packages/shared-components/CardFooter/index.jsx
anafilipadealmeida: I prefer to leave it as hideTopBorder or disabledBorder, as it is exactly what it does. Since this is just the dumb component, it makes it reusable. Right now SentPosts will hide the border, so the parent (PostFooter) will determine it. If we decide we also want to add PastReminders to that, the parent will do. If we decide we want to hide the border for Drafts, then DraftsFooter can determine it as well, without the need to change this component or pass additional props.

karelm: sounds great 🙌

PR #927, opened 2019-09-11
Moving the buffermetrics middleware

karelm:
Moving the buffermetrics middleware to before the session auth occurs to hopefully prevent cookies from being set when using buffermetrics

This needs to be tested against live to see whether the issue persists after this change or not, it may need to be reverted if it doesn't work

### DIFF
--- a/packages/server/index.js
+++ b/packages/server/index.js
@@ -218,16 +218,7 @@ app.use('*', (req, res, next) => {
   next();
 });
 
-app.use(
-  setRequestSessionMiddleware({
-    production: isProduction,
-    sessionKeys: ['publish', 'global'],
-  }),
-);
-
 app.use(bodyParser.json());
-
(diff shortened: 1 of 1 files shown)


packages/server/index.js
```diff
@@ -218,24 +218,26 @@ app.use('*', (req, res, next) => {
   next();
 });
 
+app.use(bodyParser.json());
```
albennett: Hm, so the issue was with the order of the middleware. I ran into an issue related to order with validateSessionMiddleware. Order matters! 😄

karelm: I've moved the bodyParser, as the bufferMetricsMiddleware expects a parser json object to sent through data... I suppose the question is whether the session middleware expects the bodyParser to have run or not before getting to what it does 🤔 @philippemiguet / @esclapes do you know whether this will be a safe change to move the json bodyparser above the session middleware?

PR #928, opened 2019-09-11
Prevent Analyze 'profiles' api fetch from firing aswell as publish's

karelm:
## Description
Currently, profiles from api is getting fetched twice in new publish because a call gets triggered in the analyze profile-loader middleware. Removing the profile-loader middleware seems to fix it and analytics data seems to be coming in correctly. 

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/store/index.js
+++ b/packages/store/index.js
@@ -141,7 +141,8 @@ const configureStore = initialstate => {
         compareChartMiddleware,
         datePickerMiddleware,
         postsMiddleware,
-        profileLoaderMiddleware,
+        // Removing the Analyze profile loader middleware as it was causing a double load to occur
+        // profileLoaderMiddleware,
         profileSelectorMiddleware,
         summaryTableMiddleware,
         // These need to be the last middlewares in the chain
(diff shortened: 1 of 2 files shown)


packages/store/index.js
```diff
@@ -141,7 +141,8 @@ const configureStore = initialstate => {
         compareChartMiddleware,
         datePickerMiddleware,
         postsMiddleware,
-        profileLoaderMiddleware,
```
albennett: I believe we'll still need the `profileLoaderMiddleware` to get the analyze logic to load. Let's bring up this issue at our sync tomorrow and see if we can get the nomenclature refactor done before summit. Until then, we can comment it out locally when working on story tasks.

karelm: the only thing the profileLoaderMiddleware was doing was the 'profiles' api call

image

so I think it should be okay to remove

karelm: but if it does more in future, this could be quite a messy workaround

albennett: Ok, that makes sense to me. Thanks for the context 😄  As long as everything is loading correctly in the analytics overview tab, I'd say remove it and we can monitor if any reports come up.

albennett: I'll add Federico as a reviewer so he can have context incase changes are made to the profile-loader.

PR #933, opened 2019-09-12
Managing uploaded file state + maintaining clean cloned object

karelm:
## Description

Upload images and create a new story group card as a consequence

## Checklist

#### Staging Deployment


packages/story-group-composer/components/CarouselCards/index.jsx
```diff
@@ -186,22 +185,30 @@ class CarouselCards extends React.Component {
...
                       canHaveMediaAttachmentType: () => true,
                     }}
-                    uploadType={UploadTypes.LINK_THUMBNAIL}
+                    uploadType={UploadTypes.MEDIA}
                     multiple
                     disabled={false}
                   />
                 </div>
               )}
+              {!card.empty && card.progress !== null && card.uploading && (
```
MayaUribe: This was satisfying to test 😄

packages/story-group-composer/components/StoryGroupWrapper/index.jsx
```diff
@@ -38,14 +38,15 @@ const StoryGroupWrapper = ({
...
   onComposerClick,
+  onCreateNewStoryCard,
+  onUpdateStoryUploadProgress,
   onSetShowDatePicker,
   showDatePicker,
   userData,
   draft,
 }) => {
-  // hooks: https://reactjs.org/docs/hooks-state.html
+  const cards = editingStoryGroup ? editingStoryGroup.storyDetails.stories : draft.stories;
```
MayaUribe: I think we should try to consolidate this variable to be the same everywhere so it's consistent, we have `draft` and `editingStoryGroup`, they don't seem to be the same type of object although they are both referring to storyGroup.

packages/story-group-composer/index.js
```diff
@@ -45,6 +45,24 @@ export default connect(
     onComposerClick: (showDatePicker) => {
       if (showDatePicker) dispatch(actions.setShowDatePicker(false));
     },
+    onCreateNewStoryCard: ({ id, uploaderInstance, file }) => {
```
MayaUribe: So we are calling this specifically for file uploading, correct? and when we click on Schedule is when we do the StoryGroup create/update with this saved state, correct? :)

packages/story-group-composer/reducer.js
```diff
@@ -68,11 +113,43 @@ export default (state = initialState, action) => {
...
+    }
+    case actionTypes.CREATE_NEW_STORY_CARD: {
+      const { stories } = state.draft;
+      const { file } = action.args;
+      return {
+        ...state,
+        draft: {
+          ...state.draft,
+          stories: [...stories, {
+            ...newStory(),
```
MayaUribe: Oh I see, this is really really great! we should try to implement something like this for other reducers ✨

packages/story-group-composer/index.js
karelm: yup, thats the one 😄

packages/story-group-composer/reducer.js
karelm: I'd still love to have immutable data everywhere in the ideal world, but for now this works 💯

PR #975, opened 2019-09-24
Fix update story group endpoint with form

josem:
## Description
The update request needs to follow the same format that others requests with bodies.

## Checklist

#### Staging Deployment


josem:
Hey @MayaUribe! One small thing @karelm and I noticed in the onsite is that we need to send the params as "form" because they're url-encoded. If we don't, we might hit some limits in the URLs sometimes because they are sent as GET params in the URL. 

I think it'd be good at some point to audit our packages to see if we're doing the same in other places, but no rush at all! We can chat more next week.

Thanks so much! 🙌

MayaUribe:
@josem @karelm sorry for the late message on this one! good catch Karel! thank you for the adjustment here :D

PR #983, opened 2019-09-25
[PUB-2016] Add STORY_SENT action from Pusher event

anafilipadealmeida:
## Description
Dispatch `STORY_SENT` actionType when we receive a `sent_story_group` event from Pusher, that will update the stories queue.

## Context & Notes
[PUB-2016](https://buffer.atlassian.net/browse/PUB-2016)

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/pusher-sync/middleware.js
+++ b/packages/pusher-sync/middleware.js
@@ -109,15 +109,17 @@ export default ({ dispatch }) => {
     if (action.type === tabsActionTypes.SELECT_TAB) {
       const { profileId, tabId } = action;
       if (profileId) {
-        if (!channelsByProfileId[profileId]) {
-          channelsByProfileId[profileId] = {
-            updates: pusher.subscribe(`private-updates-${profileId}`),
-            storyGroups: tabId === 'stories' ? pusher.subscribe(`private-story-groups-${profileId}`) : null,
-          };
-          bindProfileUpdateEvents(channelsByProfileId[profileId].updates, profileId, dispatch);
-          if (channelsByProfileId[profileId].storyGroups) {
-            bindProfileStoryGroupEvents(channelsByProfileId[profileId].storyGroups, profileId, dispatch);
-          }
+        // If the profile is not subscribed to any channels, subscribes to private-updates channel:
+        const newProfileChannels = channelsByProfileId[profileId] || { updates: pusher.subscribe(`private-updates-${profileId}`) };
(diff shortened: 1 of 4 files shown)


anafilipadealmeida:
@josem and @karelm Any idea how can we test this will work right now? @josem, I'm assuming we're receiving `data.story_group` based on this: https://github.com/bufferapp/buffer-web/blob/master/shared/services/StoryGroupBroadcaster.php#L194-L198

karelm:
> @josem and @karelm Any idea how can we test this will work right now? @josem, I'm assuming we're receiving `data.story_group` based on this: https://github.com/bufferapp/buffer-web/blob/master/shared/services/StoryGroupBroadcaster.php#L194-L198

you can run this locally by pulling the latest buffer-web and changing the value in this file: shared/config/aws.php to `$config['aws_queues_local_use_elasticmq'] = false;`

then run `./dev worker start push && ./dev worker start service` 

at that point you can reload the publish ui + it should pick up the websocket connection to pusher, but I guess the part where you are testing on mobile itself it likely harder... so perhaps thats just testable on staging with the new mobile client

josem:
Hey @anafilipadealmeida! I think we'll need to make some adjustments to listen to ```private-story-groups-[profile_id]``` instead of the private-updates since those events will be sent through that channel.

packages/pusher-sync/middleware.js
```diff
@@ -98,6 +106,7 @@ export default ({ dispatch }) => {
       const { profileId } = action;
       if (profileId) {
         if (!channelsByProfileId[profileId]) {
+          console.log(action);
```
karelm: 👋

anafilipadealmeida: @karelm, we need to subscribe to different channels ( `private-updates-${profileId}` for queue and drafts changes and `private-story-groups-${profileId}` for stories changes. Since we can't subscribe multiple channels in the same pusher connection, I was thinking about listening to SELECT_TAB instead of SELECT_PROFILE, since it also gives a profileId, and subscribe to private-updates if queue/drafts/approval tabs, and to subscribe to private-story-groups if stories tab. Does this make sense, or am I understanding pusher wrong?

anafilipadealmeida:
@karelm, had to change back to `SELECT_PROFILE`. Apparently `SELECT_TAB` isn't dispatched the first time we load Publish 🤦‍♀ All ready, with this PR the stories queue removes a story when we click on Share Now and the reminder is sent

PR #984, opened 2019-09-25
Supporting multiple uploaders

karelm:
## Description

Providing support for multiple composers with different uploading / downloading behaviour.

## Context & Notes

Providing support for multiple composers with different uploading / downloading behaviour, as there was an issue when using the websocket library, that if you tried to configure new pusher events, it wouldn't set them up correctly because the pusher events would take the initial creator value, and so further events bound later would never re-bind new events to the composer.

This means that when uploading files in the composer, then switching to the stories composer, progress for uploading a video would never result in a new video asset being completed, or vice versa.

This changes that

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/UploadZone/utils/WebSocket.js
+++ b/packages/UploadZone/utils/WebSocket.js
@@ -77,11 +77,17 @@ class WebSocket {
 
   bind = ({ userId, notifiers }) => {
     this.eventHandlers = this.configureEventHandlers(notifiers);
+    let { pusher } = this;
 
-    Pusher.channel_auth_endpoint = PUSHER.AUTH_ENDPOINT;
-    const pusher = new Pusher(PUSHER.API_KEY, {
-      cluster: PUSHER.CLUSTER,
-    });
+    if (!pusher) {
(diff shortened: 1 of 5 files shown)


packages/UploadZone/utils/WebSocket.js
```diff
@@ -78,6 +68,31 @@ class WebSocket {
...
+      pusher = new Pusher(PUSHER.API_KEY, {
+        cluster: PUSHER.CLUSTER,
+      });
+
+      this.pusher = pusher;
+    }
+
+    const pusherInstance = pusher.subscribe(`private-updates-${userId}`);
+    this.eventHandlers.forEach((handler, event) => pusherInstance.bind(event, handler));
+    this.hasWebSocketConnectionOpen = true;
```
albennett: so bind & subscribe are methods from pusher?

karelm: Yes, those are pusher methods. This is the old code from the composer. Then we are putting a thing veneer over those events for our own use

PR #996, opened 2019-09-30
[PUB-2053] Story group queue on load more fix

MayaUribe:
## Description

- Refactoring onLoadMore method to be more readable by removing the switch case statement and replacing it with an Object literal.
- Fixes on load more bug for story groups queue.

## Context & Notes

JIRA card -> [PUB-2053](https://buffer.atlassian.net/browse/PUB-2053)

## Checklist

#### Staging Deployment

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "@storybook/addon-storyshots": "5.1.11",
     "@storybook/addons": "5.1.11",
     "@storybook/react": "5.1.11",
+    "@types/jest": "24.0.18",
     "axe-core": "2.2.0",
     "babel-eslint": "10.0.3",
     "babel-jest": "24.9.0",
--- a/packages/profile-page/index.js
+++ b/packages/profile-page/index.js
@@ -68,7 +68,7 @@ export default hot(
               profileId,
               page,
               isFetchingMore: true,
-              needsApproval: tabId === 'awaitingApproval' || tabId === 'pendingApproval',
+              needsApproval: ['awaitingApproval', 'pendingApproval'].includes(tabId),
             },
           }),
         );
--- a/packages/profile-page/index.js
+++ b/packages/profile-page/index.js
@@ -6,6 +6,20 @@ import { actions as dataFetchActions } from '@bufferapp/async-data-fetch';
 import { actions } from '@bufferapp/publish-tabs';
 import ProfilePage from './components/ProfilePage';
 
+const requestName = tabId => ({
+  queue: 'queuedPosts',
+  drafts: 'draftPosts',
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -60,6 +60,7 @@
     "@storybook/addon-storyshots": "5.1.11",
     "@storybook/addons": "5.1.11",
     "@storybook/react": "5.1.11",
+    "@types/jest": "24.0.18",
```
karelm: Hey @MayaUribe, I'm wondering how come we need @types/jest here?

MayaUribe: Oh this was something I was testing out locally, to be able to see the autocomplete in jest, which makes it a lot easier, but I can remove it

karelm: that sounds like a wonderful addition, great stuff 💯

packages/profile-page/index.js
```diff
@@ -47,43 +61,14 @@ export default hot(
...
-          case 'stories':
-            requestName = 'stories';
-            break;
-          default:
-            requestName = 'queued';
-        }
         dispatch(
           dataFetchActions.fetch({
-            name: `${requestName}Posts`,
+            name: getRequestName(tabId),
```
karelm: does this need a `getRequestName(tabId) || getRequestName('default')` like in the other place?

packages/profile-page/index.js
```diff
@@ -47,43 +61,14 @@ export default hot(
...
             args: {
               profileId,
               page,
               isFetchingMore: true,
-              needsApproval,
+              needsApproval: tabId === 'awaitingApproval' || tabId === 'pendingApproval',
```
karelm: would this be cleaner if it were a `['awaitingApproval', 'pendingApproval'].includes(tabId)`? then adding a new tabId is just another entry in an array?

packages/profile-page/index.js
karelm: wait.. I realise what I've done, ignore this... 😂 great stuff here 👍

MayaUribe: I decided to add this function so that when we don't pass a valid tabId it will return the default one, so basically if I now pass getRequestName('aRandomTabName') I'll get 'queuedPosts' as result.

MayaUribe: Another approach is not having the function at all and just the Object literal and do that call `getRequestName(tabId) || getRequestName('default')`

packages/profile-page/index.js
MayaUribe: Yess! I like that!! thank you for the suggestion 😄 🙌

PR #1004, opened 2019-10-02
Handle batches of updates loading at the same time

karelm:
## Description

Handle batches of updates loading at the same time, and making the dom re-render and the ui non-interactive

## Context & Notes

I believe that this should fix the issues we are seeing around delayed interaction
We are batching updates to the ui, and processing as many as we can at the same time

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/story-group-composer/reducer.js
+++ b/packages/story-group-composer/reducer.js
@@ -1,6 +1,7 @@
 import clonedeep from 'lodash.clonedeep';
 import keyWrapper from '@bufferapp/keywrapper';
 import { actionTypes as dataFetchActionTypes } from '@bufferapp/async-data-fetch';
+import debounce from 'lodash.debounce';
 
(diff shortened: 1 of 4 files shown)


packages/story-group-composer/reducer.js
```diff
@@ -361,6 +381,9 @@ export const actions = {
     stories,
     storyGroupId,
   }),
+  updateStoryPogress: debounce(dispatch => dispatch({
+    type: actionTypes.UPDATE_STORY_PROGRESS,
+  }), 200, { leading: false, trailing: true }),
```
MayaUribe: @karelm I guess after checking with Adam we could know is 200 is enough 😄

karelm: I made it that amount to try to make sure the UI still felt responsive, I think the problem is if we make that too large / small too many or too few updates could happen in the UI, and people might think things arent working. So I would lve to know what this feels like for others too :)

PR #1007, opened 2019-10-03
Use resized images via Safe Image service to improve performance and reduce image download size

karelm:
## Description
Use resized images via Safe Image service to improve performance and reduce image download size

## Checklist

#### Staging Deployment


packages/shared-components/Story/index.jsx
```diff
@@ -70,7 +71,7 @@ const Story = ({
             cardWidth={cardWidth}
             largeCards={largeCards}
           >
-            {card.thumbnail_url && <CoverImage src={card.thumbnail_url} />}
+            {card.thumbnail_url && <CoverImage src={getSafeimageUrl(card.thumbnail_url, 110, 198)} />}
```
MayaUribe: ❤️

karelm:
Hey @albennett, great question! The resized image in using here is cached on the cdn, and in the users local cache, so this is really fast once the initial image is created 🙂 and since I'm using a single size image across all preview locations, queue, composer and preview, we should only see 1 download happen. But the problem is that the thumbnail is not really a thumbnail... I'm not sure why, and it only really came to me when on that call earlier 😅 so I will dig into that too, but I definitely think the safe image will be a useful proxy in the meanwhile

albennett:
Great, thanks for explaining @karelm! And great work 🚀

PR #1008, opened 2019-10-03
Track if a Publish user is on the Analyze + Publish bundle

msanroman:
## Description

To be able to know if a given user on Publish is on the Analyze + Publish bundle experiment, we want to send `multiProductBundleName` with `analyze_publish_eid_19` to Segment.

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/globalAccount/middleware.js
+++ b/packages/globalAccount/middleware.js
@@ -18,6 +18,8 @@ export default ({ dispatch, getState }) => next => (action) => {
       dispatch(analyticsActions.init(state.globalAccount._id, {
         name: state.appSidebar.user.name,
         email: state.globalAccount.email,
+        multiProductBundleName: state.globalAccount.isAnalyzePublishBundle
+          ? 'analyze_publish_eid_19' : null,
       }));
       break;
     default:
--- a/packages/globalAccount/middleware.test.js
+++ b/packages/globalAccount/middleware.test.js
@@ -34,6 +34,25 @@ describe('middleware', () => {
       .toHaveBeenCalledWith('foo1', {
         name: 'The Great Foo',
         email: 'foo@buffer.com',
+        multiProductBundleName: null,
       });
   });
+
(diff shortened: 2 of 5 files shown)


packages/globalAccount/middleware.js
```diff
@@ -18,6 +18,8 @@ export default ({ dispatch, getState }) => next => (action) => {
       dispatch(analyticsActions.init(state.globalAccount._id, {
         name: state.appSidebar.user.name,
         email: state.globalAccount.email,
+        multiProductBundleName: state.globalAccount.isAnalyzePublishBundle
+          ? 'analyze_publish_eid_19' : undefined,
```
albennett: I've noticed segment can be a bit picky when sending undefined vs null. Just wanted to double check that it's ok to send `undefined`.

msanroman: Oooh, great catch! At first, I thought it would be best to send `undefined` because it would mean we wouldn't be sending the `multiProductBundleName` in the payload, but I think we might want to set it to `null` so that Segment notices and receives a change of value (if that happens) 😄 

Thank you, @albennett! 🌮 🙌

packages/globalAccount/middleware.js
```diff
@@ -18,6 +18,8 @@ export default ({ dispatch, getState }) => next => (action) => {
       dispatch(analyticsActions.init(state.globalAccount._id, {
         name: state.appSidebar.user.name,
         email: state.globalAccount.email,
+        multiProductBundleName: state.globalAccount.isAnalyzePublishBundle
+          ? 'analyze_publish_eid_19' : null,
```
karelm: does this name not come back anywhere?

msanroman: Ah dang, didn't see this @karelm! It doesn't for now 🤔

karelm: Okay, no worries. Just thought I'd check as I imagine we will want more plans later. But it's fine for now

PR #1015, opened 2019-10-07
PUB-2042 | Adding tracking for non conforming images (with aspect ratios not 9:16 or 16:9)

karelm:
## Description

PUB-2042 | Adding tracking for non conforming images (with aspect ratios not 9:16 or 16:9)

Adding tracking for segment.io for when composer uploads an image, to check whether users images are already in the expected aspect ratio or not.

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/analytics-middleware/transformers/publish/story.js
+++ b/packages/analytics-middleware/transformers/publish/story.js
@@ -93,6 +93,7 @@ dragged.propTypes = {
  * @param ctaVersion {string|null} What is the version number of the CTA? If setting the value for the first time, set to "1".
  * @param clientId {string|null} What was the unique identifier of the client the user was in when they uploaded a nonconforming image?
  * @param clientName {string|null} Which client was the user in when they uploaded a nonconforming image? IE, `publishIos`, `publishWeb`, `publishAndroid`.
+ * @param imageAspectRatio {string|null} The actual aspect ratio of the uploaded image.
  */
 export const nonConfirmingImageUploaded = ({
   storyGroupId = null,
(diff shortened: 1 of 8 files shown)


packages/analytics-middleware/transformers/publish/story.js
```diff
@@ -18,7 +18,7 @@ const removeNullObjectKeys = (data) => {
 
   if (data) {
     Object.keys(data)
-      .filter(key => data[key] !== null)
+      .filter(key => data[key] !== null && typeof data[key] !== 'undefined')
```
anafilipadealmeida:
```suggestion
      .filter(key => data[key] != null)
```
Would this be the same here?

packages/analytics-middleware/transformers/publish/story.js
```diff
@@ -73,3 +73,82 @@ dragged.propTypes = {
...
   clientName: PropTypes.string,
 };
+
+/**
+ * Story Nonconforming Image Uploaded
+ * When did the user upload an image to a story, where the image did not conform to the 9:16 ratio required by Instagram?
+ * https://github.com/bufferapp/tracking-plan/blob/master/tracking-plans/buffer-tracking-plan/events/product/publish/story-nonconforming-image-uploaded.yaml
+ *
+ * @param storyGroupId {string|null} The unique identifier for this story group in our production database.
+ * @param product {string} The product for which the story was created
```
anafilipadealmeida: I believe the product param is being sent in the middleware for every tracking event: https://github.com/bufferapp/buffer-publish/blob/master/packages/analytics-middleware/middleware.js#L16-L18
Not sure if it applies here, but maybe it's not needed 🙂

karelm: yup, thats the same here... but I think its useful to have it as a parameter still... the removeNullObjectKeys function will clean any not set values from the object, so should be fine 😀

PR #1021, opened 2019-10-08
Update analyze package versions

albennett:
## Description
Update analyze packages
https://buffer.slack.com/archives/C151RRDL1/p1570487151098400

## Checklist

#### Staging Deployment


federicoweber:
Hey @albennett sorry I think we pushed a small update this morning as well, can we push it to ` v1.12.1` 😄 ?

One thing we will also need to update the RPC endpoints, to match the new one, let me know if you need a hand with those 🙂.

albennett:
hey @federicoweber it looks like the `@bufferapp/server-analyze` is the only one that's `v1.12.1` and we're not importing it. 

Also, we're using logic to check if we should use the analyze api https://github.com/bufferapp/buffer-publish/blob/master/packages/server/rpc/analytics/compare/index.js#L14

Are you saying that we don't need to check anymore and instead default to the analyze api?

federicoweber:
Oh you are right @albennett, sorry I didn't notice that.

Yep that's right we are getting also Twitter data from Analyze API. So I think you want to replace all RPC endpoints in https://github.com/bufferapp/buffer-publish/tree/master/packages/server/rpc/analytics with the ones from Analyze (note that `postsAnalyzeApi` has been deprecated, we are using only `posts` now).

ps. might be worth to double-check if there is any difference in global variables names for the API address, I think you use some different ones in Publish.

albennett:
Great, thanks @federicoweber ! It looks like the endpoints are coming through ok. Regarding `analyticStartDate` & `audience` rpc packages -- are they also deprecated? 

I also noticed new rpc endpoints have been added to the `getReport` rpc package. Can you confirm these new endpoints are not needed for the publish charts?

federicoweber:
oh I noticed that tests for `postsSummary` are failing, worth fixing those before merging 😉

PR #1026, opened 2019-10-09
PUB-2028 | Add share now to story group composer

karelm:
## Description

PUB-2028 | Add share now to story group composer

There is now a shareNow button in the UI dropdown as well as "Schedule story" not really sure both options are needed, will check with Lauren on that.

## Context & Notes

I've had to add the ability to always press schedule now in edit mode (unless uploading, etc..), which will open up the date picker if the scheduledAt date is in the past.

Theres a small bug still left with this, which is when you select an option, it add's a tickbox... I need to review the component in BDS and remove the tick from appearing under some flag... maybe showSelectedItem or something 🤷‍♂ 

image

## Checklist

#### Staging Deployment


packages/story-group-composer/middleware.js
```diff
@@ -88,6 +88,11 @@ const getMappedStories = (story) => {
   return createImageStory(story);
 };
 
+const getScheduledAtTime = ({scheduledAt, shareNow}) => {
```
albennett: Why add five minutes? Is that how queue share now works?

packages/story-group-composer/components/AddStoryFooter/index.jsx
```diff
@@ -28,6 +28,40 @@ const getInitialDateTime = ({
...
 };
 
+const getIsScheduleDisabled = ({
+  editMode,
+  storiesLength,
+  uploadsCompleted,
+  isScheduleLoading,
+  isScheduledAtPastDue,
+}) => {
+  if (!editMode) {
```
albennett: I wonder if there's a more readable way to do this 🤔 It looks like some of the conditions are similar whether it's editMode or not. Maybe something like:  

 ```const conditions = storiesLength < 1 || !uploadsCompleted || isScheduleLoading;   
  if (!editMode && isScheduledAtPastDue)) {
  	return true;
  } else if (conditions) {
  	return true;
  } 
  return false;```

Either works, so whatever makes sense 😄

packages/story-group-composer/middleware.js
karelm: The scheduled at must be a date in the future, and shouldn't be null, so I've set it to 5 minutes in the future as it's a required field in the API, I should probably fix that issue instead of hack around it 😅

packages/story-group-composer/components/AddStoryFooter/index.jsx
karelm: Good point, this work definitely feels like it's my b game 🙃 I will review what I did and try to make some better choices. Thanks for your very useful input 🙌

packages/story-group-composer/middleware.js
MayaUribe: I did something similar for the share now (I added an hour 🙈), although I toke in consideration the timezone to that time, I don't know if by not specifying it could set a different time to the intended.

karelm: I've got some work to solve this server side instead with the create call, that will allow me to remove this 😄

packages/story-group-composer/components/AddStoryFooter/index.jsx
albennett: No worries! The share now logic is a complex one 😄

PR #1028, opened 2019-10-10
Revert "Revert "[PUB-2085] Open payment modal in Premium trialist banner CTA""

anafilipadealmeida:
Reverts bufferapp/buffer-publish#1027
Remove the import of switch-modal tracking utils from the switchModal rpc method, to avoid having to copy the package through the server.
Apply the parsing logic to stripe middleware instead, which fetches the switchModal rpc method.

### DIFF
--- a/packages/server/rpc/switchPlan/index.js
+++ b/packages/server/rpc/switchPlan/index.js
@@ -1,12 +1,12 @@
 const { method, createError } = require('@bufferapp/buffer-rpc');
 const rp = require('request-promise');
 
-const getCtaFromSource = require('@bufferapp/publish-switch-plan-modal/utils/tracking');
-
 module.exports = method(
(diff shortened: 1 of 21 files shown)


packages/server/rpc/switchPlan/index.js
```diff
@@ -38,7 +17,7 @@ module.exports = method(
         form: {
           cycle,
           payment_method_id: paymentMethodId,
-          cta: getCtaFromSource(source),
+          cta: source,
```
karelm: would it make more sense if this param was called cta instead of source?

anafilipadealmeida: That's actually a great idea! Done 😄

albennett:
Great approach 🙌 🚀 💯

PR #1033, opened 2019-10-16
Updating moment-timezone to latest version

karelm:
## Description

Updating moment-timezone to use the latest version from npm and updating to use the caret version number to support newer versions more automatically / systematically in future.

## Context & Notes

Had to run `yarn cache clean` to get this version to work with semantic versioning, this should be backwards compatible with previous changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "micro-rpc-client": "0.1.4",
     "mini-css-extract-plugin": "^0.5.0",
     "moment": "2.22.2",
-    "moment-timezone": "0.5.13",
+    "moment-timezone": "^0.5.25",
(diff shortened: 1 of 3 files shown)


karelm:
@colinscape I did this for consistency, I'm using the caret at the start, so npm will match all versions via semantic versioning. Packages we include already use the same semantic version of `^0.5.25` so though it was worth being consistent in all the places... This way the yarn.lock file has just one moment-timezone entry, so it was a little clearer to make sure we were using one version in all the places

colinscape:
@karelm Ah, I see! Thanks so much for indulging me - I'm not too familiar with `yarn`! 👍

PR #1049, opened 2019-10-23
[PUB-2128] Add thumbnail option to set user tags

albennett:
## Description

- Add thumbnail option to set user tags if conditions are met:
	-  Image post
	- Not a reminder
	- Feature flip (temp)
	- Pro and up users + team members
	- Instagram only

- Update warning message around reminders

https://buffer.atlassian.net/browse/PUB-2128

When we merge in the backend PR, I think it would be good to add enterprise user to the user endpoint instead of checking in the parser. Added it as a to-do in backend story: https://buffer.atlassian.net/browse/PUB-2132
https://hi.buffer.com/d86e42bb9fa4
## Checklist

#### Staging Deployment


packages/composer/composer/components/Composer.jsx
```diff
@@ -812,6 +812,11 @@ class Composer extends React.Component {
 
     const composerFooterClassName = usesImageFirstLayout ? styles.imageFirstFooter : styles.composerFooter;
 
+    const canAddUserTag = hasAccessToUserTag // on the user level including feature flip
```
albennett: Struggling with nomenclature -- any suggestions, Hamish?

packages/server/parsers/src/userParser.js
```diff
@@ -89,4 +91,7 @@ module.exports = userData => ({
   showReturnToClassic: userData.has_np_app_switcher,
   helpScoutConfig: userData.helpscout_beacon_params,
   isBusinessTeamMember: userData.is_business_team_member,
+  hasAccessToUserTag: userData.features.includes('instagram_user_tags')
+    && (userData.is_pro_premium_or_business_org_user // this includes team members
+    || isOnEnterprisePlan(userData.plan)), // TO-DO: Add to api
```
hamstu: Good point on the API, I wonder if we are often excluding Enterprise when we shouldn't because they fall into this special group. (I also wonder how many Enterprise customers are set as actual 'Enterprise' in the admin, since in the past I've seen customers that are managed as enterprise by our CS team, but actually have a large business account...)

packages/composer/composer/components/Composer.jsx
hamstu: Makes sense to me! 😅  I'd say keep it. 👍

packages/server/parsers/src/userParser.js
albennett: Yes, I thought for a while that enterprise was included in business but based on the user_model, that isn't always the case. When I rolled out NP to enterprise users, I realized they weren't included in the switcher logic. 

I wonder if we've excluded them in other areas. After, I impersonated an enterprise user to make sure they had access to the paid features (I think we often look to see if a user is not a free user, etc. so enterprise is normally included). Could be a good one to go through and double check though.

hamstu: Yeah, good point, that 'inverse' logic is probably catching some of them as well. We'll keep our eyes on this one.

PR #1056, opened 2019-10-25
[PUB-2130] Add user tag form UX changes

albennett:
## Description
Add user form UX changes
https://buffer.atlassian.net/browse/PUB-2130

- The Instagram User input should remain hidden until the photo is clicked. 
- “Add Tag” button should remain disabled until a user enters any text that is not a &nbsp;
- Tagged users should be added to a list below the input with a button to remove tag and new BDS icon button to open their profile in a new browser tab (revisit scroll logic)
 - Apply a bottom-border to all list items except last item
- Show “View Profile” tooltip when hovering over new BDS icon
- If the maximum number of 20 tags is reached, hide input and display note

## Context & Notes
What's left to do from the JIRA card:

- Add tracking when a tag is added 
- Add tracking when a tag is saved 
- The “@” symbol should persist in the input, even after a user has entered text 
- The Instagram User input should remain hidden until the photo is clicked. Once clicked, show the input and autofocus it (if we can do this in an a11y friendly way)

## Checklist

#### Staging Deployment


packages/i18n/translations/en-us.json
```diff
@@ -397,5 +397,17 @@
...
+    "rightHeader": "Tag Users",
+    "rightHeaderSubtext": "Click anywhere on the photo to tag a public user.",
+    "placeholder": "username",
+    "inputLabel": "Instagram User",
+    "inputBtnLabel": "Add Tag",
+    "footerText": "Instagram API limitations prevent searching for usernames and tagging private users.",
+    "maxLimitText": "You've reached the 20 tag limit.",
+    "btnSave": "Save",
+    "btnCancel": "Cancel",
+    "tooltip": "View profile"
```
hamstu: 💯 🎉

packages/composer-user-tags/components/ImageLabel/index.jsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const calculateTagStyles = ({ tag, showTags }) => ({
```
hamstu: 💅 Love this, and thought it would be worth mentioning that there's also a `styled-components` way of doing this too!

```jsx
const ImageLabel = styled.div`
  position: absolute;
  left: ${({ tag }) => tag.clientX}px;
  top: ${({ tag }) => tag.clientY}px;
  color: white;
  background-color: black;
  border-radius: 7%;
  padding: 2px 5px;
  opacity: 0.7;
  display: ${({ showTags }) => (showTags ? "block" : "none")};
`;
```

So basically `${}` can access the props that are given to the styled-component in question. You could also write it like `${(props) => /* something with props... */}`, but I used destructuring above.

No need to refactor by the way, just thought I'd share this one as I thought it was cool when I learned it. 😄

packages/composer-user-tags/components/TagInput/index.jsx
```diff
@@ -0,0 +1,65 @@
...
+import { Button, Text } from '@bufferapp/ui';
+import Input from '@bufferapp/ui/Input';
+
+import { ButtonWrapper, MaxCount } from './style';
+
+const TagInput = ({
+  translations,
+  inputValue,
+  setInputValue,
+  coordinates,
```
hamstu: Since we only use `coordinates.y` to determine the enabled / disabled state, maybe we can remove the need to pass `coordinates` down and just let the parent component send the correct enabled / disabled state?

packages/composer-user-tags/components/ImageLabel/index.jsx
albennett: Yess love it! Thanks for bringing it up. Will refactor.

PR #1060, opened 2019-10-30
[PUB-2157] Enable Awesome users to view Payday page

chayjay:
## Description

This change allows Awesome users to view the New Publish Plans page. It also disables multiple menu options for the user, making it impossible for them to access the New Publish dashboard after visiting the Plans page, unless they choose to upgrade.

## Context & Notes
This change is required for the work on moving more Awesome customers over to New Publish and onto the Pro or Premium plans.
[https://buffer.atlassian.net/browse/PUB-2157?atlOrigin=eyJpIjoiODI5YTM3YmQwNThmNGQ5Y2I2NWYyNzY4ZTEwZWVjMWMiLCJwIjoiaiJ9](https://buffer.atlassian.net/browse/PUB-2157?atlOrigin=eyJpIjoiODI5YTM3YmQwNThmNGQ5Y2I2NWYyNzY4ZTEwZWVjMWMiLCJwIjoiaiJ9)

## Checklist

#### Staging Deployment


packages/app-shell/components/AppShell/index.jsx
```diff
@@ -58,7 +58,11 @@ function generateUserMenuItems({
   switchPlan,
   openPreferences,
   showManageTeam,
+  isAwesomeUser,
```
hamstu: I wonder if we might consider a different variable name here? For example, if we have cause to render the `AppShell` in a similar way in the future (but for another reason) it might make more sense to call this `hideMenuItems` instead of `isAwesomeUser`. What do you think?

packages/app-shell/index.js
```diff
@@ -16,6 +16,7 @@ export default connect(
     showSwitchPlan: state.appShell.showSwitchPlan,
     showManageTeam: state.appShell.showManageTeam,
     showStartProTrial: state.appShell.showStartProTrial,
+    isAwesomeUser: state.appSidebar.user.isOnAwesomePlan,
```
hamstu: Instead of pulling this out of the user object in the `appSidebar` store perhaps we could modify the app-shell store so that it's available there same as the other values here? Then this could be something like `state.appShell.isOnAwesomePlan`.

On that note, we have a TON of places in the code where we reach across stores just like this, so don't feel bad. Especially common is reaching into `appSidebar`, but I think we'll want to move away from that. For example, the `appSidebar` is actually a relic of the old design before BDS and the new AppShell, we don't even use it's React component anymore! 

Ideally we'd remove it soon, but since so many packages depend on its store / user data... we can't ;) 

cc @anafilipadealmeida this is a great one for us to think about related to [improving state mgmt](https://paper.dropbox.com/doc/Publish-Engineering-Week-Nov-2019--AoJuFU0Eh6yBQlZOdwZzfnjzAg-ZxNAjn0mrzu08aQ5FD9F7#:uid=963272709250763488754179&h2=State-Management)! We could also consider removing the presentational components from `appSidebar` and renaming it to be some kind of central store for user data.

packages/beta-redirect/components/EnsurePublishBetaUser/index.jsx
```diff
@@ -4,13 +4,15 @@ import PropTypes from 'prop-types';
 
 import { BufferLoading } from '@bufferapp/publish-shared-components';
 
+const onPaydayPage = window.location.pathname.endsWith('plans');
```
hamstu: I noticed you've got this line in a few places, including the store. Ideally you could remove all references including this one and just pass the value down from the store – keep things more DRY. Does that make sense?

packages/beta-redirect/index.js
```diff
@@ -7,6 +7,7 @@ export default connect(
     hasPublishBeta: state.betaRedirect.hasPublishBeta,
     hasNewPublish: state.betaRedirect.hasNewPublish,
     loading: state.betaRedirect.loading,
+    onPaydayPage: state.betaRedirect.onPaydayPage,
```
hamstu: Re: the previous comment, you're already passing this down to the component, so you should be able to use it prettty easily :)

packages/beta-redirect/middleware.js
```diff
@@ -10,6 +10,8 @@ const getClassicBufferURL = () => {
   return 'https://buffer.com/app';
 };
 
+const onPaydayPage = window.location.pathname.endsWith('plans');
```
hamstu: 🗑 😄

packages/beta-redirect/reducer.js
```diff
@@ -20,6 +21,7 @@ export default (state = initialState, action) => {
         hasPublishBeta: features.includes('new_publish_beta'),
         hasPublishBetaRedirect: features.includes('new_publish_beta_redirect'),
         hasNewPublish,
+        onPaydayPage: window.location.pathname.endsWith('plans'),
```
hamstu: Keep just this pathname check, if that works 👍

packages/plans/index.js
```diff
@@ -16,6 +23,7 @@ export default connect(
     isNonprofit: state.appSidebar.user.isNonprofit,
     isOnProPlan: state.appSidebar.user.plan === 'pro',
     selectedPremiumPlan: state.plans.selectedPremiumPlan,
+    isAwesomeUser: state.appSidebar.user.isOnAwesomePlan,
```
hamstu: Related to the above, I'd actually say to _keep_ this one as it is... since a few other props already reference `appSidebar`, so we could clean it up another time.

packages/plans/index.js
```diff
@@ -27,7 +35,10 @@ export default connect(
       }
       dispatch(modalsActions.showSwitchPlanModal({ source, plan }));
     },
-    onBackToDashboardClick: ({ selectedProfileId, profiles }) => {
+    onBackToDashboardClick: ({ selectedProfileId, profiles, isAwesomeUser }) => {
+      if (isAwesomeUser) {
+        window.location.replace(getClassicBufferURL());
```
hamstu: Maybe toss a `return;` after this just to prevent any other code from running unnecessarily? 😄

packages/app-shell/components/AppShell/index.jsx
chayjay: Great idea!

packages/app-shell/index.js
chayjay: Understood, thanks for thinking through this and for all your great advice ☺

packages/beta-redirect/components/EnsurePublishBetaUser/index.jsx
chayjay: Absolutely!

packages/plans/index.js
chayjay: Done!

PR #1069, opened 2019-11-01
[PUB-2171] Only allow user to tag one username in a post

albennett:
## Description
Add an input error message if user tries to tag the same ig username twice in a post. 
https://buffer.atlassian.net/browse/PUB-2171
http://hi.buffer.com/b7cfd5661270 

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/composer-user-tags/components/UserTags/index.jsx
+++ b/packages/composer-user-tags/components/UserTags/index.jsx
@@ -70,7 +70,7 @@ const UserTags = ({
       return;
     }
 
-    if (inputError) setInputError(false);
+    setInputError(false);
 
     const { x, y, clientX, clientY } = coordinates;
     const userTag = {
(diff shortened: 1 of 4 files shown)


packages/composer-user-tags/components/UserTags/index.jsx
```diff
@@ -62,6 +63,15 @@ const UserTags = ({
...
 
   const addTag = e => {
+    const usernameAlreadyAdded = tags.some(tag => tag.username === inputValue);
+    if (usernameAlreadyAdded) {
+      setInputError(true);
+      if (e) e.preventDefault();
+      return;
+    }
+
+    if (inputError) setInputError(false);
```
hamstu: Small one, but since it doesn't matter if `inputError` is already false you can just remove the conditional here. 👍

albennett: I thought it might be better performance-wise to not set it every time but it prob doesn't matter much in this case 😄

hamstu: Yeah! I think in the case of a simple value it _shouldn't_ affect performance at all. (I.e., I think React will not re-render unncessarily), but I'm glad you're thinking about this! :raised

PR #1081, opened 2019-11-07
 PUB-2142 | Implementing Iterate script and removing qualaroo

karelm:
## Description

Adding Iterate script and removing qualaroo script to our codebase


packages/thirdParty/middleware.js
```diff
@@ -32,13 +32,25 @@ export default ({ dispatch, getState }) => next => (action) => {
...
-    case actionTypes.QUALAROO:
-      if (window && window._kiq) {
-        const { id } = action.result;
-        window._kiq.push(['identify', id]);
+    case actionTypes.ITERATE:
+      if (window && window.Iterate) {
+        const { result } = action;
+        window.Iterate('identify', {
+          first_name: result.name,
+          last_name: ' ',
```
karelm: this is done on purpose as we don't split these up ourselves

PR #1093, opened 2019-11-15
Increase loading speed, by making Stripe script Async and loaded with lower priority

karelm:
## Description
Increase loading speed, by making Stripe script Async and loaded with lower priority

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/credit-card-form/components/CreditCardForm/index.jsx
+++ b/packages/credit-card-form/components/CreditCardForm/index.jsx
@@ -9,7 +9,6 @@ import CreditCardForm from './form';
 const LoaderWrapper = styled.div`
   display: flex;
   justify-content: center;
-  vertical-align: center;
 `;
 
 class StripeWrapper extends Component {
--- a/packages/credit-card-form/components/CreditCardForm/index.jsx
+++ b/packages/credit-card-form/components/CreditCardForm/index.jsx
@@ -9,13 +9,28 @@ class StripeWrapper extends Component {
     createSetupIntentRequest: PropTypes.func.isRequired,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = { stripe: null };
+  }
+
   componentDidMount() {
+    const { stripePublishableKey } = this.props;
+
+    if (window.Stripe) {
+      this.setState({ stripe: window.Stripe(stripePublishableKey) });
+    } else {
+      document.querySelector('#stripe-js').addEventListener('load', () => {
+        // Create Stripe instance once Stripe.js loads
+        this.setState({ stripe: window.Stripe(stripePublishableKey) });
+      });
+    }
     const { createSetupIntentRequest } = this.props;
     createSetupIntentRequest();
   }
 
   render() {
-    const { stripePublishableKey, stripe } = this.props;
+    const { stripe } = this.state;
 
     return (
       <StripeProvider stripe={stripe}>
--- a/packages/credit-card-form/components/CreditCardForm/index.jsx
+++ b/packages/credit-card-form/components/CreditCardForm/index.jsx
@@ -1,8 +1,16 @@
 import React, { Component } from 'react';
 import { Elements, StripeProvider } from 'react-stripe-elements';
 import PropTypes from 'prop-types';
+import { LoadingAnimation } from '@bufferapp/components';
+import styled from 'styled-components';
 
 import CreditCardForm from './form';
+
+const LoaderWrapper = styled.div`
+  display: flex;
+  justify-content: center;
+  vertical-align: center;
+`;
 
 class StripeWrapper extends Component {
   static propTypes = {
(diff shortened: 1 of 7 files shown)


karelm:
If you remove the whitespace diff, you can see the direct changes... apologies if it muddies up the PR, but the file was auto-formatted by prettier as part of the work

packages/credit-card-form/components/CreditCardForm/index.jsx
```diff
@@ -9,13 +9,28 @@ class StripeWrapper extends Component {
...
+    const { stripePublishableKey } = this.props;
+
+    if (window.Stripe) {
+      this.setState({ stripe: window.Stripe(stripePublishableKey) });
+    } else {
+      document.querySelector('#stripe-js').addEventListener('load', () => {
+        // Create Stripe instance once Stripe.js loads
+        this.setState({ stripe: window.Stripe(stripePublishableKey) });
+      });
+    }
```
karelm: Hey @MayaUribe + @philippemiguet 

We are switching over to use window.Stripe if its available, or if not to load the stripe-js after its been initialized. It should be one or the other, so this code should be okay to replace the synchronous code... but I'm slightly concerned about the UI as a consequence of this change.

The reason I'm concerned about the UI is that stripe will be null, and I think that will mean that we won't show any stripe inputs, which makes me concerned about how we might better handle this? Perhaps introducing a loading spinner or something else here might be useful? Any input would be appreciated

philippemiguet: Hey @karelm that's a good one! I will loop in @oscoda to collaborate on this because we have a technical debt ticket to review this code and provide a component for teams to use.

@karelm, have you tried to remove `setState` lines above and leave `this.state.stripe = null` to see what happens when we open that modal? This will give us great insights on what may happen if Stripe isn't loaded yet and how we want to approach this "loading" state ;) 

Also, do you know if that change anything to pass the the stripeApiKey to the StripeProvider instead of the Stripe instance? This may not be possible with async loading though. 

From https://stripe.com/docs/recipes/elements-react:
```
<StripeProvider apiKey="pk_test_CvOaedJTBPQLmI0YSnQsitzN">
```

karelm: if stripe = null then we end up with the following:

image

as for just using the API key  🤔
that looks to solve the problem, I'm seeing the components show up when I specify the api key as the param instead... will test it out a little bit and see what happens 👍 

Thanks for the suggestions @philippemiguet  🙌

karelm: Actually... there is still an issue, if I use the `apiKey` property, it expects the javascript code to be there, whereas with the `stripe` instance via state, we are handling the presence or absence of stripe. So if a user loads the page without stripe having loaded yet whilst we are loading it asynchronously, there will also be issues + far worse an error being thrown.

I think we might need to handle the loading state for the credit card form if we are going to asynchronously load in the javascript code here. 

I would absolutely love your input here @oscoda as I'm hoping to get this wrapped up this week during Publish's engineering week. If you haven't got time though, I understand and can always investigate the approach necessary later 👍

oscoda: Hi @karelm , thanks for taking the time to handle this. I agree with your suggestion here:

> The reason I'm concerned about the UI is that stripe will be null, and I think that will mean that we won't show any stripe inputs, which makes me concerned about how we might better handle this? **Perhaps introducing a loading spinner or something else here might be useful**? Any input would be appreciated

I like this idea of displaying a spinner / loader while `this.state.stripe` is `null` and we're waiting for Stripe to load. 👍I know Juliana worked on [a Loader component recently](https://github.com/bufferapp/ui/pull/142), but this might be more used as an overall page loader, rather than indicating that a smaller section of the page is loading. I can't recall seeing a spinner component so you might need to create one for it, but I agree that this seems like the best approach. 

Somewhat related, but something I've been thinking about playing around with in Account is [content placeholders](https://www.codepicky.com/content-placeholders/), similar to what Slack and Facebook does. So instead of having a spinner and replacing it, we render a skeleton of the content and then load in the content when ready. We have some sections that just pop on-screen, so this approach would be a little less jarring for users. It doesn't quite work with a Stripe form since it's only a few input fields - so that's why I'd lean towards a small spinner. 🙏

packages/profile-sidebar/components/ProfileListItem/index.jsx
```diff
@@ -80,13 +85,12 @@ const ProfileListItem = ({
   };
 
   return (
-    <Link
-      href={'#'}
```
oscoda: Not entirely sure, but this might have been a `<Link />` component for accessibility?

packages/credit-card-form/components/CreditCardForm/index.jsx
philippemiguet: I love the idea of the content placeholder in buffer-account @oscoda 😍 

@karelm, it seems you have all you need to move this forward, right? Let us know if you need anything. And, thank you for driving this effort forward! This will be so helpful to other repositories 🤗

packages/profile-sidebar/components/ProfileListItem/index.jsx
karelm: This is a nested component, an anchor inside of another anchor. So this wasnt needed at this level, so I've removed it. Thanks for asking though 🙂

packages/credit-card-form/components/CreditCardForm/index.jsx
karelm: Yes, thanks @philippemiguet and @oscoda, I really appreciate your input 🙌 will get this loader/spinner in place and get it deployed to live 🤗

karelm: this is what the new version looks like:
http://hi.buffer.com/9cf7c650c56b

PR #1100, opened 2019-11-20
䷖ [PUB-2195] Add Bundle / Code Splitting

hamstu:
## Description

* Put most of node_modules in a `vendor` bundle
* Dynamically import the `composer` and the components that we borrow from Analyze to decrease size of main bundle.
* Changes to various things to support the above:
  * Use new buffer-static-uploader in the `pre-build.sh` that uses `-skip-versioning` on the uploaded files since we now add hashes at build time w/webpack. (This will improve our long-term caching when deps don't change between releases!)

## Context & Notes

* https://buffer.atlassian.net/browse/PUB-2195
* [Publish Eng. Week](https://paper.dropbox.com/doc/Publish-Engineering-Week-Nov-2019-ZxNAjn0mrzu08aQ5FD9F7)

## 🛑 Before Code Splitting

image

| Name      |   Gzipped Size |
|-----------|----------:|
| vendor.js | 130.57 KB |
| bundle.js |   1.31 MB |
| **Total initially downloaded** |   1.44 MB |

## ✅ After Code Splitting

image

| Name                                     | Gzipped Size| Diff.|
|------------------------------------------|----------:|------|
| vendor.907d8de8.js                       | 676 KB    | +135%\*\* |
| bundle.b31b802e.js                       | 250.84 KB | -80% 😍 |
| └ _[async]_ composer.b23a5841.js         | 288.26 KB |      |
| └ _[async]_ analyze.9571fdc9.js          | 291.2 KB  |      |
| └ _[async]_ analyze~composer.10ff86b9.js | 11.39 KB  |      |
| **Total initially downloaded**           | 926.84 KB | **-35.6%** 🤩 |

\*\* **Yes the `vendor` bundle is larger**, but we're now putting almost all of our `node_modules` deps in vendor now, instead of the main bundle. This will result in better performance on return visits as our users will be able to use their cached copy instead of having to fetch from our CDN again. (Until we change something in vendor of course, which would invalidate it.)

## Checklist

#### Staging Deployment


packages/composer-popover/components/ComposerWrapper/index.jsx
```diff
@@ -1,20 +1,29 @@
...
 import { actions as trialActions } from '@bufferapp/publish-trial';
 import { actions as analyticsActions } from '@bufferapp/publish-analytics-middleware';
+// import { bufferPublishComposer as Composer } from '@bufferapp/publish-composer';
 
-const ComposerWrapper = props => <Composer {...props} />;
+const Composer = lazy(() =>
+  import(
+    /* webpackChunkName: "composer" */ '@bufferapp/publish-composer/interfaces/buffer-publish'
+  )
+);
```
karelm: If we take a long time between loading the page and opening the composer, the lazy load can take a while as its got to download the composer package... could we add a loader whilst its still getting ready for those situations?

packages/analytics/components/AnalyticsList/index.jsx
```diff
@@ -1,126 +1,50 @@
...
-              <PostsTable />
-            </React.Fragment>
-          }
-        </div>
-      </ErrorBoundary>
-    );
-  }
-
-  return <Notification isInstagramBusiness={isInstagramBusiness} service={profile.service} />;
+  useEffect(() => {
```
msanroman: Using `useEffect` would trigger a fetch not only on mount, but also on update and on unmount. You might want to change this to a class or find a way to only do it on mount 🤔

packages/composer-popover/components/ComposerWrapper/index.jsx
msanroman: Yup, you could tweak the `fallback` on https://github.com/bufferapp/buffer-publish/pull/1100/files#diff-3fac5acdffa26334f99cf4134014dbf1R15 and create maybe a mocked Composer UI with a loading state :)

packages/analytics/components/AnalyticsList/index.jsx
anafilipadealmeida: I believe it's not doing it on unmount, as there's no return in the function 🤔Passing a second argument makes it skip the on update, but I'm not sure we could apply it here

anafilipadealmeida: In fact Hamish has the [] in the end, so in theory this is only happening on mount 😅

hamstu: Thanks @msanroman and @anafilipadealmeida! 

Yep, it's only running once, each time the component mounts thanks to the `[]`. [[ref](https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect)]

msanroman: Aaah, didn't see the second parameter! 👍 Thanks @anafilipadealmeida and @hamstu! 🙏

PR #1112, opened 2019-11-27
Fixes and improvements to code splitting

hamstu:
## Description

* Fixed code splitting that broke in https://github.com/bufferapp/buffer-publish/commit/a90f9216acf7eef0bef8f2c4d6cb8036bca9d679
* Fix hashing of CSS so that it's not linked to changes in JS
* Split out the runtime/manifest chunk and inline that into the HTML (production only) -- this should prevent hashes changing in the main bundles. [[ref]](https://survivejs.com/webpack/optimizing/separating-manifest/)
* Added new `ManifestPlugin` which outputs a JSON file that we're using now instead of `staticAssets.json` since it is keyed by non-hashed filenames. (cc @msanroman in case that might be helpful for Analzye, it simplified some of the server code.) [[ref]](https://github.com/bufferapp/buffer-publish/compare/task/code-splitting-improvements?expand=1#diff-068e7c6acbb89d55e9fa2192f813a6d8R66-R68)

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/server/index.js
+++ b/packages/server/index.js
@@ -40,6 +40,7 @@ const pusher = require('./lib/pusher');
 const maintenanceHandler = require('./maintenanceHandler');
 const { getFaviconCode, setupFaviconRoutes } = require('./lib/favicon');
 const { getBugsnagClient, getBugsnagScript } = require('./lib/bugsnag');
+const commonWebpackConfig = require('./webpack.config.common');
 
 const app = express();
 const server = http.createServer(app);
@@ -52,6 +53,7 @@ let segmentKey = 'qsP2UfgODyoJB3px9SDkGX5I6wDtdQ6a';
 setupFaviconRoutes(app, isProduction);
 
 let staticAssets = {
+  'runtime.js': 'https://local.buffer.com:8080/static/runtime.js', // webpack runtime
   'bundle.js': 'https://local.buffer.com:8080/static/bundle.js',
   'bundle.css': 'https://local.buffer.com:8080/static/bundle.css',
   'vendor.js': 'https://local.buffer.com:8080/static/vendor.js',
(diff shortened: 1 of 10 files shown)


packages/server/webpack.config.prod.js
```diff
@@ -2,19 +2,41 @@
...
+
 const plugins = [
   new MiniCssExtractPlugin({
-    filename: '[name].[chunkhash:8].css',
-    chunkFilename: '[name].[chunkhash:8].css',
+    /**
+     * Using `contenthash` here instead of `chunkhash` to ensure the CSS hash doesn't change unnecessarily
+     * See: https://survivejs.com/webpack/optimizing/adding-hashes-to-filenames/#setting-up-hashing
+     */
+    filename: '[name].[contenthash:8].css',
```
karelm: are we using contenthash for the js assets too?

packages/server/webpack.config.prod.js
```diff
@@ -23,14 +45,13 @@ if (process.env.ANALYZE_BUNDLE) {
...
-    nodeEnv: 'production',
-  },
+  optimization,
   plugins,
   output: {
-    publicPath: 'https://static.buffer.com/publish/',
+    publicPath,
     filename: '[name].[chunkhash:8].js',
   },
+  performance: { hints: false }, // don't warn that the bundles are big, we know ;)
```
karelm: > don't warn that the bundles are big, we know
totally!!!  😅

hamstu: haha, I had to get this one in!!

packages/server/index.js
```diff
@@ -52,6 +53,7 @@ let segmentKey = 'qsP2UfgODyoJB3px9SDkGX5I6wDtdQ6a';
 setupFaviconRoutes(app, isProduction);
 
 let staticAssets = {
+  'runtime.js': 'https://local.buffer.com:8080/static/runtime.js', // webpack runtime
```
karelm: I think theres a benefit in having the runtime present in all the bundles, because if any bundle loads first, it doesn't ruin the other bundles having been loaded, or something similar (can't quite remember)... so not sure if this is an ideal or not... but I guess its worth discovering if this one works

hamstu: Thanks @karelm! 🤔 Yeah, I _think_ this way is better. I read, for example:

> If the hashes webpack generates change, then the manifest changes as well. As a result, the contents of the vendor bundle change, and become invalidated. The problem can be eliminated by extracting the manifest to a file of its own or by writing it inline to the index.html of the project.
— [Separating a Manifest](https://survivejs.com/webpack/optimizing/separating-manifest/)

So I'm hoping this is a good change! Does that make sense?

hamstu: I've been making a lot of assumptions though. Hoping to run through some manual tests just to see if I'm thinking right. Ex. 'does changing this file or that file produce files/hashes that we expect?'

hamstu: And making sure that hashes stay consistent when we've not changed anything too! :)

PR #1142, opened 2019-12-10
Change color picker divs to buttons to make more accessible

albennett:
## Description
Change color picker elements to buttons to allow user to tab through and select color with keyboard.

## Context & Notes
I decided to change the elements to buttons as submit works a bit easier on keyboard. (If we kept div, we'd have to add a keyboard listener to submit on enter). 

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/shared-components/ColorPicker/styles.js
+++ b/packages/shared-components/ColorPicker/styles.js
@@ -60,13 +60,14 @@ export const StyledButton = styled(Button)`
   justify-content: space-between;
 `;
 
-export const CircleColor = styled.div`
+export const CircleColor = styled.button`
   position: relative;
   display: flex;
   align-items: center;
   justify-content: center;
   width: 22px;
   height: 22px;
+  padding: 0px;
   border-radius: 50%;
   background-color: ${props => props.color || DEFAULT_COLOR};
   color: ${props => props.colorContrast || colorSwatches.white};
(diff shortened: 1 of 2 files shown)


packages/shared-components/ColorPicker/styles.js
```diff
@@ -60,13 +60,14 @@ export const StyledButton = styled(Button)`
...
 
-export const CircleColor = styled.div`
+export const CircleColor = styled.button`
   position: relative;
   display: flex;
   align-items: center;
   justify-content: center;
   width: 22px;
   height: 22px;
+  padding: 0px;
```
albennett: buttons come with a set padding which was messing up the hover css.

packages/shared-components/ColorPicker/styles.js
```diff
@@ -81,7 +82,7 @@ export const CircleColor = styled.div`
       display: ${props => (props.selectable ? 'block' : 'none')};
       content: '';
       width: 30px;
-      height: 30px;
+      height: 28px;
```
albennett: I'm not quite sure why this needed to be changed 🤔Buttons must have 2px styling somewhere that caused this change.

MayaUribe: That's weird 🤔

packages/shared-components/ColorPicker/styles.js
MayaUribe: I didn't think about changing to button! makes sense 🙂

PR #1143, opened 2019-12-10
Upgrade to Node 12 LTS + Remove `buffermetrics`

hamstu:
## Description

* Upgrade to Node 12 LTS (matches the same version other services have been upgraded to.)
* Remove `buffermetrics` and all related tracking, and therefore no more `grpc` errors! 🎉 ❤️ 

## Context & Notes

* [Slack Thread](https://buffer.slack.com/archives/C151RRDL1/p1576010094028200) re: removing `buffermetrics` and the feedback modal.

## Checklist

#### Staging Deployment


albennett:
Hey @hamstu I haven't had a chance to look through all of this PR yet, but I did notice that you removed the app-switcher logic. Just wanted to confirm this is around removing storing feedback, and a user is still able to switch back to classic if they're on a b4b+ plan.

hamstu:
Thanks so much @albennett! Yep, I can confirm switching back still works. 👍

packages/server/index.js
```diff
@@ -294,14 +293,6 @@ app.use('*', (req, res, next) => {
 
 app.use(bodyParser.json());
```
albennett: Should we remove line 47 & 320 too? `const composerAjaxBuffemetrics = require('./lib/composerAjaxBuffermetrics');`
`app.post(
  '/ajax/buffermetrics',`

packages/server/index.js
```diff
@@ -395,5 +386,7 @@ app.get('*', (req, res) => {
 app.use(apiError);
 
 server.listen(80, () => console.log('listening on port 80')); // eslint-disable-line
+server.keepAliveTimeout = 61 * 1000;
```
albennett: Is this change just good practice or related to buffermetrics? Mainly curious :)

packages/server/index.js
hamstu: Ah yes, good Q! This is just an endpoint that the composer hits to track things. It doesn't do anything in publish, it just returns `'success'` 😂, but keeps the app from showing a bunch of 404s. 

So yeah, all that said it doesn't do anything with buffermetrics, since it's basically a no-op. :)

packages/server/index.js
hamstu: Yep, we need to do this as was discovered by some other teams, or we would probably start seeing timeouts. See https://threads.com/34372893337.

hamstu: To clarify and answer your Q better; it's related to the Node upgrade, not buffermetrics :)

packages/server/index.js
albennett: Ah, gotcha! Thanks for explaining :)

albennett: We may want to circle back in the future to remove all buffermetrics references in the composer that way we can remove these lines, too. 🗑

diannemcewan: @albennett @hamstu just curious, if we are busy removing buffermetrics from this codebase why would we not be removing the composer references too as a part of this work?

packages/queue/index.js
```diff
@@ -212,19 +211,10 @@ export default connect(
...
-        }
-      };
-      trackAction(
-        { location: 'queue', action: trackingAction },
-        {
-          success: openAfterTrack,
-          error: openAfterTrack,
-        }
-      );
+    onCalendarClick: weekOrMonth => {
```
karelm: does whatever is calling this need to be cleaned up aswell @hamstu ?

packages/sidebar/index.js
```diff
@@ -1,22 +1,10 @@
...
-import { actions as appSwitcherActions } from '@bufferapp/publish-app-switcher';
 
 import Sidebar from './components/Sidebar';
 
-export default connect(
-  state => ({
-    userFeatures: state.appSidebar.user.features,
-  }),
-  dispatch => ({
-    onReturnToClassicClick() {
```
karelm: I'm just wondering @hamstu does this need to have a onReturnToClassicClick function implemented for the sake of the AppSidebar? https://github.com/bufferapp/buffer-publish/blob/7242cb66a098b5c601d3c17cd2b553c514edecdd/packages/sidebar/components/Sidebar/index.jsx#L13

packages/server/index.js
hamstu: Good question @diannemcewan! The composer – as always 😅 – is not like the rest of Publish. While there are some references to `buffermetrics` in the composer code, it's not at all the same `buffermetrics` library I removed in this PR (We used the name `buffermetrics-*` for a lot of things over the years.). Instead, the tracking in the composer uses [this very old library](https://github.com/bufferapp/buffer-js-metrics/blob/master/src/Metrics.js) which `POST`s all tracking payloads to the endpoint `/ajax/buffermetrics` — which is what Amy Lee's question is about here.

When we moved the composer into New Publish some years ago we started to see a lot of 404s to `/ajax/buffermetrics`, since that is a route that existed on buffer.com (Classic) and not publish.buffer.com. To silence the errors until we figured out what to do I created a `/ajax/buffermetrics` route on the publish server that takes the payloads and simply returns 'success'. I did for a while try to pass along the events correctly, but that didn't work and actually briefly caused our data pipeline queue to nearly overflow, [so it was disabled.](https://github.com/bufferapp/buffer-publish/pull/508) — the reason wasn't clear at the time, and fixing composer tracking was never prioritized after that. 

All that said, it's been _years_ since we've had any of this legacy tracking working in the composer, so it is certainly not data we've needed, and is ripe for removing — it just felt like a task separate to this PR. :)

packages/queue/index.js
hamstu: Ah nice catch, I can fix that up ✨

packages/sidebar/index.js
hamstu: Yes... and no – but mostly no 😅 — We don't actually render the `sidebar` package anymore (it was the pre-BDS AppShell navigation) so not having this method will never cause an issue. We just use the package's `store` as [a place to pull user data](https://github.com/bufferapp/buffer-publish/search?q=state.appSidebar&unscoped_q=state.appSidebar), which is why the package is still around.

hamstu: I'll create an engineering task to remove `appSidebar` or rename it to be used as a more generic user data package.

PR #1193, opened 2020-01-22
fix(rpc): load the impersonated user in globalAccount method

esclapes:
When an admin impersonates a user, the `session.global` part of
the session remains pointing to the impersonating user

This change loads the global account for the impersonated user instead.

This is being fixed as part of an issue with Mixpanel tracking

## Description

The `getAccount` RPC enpoint will now use the publishId to fetch the user from the Core Authentication Service

## Context & Notes

During impersonation, data gets mixed in Mixpanel.

https://buffer.atlassian.net/browse/CORE-916

## Checklist

#### Staging Deployment


esclapes:
Hey @albennett 

>  Will this affect viewing past data since the user id will be different?

Nor sure what you mean about past data, but I think the `_id` used here won't really change across the app. This endpoint is usually called at app `INIT` so it will return a global Account. 

The main change here is making sure we fetch the impersonated user's global account instead of the impersonating user one. Before this PR you can see in the REDUX state that the `user` slice and the `globalAccount` slice show different information while impersonating.

I see how this is a bit tricky to fully grasp without looking at how impersonation works, but the gist of it is that we can pull a global account based on a the Publish user id, and that one is more reliable during imersonation that using the `accountId` directly.

Let me know if this makes sense :smile: , I am happy to share the difference in session content if you are curious about the impersonation.

packages/server/rpc/globalAccount/index.js
```diff
@@ -1,17 +1,20 @@
...
-//
 module.exports = method(
   'globalAccount',
   'fetch global account data',
   async (_, req) => {
-    const user = await getGlobalUser(req.session.global.accountId);
+    // Get the global account based on the publish session data
+    // because that's the only session key updated during impersonation
+    const publishId = req.session.publish.foreignKey;
+    const user = await authenticationService.getAccountForPublishId({
```
philippemiguet: @esclapes, can we call that variable `account` to better represent what it declares

esclapes:
I tried to rebase and something went wrong here :sweat_smile: closing this one to avoid issues.

packages/server/rpc/globalAccount/index.js
esclapes: Done in #1215

PR #1205, opened 2020-02-03
[PUB-2295] Awesome to Pro upgrade banner

chayjay:
## Description

Awesome to Pro upgrade banner for users who are being auto-upgraded to the Pro plan with billing changes.

## Context & Notes
[https://buffer.atlassian.net/browse/PUB-2295](https://buffer.atlassian.net/browse/PUB-2295)

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/temporary-banner/components/TemporaryDashboardBanner/index.jsx
+++ b/packages/temporary-banner/components/TemporaryDashboardBanner/index.jsx
@@ -25,6 +25,7 @@ const TemporaryDashboardBanner = ({
   enabledApplicationModes,
   displayRemindersBanner,
   usernamesRemindersList,
+  awesomeToProUpgradeDetails,
 }) => {
   const [hidden, hideBanner] = useState(false);
 
@@ -35,7 +36,11 @@ const TemporaryDashboardBanner = ({
   const getEnabledApplicationMode = tag =>
     enabledApplicationModes.filter(mode => mode.tag === tag)[0];
 
-  if (!enabledApplicationModes && !displayRemindersBanner) {
+  if (
+    !enabledApplicationModes &&
+    !awesomeToProUpgradeDetails &&
+    !displayRemindersBanner
+  ) {
     return null;
   }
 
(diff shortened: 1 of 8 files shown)


packages/temporary-banner/middleware.js
```diff
@@ -15,6 +16,15 @@ export default ({ dispatch }) => next => action => {
           },
         })
       );
+      if (
+        action.result.features.includes('awesome_pro_forced_upgrade_batch_2') // TODO: update to correct feature name
```
anafilipadealmeida: @chayjay, you might want to move this back to the original one 🙂

packages/temporary-banner/reducer.js
```diff
@@ -1,23 +1,53 @@
...
       return {
         ...state,
         enabledApplicationModes: action.result.enabledApplicationModes,
       };
+    case `awesomeToProUpgradeDetails_${dataFetchActionTypes.FETCH_SUCCESS}`:
+      return {
+        ...state,
+        awesomeToProUpgradeDetails:
+          action.result &&
+          action.result.success &&
```
anafilipadealmeida: i'm not very familiar with it, is this validation needed? If the value I want to use is `action.result.details.noticeHtml`, I validate we have action.result and action.details. If I add more conditions whose name doesn't seem clear, I usually create a variable with them outside with a better name.

packages/temporary-banner/middleware.js
chayjay: I'm not too sure what the actual feature name is going to be. I did change the read message back to the original one in my other PR though! https://github.com/bufferapp/buffer-web/pull/16775/files#diff-cbc48474f17890e24eba1543d58c8078R1336

packages/temporary-banner/reducer.js
chayjay: Fixed!

packages/temporary-banner/components/TemporaryDashboardBanner/index.jsx
```diff
@@ -49,6 +59,15 @@ const TemporaryDashboardBanner = ({
     });
   }
 
+  //  Displays Temporary Banner with Awesome to Pro Upgrade Details
+  if (awesomeToProUpgradeDetails) {
+    return TopBanner({
+      status: hidden,
+      content: awesomeToProUpgradeDetails,
+      onCloseBanner: onCloseBannerClick,
+    });
```
hamstu: Nitpick, I think it might be clearer to return a JSX component here: i.e., `<TopBanner status={hidden} ...>`.

packages/temporary-banner/middleware.js
hamstu: @chayjay My understanding is that this is the same batch of 2000 users we're building this for right now, so the feature would be the same as before: `awesome_pro_forced_upgrade_batch_1`.

(I would think of a `batch_2` as being all the rest of Awesome customers, whenever we get there - though maybe we'll use a different name if it is indeed everyone.)

PR #1222, opened 2020-02-21
[PUB-2515] Display campaign tag in queue

chayjay:
## Description

Update regular queue and past reminders to include the campaign tag for posts that have a campaign_id.

## Context & Notes
[https://buffer.atlassian.net/browse/PUB-2515](https://buffer.atlassian.net/browse/PUB-2515)

## Checklist

#### Staging Deployment


packages/server/parsers/src/postParser.js
```diff
@@ -216,5 +216,8 @@ module.exports = post => {
     sharedBy: post.shared_by,
     commentEnabled: post.comment_enabled,
     commentText: post.comment_text,
+    campaignName: '#SaveOurSeas Week',
+    campaignColor: '#9400D3',
+    campaign_id: '123',
```
MayaUribe: Heya @chayjay a suggestion here 🙌 this value will be returned from the api as `campaign_id`, however, because you are returning it in the parser, and the parser takes care of formatting the response from the api for us to keep consistency in the way we handle data in the frontend, I think it'd be nice to return it in camel case as `campaignId` instead, which has more consistency with the rest, what do yo think? 🙂

packages/shared-components/ImagePost/index.jsx
```diff
@@ -89,6 +89,9 @@ const ImagePost = ({
   hasPushNotifications,
   onSetRemindersClick,
   service_user_tags: userTags,
+  campaignName,
```
MayaUribe: A small one here @chayjay 😄 I think it might make sense to send these campaign details as `campaignDetails` instead, which would contain the `name` and `color`, the reason is: trying to think ahead, if we need to send a third value in a later version, we would need to change all these files by adding the new value (s), and by using an object here that encapsulates the campaigns information, we wouldn't need to go through all this process again, and we would only need to change it in the tag component itself, what do you think? 🤔

packages/campaigns/components/CampaignTag/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+  margin-left: 16px;
+  margin-bottom: 16px;
+  padding: 8px;
+  font-family: Roboto;
+  font-style: normal;
+  font-weight: 500;
+  font-size: 14px;
+  line-height: 16px;
+  opacity: ${props => (props.dragging ? 0 : 1)};
+  background-color: ${props => props.campaignColor};
```
MayaUribe: @chayjay I noticed here that the button is still maintaining the original button border (https://share.buffer.com/xQuggYAR), so I think you should change the border color to be the same as the background-color 🙂

packages/shared-components/Post/index.jsx
```diff
@@ -191,6 +195,16 @@ const Post = ({
               children
             )}
           </div>
+
+          {campaignName && (
+            <CampaignTag
+              campaignName={campaignName}
+              campaignColor={campaignColor}
+              dragging={dragging}
```
MayaUribe: @chayjay curious to know why we need this `dragging` prop for CampaignTag 🤔

packages/shared-components/Post/index.jsx
```diff
@@ -191,6 +195,16 @@ const Post = ({
               children
             )}
           </div>
+
+          {campaignName && (
```
MayaUribe: @chayjay I just noticed here that we are displaying this button for all users, we should be validating here that the user has the campaigns feature flip, otherwise everyone will be able to see this button if you merge this 😅, an example is this account I have without campaign feature flip and I can see it -> https://share.buffer.com/2NuBBEN8

packages/campaigns/components/CampaignTag/index.jsx
chayjay: I was wondering why it looked slightly off 😅 Thank you!

packages/shared-components/ImagePost/index.jsx
chayjay: I think that's a great idea, @MayaUribe. In the same vein, I'm thinking that getting a response from the backend with the id, name, and color as an object for the campaign details would help a lot with this particular task. I'm happy to speak more about this and left a note in the JIRA ticket associated with this task!

packages/shared-components/Post/index.jsx
chayjay: Absolutely, great point. I had planned on making this change before officially sending this PR out for review! 😄

packages/shared-components/Post/index.jsx
chayjay: Without this prop whenever I would drag an item in the queue, the campaign tag stayed present. I took a look at the other components with the dragging prop and followed the convention of basically giving the component an opacity of 0 whenever the queue item is being dragged. Hope this makes sense!

packages/server/parsers/src/postParser.js
```diff
@@ -216,5 +216,6 @@ module.exports = post => {
     sharedBy: post.shared_by,
     commentEnabled: post.comment_enabled,
     commentText: post.comment_text,
+    campaignDetails: post.campaign_details ? post.campaign_details : null,
```
MayaUribe: @chayjay great call 🙌

packages/shared-components/Post/index.jsx
```diff
@@ -191,6 +195,16 @@ const Post = ({
               children
             )}
           </div>
+
+          {hasCampaignsFeature && campaignDetails && (
```
MayaUribe: Great! thank you for adding this one 🙌

packages/shared-components/Post/index.jsx
MayaUribe: Oh I see, probably something we should look into and improve in the future 🤔 Thank you for the explanation!

chayjay: No problem!

packages/campaigns/components/CampaignTag/index.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import styled from 'styled-components';
+import { Button } from '@bufferapp/ui';
+
+const CampaignButton = styled(Button)`
+  margin-left: 16px;
+  margin-bottom: 16px;
+  padding: 8px;
+  font-family: Roboto;
```
albennett: Hey @chayjay there's a few styles that come by default with certain bds components. For example, button has these styles by default:  font-size: 14px; font-family: Roboto; and line-height: 16px; Here's where it happens in the ui repo: https://github.com/bufferapp/ui/blob/master/src/components/Button/style.js
https://github.com/bufferapp/ui/blob/master/src/components/style/fonts.js

If you do need to be more explicit with styles, you can often import the styles from the ui repo. An example would be: `import { red } from '@bufferapp/ui/style/colors';`

packages/queue/middleware.js
```diff
@@ -148,6 +154,22 @@ export default ({ dispatch, getState }) => next => action => {
         })
       );
       break;
+    case actionTypes.VIEW_CAMPAIGN_PAGE:
```
albennett: Curious what exactly the first dispatch does before it redirects to the campaign page route.

chayjay: The first dispatch highlights the 'Campaigns' tag in the sidebar. I'm not sure if [https://share.buffer.com/kpuYLpqY](https://share.buffer.com/kpuYLpqY) or [https://share.buffer.com/DOu8Q57Z](https://share.buffer.com/DOu8Q57Z) is preferred though I personally prefer the first!

packages/campaigns/components/CampaignTag/index.jsx
chayjay: Good catch! I removed the redundant styling 🙌🏾

packages/queue/middleware.js
chayjay: @albennett I took a look at the Figma file for viewing a single campaign and it looks like the Campaign tab [isn't highlighted in the view](https://share.buffer.com/RBudqelk) so I'mma remove the first dispatch for now! I'm happy to add it back at a later time if we decide to do so ☺

PR #1236, opened 2020-03-03
[PUB-2521] Add Edit functionality to CampaignForm

chayjay:
## Description

This PR will make it so the user has the ability to edit their previously set values for a campaign.
This PR also adds tracking to the Edit Campaign event.

## Context & Notes
PUB-2521

## Checklist

#### Staging Deployment


packages/i18n/translations/en-us.json
```diff
@@ -28,8 +28,9 @@
       "learnMore": "Learn More",
       "imageTag": "Example of campaign with multiple posts scheduled."
     },
-    "createCampaign": {
-      "title": "Create Campaign",
+    "campaignForm": {
```
MayaUribe: Nice refactor here @chayjay 😊

packages/campaigns/middleware.js
```diff
@@ -54,6 +54,30 @@ export default ({ dispatch }) => next => action => {
...
+      const { id, name, color, previousName, previousColor, orgId } = action;
+      const metadata = {
+        campaignId: id,
+        campaignName: name,
+        campaignColor: color,
+        previousCampaignName: previousName,
+        previousCampaignColor: previousColor,
+        organizationId: orgId,
+      };
+      dispatch(analyticsActions.trackEvent('Campaign Edited', metadata));
```
MayaUribe: Hey @chayjay not a big one, but we usually track these events on success, to make sure the event did happen and prevent tracking in case of an error 🤔

packages/campaigns/components/CampaignForm/index.jsx
```diff
@@ -79,13 +79,19 @@ const colors = [
...
+  inEditMode,
+  campaignDetails,
 }) => {
   // State
-  const [campaignName, setName] = useState('');
-  const [colorSelected, setColor] = useState(purple);
+  const defaultName = campaignDetails?.name ?? '';
+  const defaultColor = campaignDetails?.color ?? purple;
+  const campaignId = campaignDetails?.id ?? null;
+  const orgId = campaignDetails?.globalOrganizationId ?? null;
```
MayaUribe: Hey @chayjay I noticed you are creating a few variables here for campaignDetails, I think a nice refactor would be deconstructing these into a single one, something like this:

```suggestion
const {
    id: campaignId,
    globalOrganizationId: orgId = null,
    name: defaultName = '',
    color: defaultColor = purple,
} = campaignDetails ?? {};
```

packages/campaigns/components/CampaignForm/story.jsx
```diff
@@ -6,12 +6,28 @@ import translations from '@bufferapp/publish-i18n/translations/en-us.json';
...
 import CampaignForm from './index';
 
+const campaignDetails = {
+  id: '123',
+  name: '#SaveOurSeasWeek',
+  color: '#00C8CF',
+  globalOrganizationId: '123456',
+};
+
 storiesOf('Campaigns|CampaignForm', module)
```
MayaUribe: Hey @chayjay I have a few thoughts for this one 😊 I think is great you are testing this out with Stories as this helps isolate this component and truly making sure it's visually as we expect, however, I think that a good way to add more value in this tests would be to include a few more scenarios, for instance:
- How should this component look like if some of these campaignDetails were missing?
- How should this component look like if the campaign being edited doesn't exist?
- How should this component look like if there is an error updating the campaign?

packages/campaigns/middleware.js
chayjay: Hey @MayaUribe! I had a conversation about this with @anafilipadealmeida and she gave the recommendation to track this particular event before getting the success response due to the fact that we need the previous name and color for the tracking event. I'm happy to change this if we get all the tracking data we need from the backend response. This would be the name, color, previous name, previous color, campaign id, and the global organization id. 

I believe she cited that we do something similar for [Hashtag groups](https://github.com/bufferapp/buffer-publish/blob/master/packages/hashtag-group-manager/middleware.js#L61).

packages/campaigns/components/CampaignForm/story.jsx
chayjay: Great questions here @MayaUribe. I'm not entirely sure about all of the answers, should this be something I discuss with Lauren? Here are some of my thoughts:
> How should this component look like if some of these campaignDetails were missing?
I suppose it would suffice to have the default value be the value that it would be if it were a 'new' campaign. So the placeholder for the name and purple for the color.

> How should this component look like if the campaign being edited doesn't exist?
I had a few thoughts for this: 
- Direct them to their list of campaigns (or the campaign landing page, if they don't have any)
- Render the 'create' campaign form instead, so they can just create a new campaign
- Render the queue

I think for me it makes the most sense to direct them to their list of campaigns, but am definitely open to suggestions here.

> How should this component look like if there is an error updating the campaign?
I think there should be some kind of messaging around the error. I'm not sure how the design would look here though. Might this be a question for Lauren?

Thanks again for your thoughts here, Maya!

diannemcewan: Hey Char! 

Thanks for sharing your thoughts and questions so clearly.  Maya can respond to them. I just wanted to add that I would prefer if you don't yet merge what you have just yet because then the context of some of this work will be lost and once merged it is much harder to see the why in the PR's.

The goal behind merging in work and smaller PR's is to deliver working and complete functionality in smaller chunks. So think one small bit of completed functionality with tests.   The with tests is key, because the smaller the PR, the easier it is to see what should have tests or what tests are missing or what edge cases have not been catered for.

packages/campaigns/middleware.js
diannemcewan: @chayjay From my understanding, we are not sending the previous name and color anymore. I tend to agree with Maya here that if this request fails for any reason that our tracking of this change will not be accurate so doing this in the success response makes the most sense.  

Even if we did need to send them, I am a little confused about the issue around the previous name and previous color. When this component is rendered in edit mode you should already have this information from somewhere. You can't edit or update data without seeing what the previous values were. 🤔

packages/campaigns/components/CampaignForm/story.jsx
MayaUribe: Hey there @chayjay! Thank you for your reply, makes sense 🙂, however what I'm trying to convey is mostly, validating these edge cases, for instance, if the user doesn't have access and is trying to access a campaign form, should we display an error message? or just redirect to the queue? this is something that is also handled from the backend, (you'd probable get an error message when trying to get that campaign in the first place), but I think is important to think about in the frontend also 🤔

chayjay:
This is a strange one, but I believe the snapshots are failing on this PR because of an issue with timezones. There's a failing test that expects the date to be one day away from the actual date. Here are some screenshots showing what I mean:
[Screenshot of time expected given the unix timestamp](https://share.buffer.com/yAuv0G0R)
[Screenshot of failing test](https://share.buffer.com/12u1Pp00)

packages/campaigns/reducer.js
```diff
@@ -85,6 +86,22 @@ export const actions = {
     type: actionTypes.DELETE_CAMPAIGN,
     campaignId,
   }),
+  handleEditCampaignClick: ({
+    id,
+    name,
+    color,
+    previousName,
```
anafilipadealmeida: These 2 previous props are no longer being used for tracking, so you can delete them @chayjay

packages/campaigns/middleware.js
```diff
@@ -25,7 +25,7 @@ export default ({ dispatch }) => next => action => {
           args: {
             campaignId,
             past: false,
-            fullItems: false,
+            basicItems: true,
```
anafilipadealmeida: This argument was changed 7 days ago to `fullItems` across the code. The best way to confirm the argument name is to go to the rpc method: https://github.com/bufferapp/buffer-publish/blob/master/packages/server/rpc/campaigns/getCampaign/index.js. Switching this back to basicItems would cause a code regression

packages/campaigns/components/CampaignForm/story.jsx
```diff
@@ -6,12 +6,29 @@ import translations from '@bufferapp/publish-i18n/translations/en-us.json';
...
   .addDecorator(withA11y)
-  .add('default', () => (
+  .add('default create', () => (
```
anafilipadealmeida: This was named `default` because there was only one story here. Once you added a story for 'edit form', you could probably name this 'create form'. 🙂

packages/campaigns/components/CampaignsPage/index.jsx
```diff
@@ -50,7 +54,7 @@ const CampaignsPage = ({
           translations={translations}
           onCreatePostClick={() => {}}
           onDeleteCampaignClick={() => {}}
-          onEditCampaignClick={() => {}}
+          onEditCampaignClick={() => setViewMode(campaignPages.EDIT_CAMPAIGN)}
```
anafilipadealmeida: This function should be receiving the onEditCampaignClick from the props! If you take a look at packages/campaigns/index.js you'll see the logic there to redirect to the Edit_Campaign page, so no need to change the state here

packages/campaigns/components/CampaignsPage/index.jsx
```diff
@@ -61,19 +65,24 @@ const CampaignsPage = ({
           isUsingPublishAsTeamMember={isUsingPublishAsTeamMember}
           onOpenCampaign={() => setViewMode(campaignPages.VIEW_CAMPAIGN)}
           onOpenCreateCampaignClick={onOpenCreateCampaignClick}
+          onEditCampaign={() => setViewMode(campaignPages.EDIT_CAMPAIGN)}
```
anafilipadealmeida: You can delete this line of code @chayjay, there's an `onEditCampaignClick` already taking care of this logic in the following line

packages/campaigns/index.js
```diff
@@ -36,20 +36,43 @@ export default connect(
...
-        actions.handleCreateCampaignClick({
-          name: campaignName,
-          color: colorSelected,
-        })
-      );
+    onCreateOrUpdateCampaignClick: ({
+      campaignId,
+      colorSelected,
+      campaignName,
+      previousName,
```
anafilipadealmeida: These 2 previous props are no longer being used for tracking, so you can delete them @chayjay

chayjay:
Thank you for taking a look here @anafilipadealmeida. I'd love it if you could take one more look after merging the refactor!

packages/campaign-form/middleware.js
```diff
@@ -4,13 +4,29 @@ import {
...
+    case actionTypes.FETCH_CAMPAIGN: {
+      const { campaignId } = action;
+
+      dispatch(
+        dataFetchActions.fetch({
+          name: 'getCampaign',
+          args: {
+            campaignId,
+            past: false,
+            basicItems: true,
```
MayaUribe: Hey @chayjay `basicItems` was replaced by `fullItems` 🙂, fullItems=true if you want to fetch posts in the campaign, fullItems=false if you want the items without the posts.

packages/campaign-form/middleware.js
```diff
@@ -49,6 +80,22 @@ export default ({ dispatch }) => next => action => {
       break;
     }
 
+    case `updateCampaign_${dataFetchActionTypes.FETCH_SUCCESS}`: {
```
MayaUribe: (1/2) @chayjay is a good one to also add the `updateCampaign_${dataFetchActionTypes.FETCH_FAIL}` to handle the error, similar to the one for create campaign

packages/campaign-form/middleware.js
```diff
@@ -49,6 +80,22 @@ export default ({ dispatch }) => next => action => {
...
+      };
+      dispatch(analyticsActions.trackEvent('Campaign Edited', metadata));
+
+      if (id) {
+        dispatch(campaignScheduled.goTo({ campaignId: id }));
+      }
+      break;
+    }
+
     case `createCampaign_${dataFetchActionTypes.FETCH_FAIL}`:
```
MayaUribe: (2/2) Usually to prevent repeating these lines, we would just add the updateCampaign_FAIL here also, because both responses should display the same notification error.

MayaUribe:
Hey there @chayjay I just left a few comments here 😊

packages/campaign-form/middleware.js
```diff
@@ -11,6 +11,22 @@ import { actionTypes } from './reducer';
 export default ({ dispatch }) => next => action => {
   next(action);
   switch (action.type) {
+    case actionTypes.FETCH_CAMPAIGN: {
```
anafilipadealmeida: Hey @chayjay, I believe this action is no longer being used and can be deleted! If you take a look at packages/campaign-form/index.js, line 50, we are using the action from packages/campaign to fetch the data, so you no longer need this case in this middleware file

packages/campaign-form/reducer.js
```diff
@@ -3,6 +3,7 @@ import { actionTypes as dataFetchActionTypes } from '@bufferapp/async-data-fetch
 
 export const actionTypes = keyWrapper('CAMPAIGNS', {
```
anafilipadealmeida:
```suggestion
export const actionTypes = keyWrapper('CAMPAIGN_FORM', {
```

packages/campaign/index.js
```diff
@@ -37,6 +37,8 @@ export default connect(
       window.location.assign(`${getURL.getAnalyzeReportUrl()}`);
     },
     onEditCampaignClick: campaignId => {
+      console.log('clicked on edit campaign');
+      console.log('campaign id: ', campaignId);
```
anafilipadealmeida:
```suggestion
```

packages/campaign-form/middleware.js
chayjay: Thank you!!

packages/campaign/index.js
chayjay: Thank you for catching this 😅

PR #1241, opened 2020-03-05
Add RPC for deleting campaigns

josem:
## Description
Add RPC function to delete a campaign, I also had to add an extra condition to parse the right error message.

## Context & Notes
- [JIRA ticket](https://buffer.atlassian.net/browse/PUB-2545)

### DIFF
--- a/packages/server/rpc/campaigns/delete/index.test.js
+++ b/packages/server/rpc/campaigns/delete/index.test.js
@@ -1,0 +1,51 @@
+import RPCEndpoint from '.';
+
+const post = require('../../../requestMethods/post');
+
+jest.mock('../../../requestMethods/post');
+
+const accessToken = 'AN ACCESS TOKEN';
+const session = {
+  publish: {
+    accessToken,
+  },
+};
+
+const params = {
+  campaign_id: '12345',
+};
+
+const deleteCampaign = () => RPCEndpoint.fn(params, { session });
+const DELETE_CAMPAIGN_RESPONSE = {
+  message: 'The campaign was deleted',
+  success: true,
+};
+
+describe('RPC | Delete campaign', () => {
+  it('deletes a campaign correctly', async () => {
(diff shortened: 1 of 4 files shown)


packages/server/utils/index.js
```diff
@@ -3,9 +3,15 @@ const { createError } = require('@bufferapp/buffer-rpc');
 module.exports = {
   handleError: err => {
     if (err.error) {
-      const { message } = JSON.parse(err.error);
+      let { message } = JSON.parse(err.error);
+
+      if (!message) {
+        message = JSON.parse(err.message);
```
josem: I'm not sure how to test this line @MayaUribe, I think my test is not really representing that the error message will come in the "message" key instead of "error".

MayaUribe: Great catch @josem! I'm leaving you a code suggestion for that one!

packages/server/rpc/campaigns/delete/index.test.js
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      await deleteCampaign(params).then(response => {
+        throw new TypeError(response);
+      });
+    } catch (err) {
+      expect(err.message).toEqual(
+        "User does not have access to campaign or it doesn't exist"
+      );
+    }
+  });
```
MayaUribe:
```suggestion
    post.mockReturnValueOnce(
      Promise.reject(
        new Error(
          "User does not have access to campaign or it doesn't exist"
        )
      )
    );
    try {
      await deleteCampaign(params).then(response => {
        throw new Error(response);
      });
    } catch (err) {
	  expect(err.error).toBeUndefined();
      expect(err.message).toEqual(
        "User does not have access to campaign or it doesn't exist"
      );
    }
  });
```

MayaUribe: There is also a second approach to test this one:

```
post.mockReturnValueOnce(Promise.reject(new Error("User does not have access")));
await expect(deleteCampaign(params)).rejects.toThrow("User does not have access");
```

PR #1245, opened 2020-03-05
[PUB-2537] Change campaigns state

anafilipadealmeida:
## Description
- Change campaigns state to handle multiple campaigns.
- Fetch single campaign inside ViewCampaign component.

## Context & Notes
There's still a lot to do, like handle loading state, but it will be done in a different PR.
Also, the campaignId is now being stored when we fetch a single campaign or create one (in the future, once we have the campaigns list, clicking in one campaign should also change it), but I'll revisit it down the road to check if this approach will still makes sense!
https://buffer.atlassian.net/browse/PUB-2537

## Checklist

#### Staging Deployment


packages/campaigns/components/ViewCampaign/index.jsx
```diff
@@ -13,30 +13,36 @@ const Container = styled.div`
...
         isUsingPublishAsTeamMember={isUsingPublishAsTeamMember}
         translations={translations.viewCampaign}
         onCreatePostClick={onCreatePostClick}
         onDeleteCampaignClick={onDeleteCampaignClick}
         onEditCampaignClick={onEditCampaignClick}
       />
-      {hasPosts ? (
+      {campaignHasPosts ? (
         <React.Fragment>
           <nav role="navigation">
```
albennett: 💯

packages/campaigns/middleware.js
```diff
@@ -16,7 +16,7 @@ import { actionTypes } from './reducer';
 export default ({ dispatch }) => next => action => {
   next(action);
   switch (action.type) {
-    case actionTypes.HANDLE_CAMPAIGN_CLICK: {
+    case actionTypes.FETCH_CAMPAIGN: {
```
albennett: I may not have all the backend context, but curious why we're using dataFetch here instead of filtering the campaigns data in the redux store by campaignId.

anafilipadealmeida: Great question @albennett! I believe this is because of the posts! The fetch to getAllCampaigns will get us all the campaigns, but only with the basic details like name and color. This single getCampaign gives us only one campaign, but includes one property called items, that will render the updates needed to be displayed in the ViewCampaign. I guess at some point the discussion was around making an endpoint that would only give the posts without any other data, but this was the best solution for Analyze and Publish 🙂

packages/campaigns/reducer.js
```diff
@@ -37,12 +38,30 @@ export default (state = initialState, action) => {
...
-    case `createCampaign_${dataFetchActionTypes.FETCH_SUCCESS}`: {
-      const campaignDetails = action.result || {};
+    case `createCampaign_${dataFetchActionTypes.FETCH_SUCCESS}`:
+    case `getCampaign_${dataFetchActionTypes.FETCH_SUCCESS}`: {
+      let { campaigns, currentCampaign, campaignId } = state;
+
+      currentCampaign = action.result;
+      campaignId = action.result.id;
+
+      if (campaigns.some(campaign => campaign.id === action.result.id)) {
```
albennett: I often find method logic easier to read when it's set to a const. More of an opinion though, so feel free to ignore 😄

packages/campaigns/middleware.js
albennett: That makes sense, thanks @anafilipadealmeida !

packages/campaigns/reducer.js
anafilipadealmeida: It's not only about being easier to read! I should't be manipulating the state directly, this should be a pure function. Let me refactor it this afternoon to make it right 🙂

packages/campaigns/reducer.js
```diff
@@ -37,12 +38,30 @@ export default (state = initialState, action) => {
...
+        campaigns = campaigns.map(campaign => {
+          if (campaign.id === action.result.id) {
+            return action.result;
```
MayaUribe: Hey @anafilipadealmeida a small one, but I'm wondering if `action.result.id` here is the same as `campaignId` 🤔 and `action.result` the same as `currentCampaign`? I could be wrong and missing something obvious 😂

packages/campaigns/components/ViewCampaign/index.jsx
```diff
@@ -13,30 +13,36 @@ const Container = styled.div`
...
   isUsingPublishAsTeamMember,
   translations,
   onCreatePostClick,
   onDeleteCampaignClick,
   onEditCampaignClick,
+  campaignId,
+  fetchCampaign,
 }) => {
+  // Fetch Data
+  useEffect(() => {
```
MayaUribe: 😍

packages/campaigns/reducer.js
anafilipadealmeida: It should be ok now 😄 Also added some tests!

packages/campaigns/reducer.js
anafilipadealmeida: Not sure I got the question 😅 but added some tests that might make it more clear. Image our state has campaigns= [{id:'1', name:'foo'}, {id:'2', name: 'bar'}], and we update the second to {id: '2', name:'bas'}. This goes to the array of existing campaigns and for each campaign it checks if we are already in the campaign with id 2. Once it gets there, it replaces that object!

PR #1251, opened 2020-03-09
Refactor RPC methods with dependency injection

hamstu:
## Description

Further refactor and improve our RPC Methods. 

* Automatically `require()` all the ones in the `server/rpc` directory 
* Update `@bufferapp/buffer-rpc` with it's new dependency injection feature so methods are easier to write and test.

## Context & Notes

* All RPC Methods can now access `PublishAPI` and `parsers` by destructuring the last parameter to their method. See diff for examples of usage. For how this change works in the RPC Server see https://github.com/bufferapp/buffer-rpc/pull/10
* These changes make RPC methods more simple to write; [for an example see this diff](https://github.com/bufferapp/buffer-publish/pull/1251/files#diff-34cecff1daede82048d6ccb21b0bd193L1):
image

* This makes testing bit easier as we can inject a fake `PublishAPI` now instead of having to use jest to mock the import.
  * We still deep require the `parsers` to pass in the tests, but this could be mocked as well since I feel like the purpose in these RPC tests is not to test the parsers. (Ideally the parsers would be tested separately.) This would require rewriting larger parts of these tests however, so I left it out. cc @MayaUribe, would love to know what your thoughts are on that!

## Checklist

#### Staging Deployment


packages/server/rpc/index.js
```diff
@@ -1,173 +1,32 @@
...
-  removeUserTag,
-  getShopGridPostsMethod,
-  updateShopGridPostLink,
-  updateCustomLinksMethod,
-  updateSingleCustomLinkMethod,
-  deleteCustomLinkMethod,
-  createCampaign,
-  updateCampaign,
-  getCampaign,
+async function makeRPCHandler() {
```
MayaUribe: This is great @hamstu! ❤️

packages/server/rpc/campaigns/create/index.test.js
```diff
@@ -29,9 +24,20 @@ const CREATE_CAMPAIGN_RESPONSE = {
   success: true,
 };
 
+const PublishAPI = {
```
MayaUribe: 🙌

hamstu:
Thanks so much for the speedy review @MayaUribe! 💯

packages/server/rpc/index.js
hamstu: Thanks! It was starting to feel a little crazy how big that file was growing. 😅

PR #1254, opened 2020-03-13
RPC | Fetching campaigns list from API

MayaUribe:
## Description

Updated mockup response with actual endpoint call in RPC and updated the tests.

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/queue/middleware.js
+++ b/packages/queue/middleware.js
@@ -252,7 +252,7 @@ export default ({ dispatch, getState }) => next => action => {
 
     case `user_${dataFetchActionTypes.FETCH_SUCCESS}`: {
       const hasCampaignsFeature =
-        getState().appSidebar.user.features?.includes('campaigns') ?? false;
+        action.result?.features?.includes('campaigns') ?? false;
 
       if (hasCampaignsFeature) {
         dispatch(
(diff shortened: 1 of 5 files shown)


packages/queue/middleware.js
```diff
@@ -250,6 +250,21 @@ export default ({ dispatch, getState }) => next => action => {
       break;
     }
 
+    case `user_${dataFetchActionTypes.FETCH_SUCCESS}`: {
```
MayaUribe: I know this might not be ideal, but I'm thinking to fetch the campaigns list on user success, so the request happens only once, and because all we need is the user to be loaded to fetch the campaigns, and the reason I decided to add this here in the queue, is because I think is more efficient to request the list of campaigns once at start, than every time the user opens the composer, or every time the posts change.

packages/queue/reducer.js
```diff
@@ -436,6 +436,17 @@ export default (state = initialState, action) => {
         editMode: false,
         emptySlotMode: false,
       };
+
+    case `getCampaignsList_${dataFetchActionTypes.FETCH_SUCCESS}`:
+      return {
+        ...state,
```
MayaUribe: I'm planning on adding here the campaigns response and saving it in the state to use it in the composer, but I might do that in a different PR, I wanted to focus this PR only in fetching the data from the API.

PR #1256, opened 2020-03-16
[PUB-2585] Refactor campaigns structure

anafilipadealmeida:
## Description
- Move Campaign tag to composer package
- Delete Campaigns package
- Create campaign, campaigns-list and campaign-form packages
- Refactor campaigns routes
- Refactor how we select the Campaign button in the sidebar (replaced redux action by data from route)
- Move PageWrapper (with sidebar) to app-pages
- Delete campaigns redirect from initial-loading

## Context & Notes
https://buffer.atlassian.net/browse/PUB-2585

## Checklist

#### Staging Deployment


packages/campaign-form/middleware.js
```diff
@@ -73,14 +44,7 @@ export default ({ dispatch }) => next => action => {
...
       if (id) {
-        dispatch(
-          push(
-            generateCampaignPageRoute({
-              campaignId: id,
-              selectedPage: campaignPages.VIEW_CAMPAIGN,
-            })
-          )
-        );
+        dispatch(campaignScheduled.goTo({ campaignId: id }));
```
karelm: 😍

packages/campaign/index.js
```diff
@@ -0,0 +1,51 @@
...
+export default connect(
+  (state, ownProps) => {
+    return {
+      campaign: state.campaign.campaign,
+      showComposer: state.campaign.showComposer,
+      translations: state.i18n.translations.campaigns.viewCampaign,
+      isUsingPublishAsTeamMember:
+        state.appSidebar.user.isUsingPublishAsTeamMember,
+      isLoading: state.campaign.isLoading,
+      campaignId: state?.campaign?.campaignId || ownProps?.match?.params?.id,
```
hamstu: I think the first `?` is unnecessary, `state` should always be defined when this runs. Same for `ownProps` I would think.

packages/campaign/components/ViewCampaign/EmptyState/index.jsx
```diff
@@ -45,28 +40,7 @@ const EmptyState = ({
     <Button
       onClick={onCreatePostClick}
       type="primary"
-      isSplit
       label={translations.createPosts}
-      onSelectClick={selectedItem => {
```
albennett: Thanks for making this change for the updated figma!

packages/campaign/index.js
```diff
@@ -0,0 +1,51 @@
...
+      dispatch(modalsActions.showCloseComposerConfirmationModal());
+    },
+    onCreatePostClick: campaignId => {
+      dispatch(actions.handleOpenComposer(campaignId));
+    },
+    onDeleteCampaignClick: campaignId => {
+      dispatch(actions.handleDeleteCampaignClick(campaignId));
+    },
+    goToAnalyzeReport: () => {
+      window.location.assign(`${getURL.getAnalyzeReportUrl()}`);
```
hamstu: Since this is going to a regular URL I would rather we use a plain `<a href={url}>...</a>` vs. calling `window.location.assign`, for a11y and simplicity. I know we do this elsewhere but I'd like to start moving away from it for those reasons. 😄

packages/campaigns-list/index.js
```diff
@@ -0,0 +1,52 @@
...
+    goToAnalyzeReport: () => {
+      window.location.assign(`${getURL.getAnalyzeReportUrl()}`);
+    },
+    onEditCampaignClick: campaignId => {
+      if (campaignId) {
+        dispatch(campaignEdit.goTo({ campaignId }));
+      }
+    },
+    fetchCampaigns: () => {
+      dispatch(actions.fetchCampaigns());
```
hamstu: I wonder if we could optimize this a bit. `fetchCampaigns` here is dispatching an action that in turn just dispatches the _actual_ fetch action. 

Perhaps we could change this code to do that directly and skip the intermediate step. I.e., change the `dispatch` here so it calls `dataFetchActions.fetch({ name: 'getCampaignsList', ... });`

Let me know if that makes sense! 😄

packages/campaigns-list/middleware.js
```diff
@@ -0,0 +1,33 @@
...
+  switch (action.type) {
+    case actionTypes.FETCH_CAMPAIGNS: {
+      dispatch(
+        dataFetchActions.fetch({
+          name: 'getCampaignsList',
+          args: {},
+        })
+      );
+      break;
+    }
```
hamstu: 🔍 See previous comment, re: skipping this step.

packages/initial-loading/middleware.js
```diff
@@ -65,24 +59,6 @@ export default ({ getState, dispatch }) => next => action => {
...
-            campaignsActions.handleCampaignRouteLoaded({
-              campaignId: campaignParams.campaignId,
-              selectedPage: campaignParams.selectedPage,
-            })
-          );
-        } else {
-          window.location = getURL.getPublishUrl();
-        }
-      }
-
```
hamstu: ❤️ Love all the clean up here!

packages/campaign/index.js
anafilipadealmeida: True! Got a bit too defensive here 😂

packages/campaign/index.js
anafilipadealmeida: This is a great observation! I'm gonna leave it to a different time, as this is a weird situation: it's a button but it's linking to an external url. However I'm not sure if a11y link best practices apply here, as we might be considering navigation between Buffer products as internal navigation 🤔

packages/campaigns-list/index.js
anafilipadealmeida: I'm gonna change this in another PR, so I can also add proper testing 😅

packages/campaigns-list/components/ListCampaigns/index.jsx
```diff
@@ -27,7 +28,23 @@ const ListCampaigns = ({
   goToAnalyzeReport,
   onOpenCreateCampaignClick,
   isUsingPublishAsTeamMember,
+  hasCampaignsFlip,
+  fetchCampaigns,
+  isLoading,
 }) => {
+  if (!hasCampaignsFlip) {
+    window.location = getURL.getPublishUrl();
```
karelm: @hey @anafilipadealmeida, I'm just wondering how come this redirect is happening here? we already know that `!hasCampaignsFlip === true` at this point, so we've already determined that so could be be doing this in a better way/elesewhere... or at the very least a consistent way? firing off an action instead?

anafilipadealmeida: Not really, I guess 😅 Without the feature flip you cannot see the sidebar button for campaigns. However you can still type publish.buffer.com/campaigns. I've added it there so we don't forget about it, even though there should be a better way to do it!

karelm: Okay cool, thats fine. Just wanted to check with you 🙌

packages/campaign/index.js
hamstu: No worries leaving it for now, just something for us to think about. 👍  

I do feel like this sort of thing makes the most sense as a anchor/link, even if it looks like a button. For example, I don't think you can right click and 'open in new tab' for a button like this and so the user is forced to navigate away in the same page. I know that sort of thing has annoyed me before in other apps – but I might just be biased there 😅

PR #1259, opened 2020-03-17
PUB-2518  | Save campaign in a post in the composer

MayaUribe:
## Description

- Fetching real campaigns data and displaying it in the composer.

## Context & Notes

* Jira card | PUB-2518
* This PR is mainly to focus on the functional part for selecting a campaign when saving a post, I'll take care of the stylings in a new PR.
* The campaigns list in the composer is available in the queue, past reminders and sent posts.

## Screenshots
Image 2020-03-19 at 9 51 55 AM

## Checklist

#### Staging Deployment


packages/composer/composer/components/AppStateless.jsx
```diff
@@ -194,6 +194,8 @@ class AppStateless extends React.Component {
 
     const shouldDisplayCampaignHeader =
       userData?.hasCampaignsFlip && userData?.isProAndUpOrTeamMember && !draftMode;
+    const campaignId = metaData.campaignDetails?.id ?? null;
```
albennett: I was confused at first at the difference of `??` and `||` [but just looked it up](https://stackoverflow.com/questions/476436/is-there-a-null-coalescing-operator-in-javascript) and it makes sense. Neat that an empty string here would be `false`

packages/sent/reducer.js
```diff
@@ -197,6 +198,7 @@ export default (state = initialState, action) => {
       profileId = getProfileId(action);
       if (profileId) {
         return {
+          ...state,
```
albennett: @MayaUribe Found the bug 🕵️‍♀️ It looks like we were overwriting the original state with `byProfileId` because `...state` was missing. I wonder how long this has been as issue 😅

packages/queue/reducer.js
```diff
@@ -438,9 +439,9 @@ export default (state = initialState, action) => {
       };
 
     case `getCampaignsList_${dataFetchActionTypes.FETCH_SUCCESS}`:
-      // This is a WIP to save campaigns and fetch thos in the composer
       return {
         ...state,
+        campaigns: action.result,
```
albennett: I remember Lauren mentioned in this [jira comment](https://buffer.atlassian.net/browse/PUB-2457) to sort campaigns by last updated. We could change it to `sortCampaignsByUpdatedAt(action.result)` and add a sorting method:

```const sortCampaignsByUpdatedAt = campaigns =>
  campaigns?.sort((first, second) => first.updatedAt - second.updatedAt);```

albennett: It looks like `updatedAt` will have the value of `createdAt` if a user hasn't updated, so the value shouldn't ever be null. Just wanted to confirm though!

MayaUribe: @albennett! Jose already made those changes in the backend, we thought it made sense to have consistency for all clients 😊 however, is something we talked about yesterday to keep in mind, in case we want add a sorting option in the frontend, then we would have to add it here.

packages/sent/reducer.js
MayaUribe: *Thank youu!!!* @albennett this makes so much sense! and +1 to that, I wonder what else is failing because of this 🤔

packages/queue/reducer.js
albennett: Ah yes, so I think we were sorting it by created_at in the list view (which is the sorting happening in the backend), and in the composer dropdown Lauren made a request to sort it by updated_at (which could be different than created_at). Does that make sense?

MayaUribe: @albennett Aaaah that makes sense! my bad 😂 I missed that one! Will take care of it in a new PR then, thank you for letting me know ☺️

PR #1260, opened 2020-03-17
[PUB-2588] Fix Campaign tag

anafilipadealmeida:
## Description
- Refactor CampaignTag to become a reusable shared-component called PostTag.
- Use new route navigation on TagClick.

## Context & Notes
This is just a quick fix: I haven't changed the place of the tag to match the figma file nor added the post stories and snapshots.
https://buffer.atlassian.net/browse/PUB-2588

Screenshot 2020-03-17 at 12 32 07
Screenshot 2020-03-17 at 12 55 39

## Checklist

#### Staging Deployment


packages/shared-components/PostTag/index.jsx
```diff
@@ -0,0 +1,42 @@
...
+  border-color: ${props => props.color};
+  :hover {
+    background-color: ${props => props.color};
+    color: ${props => props.textColor};
+  }
+  color: ${props => props.textColor};
+`;
+
+const PostTag = ({ name, color, dragging, onPostTagClick }) => (
+  <div>
```
MayaUribe: Hey @anafilipadealmeida is there a particular reason why this is wrapped by a div? 🤔

anafilipadealmeida: Yes 😂 To have a display block and keep the size instead of taking a full-width, as the place where it is currently shown has display:flex 😅 Quickest fix I could find!

packages/queue/middleware.js
```diff
@@ -153,16 +149,15 @@ export default ({ dispatch, getState }) => next => action => {
...
-      dispatch(
-        push(
-          generateCampaignPageRoute({
-            campaignId: action.campaignId,
-            selectedPage: campaignPages.VIEW_CAMPAIGN,
-          })
-        )
-      );
+    case actionTypes.VIEW_CAMPAIGN_PAGE: {
+      const { campaignId, isSent } = action;
```
MayaUribe: @anafilipadealmeida maybe not important, but what would happen here if for some reason `campaignId` is undefined?

packages/shared-components/PostTag/index.jsx
anafilipadealmeida: However, thinking about it I'll remove the div and add the display: block to the styled-component 😄

packages/queue/middleware.js
anafilipadealmeida: Great question! It will generate something like: /campaigns/null/scheduled. Once it gets to that url it displays the list of campaigns instead and one error message saying something like: campaign not found

packages/shared-components/PostTag/index.jsx
anafilipadealmeida: Nop, it doesn't work. It has to be the parent 😅

PR #1262, opened 2020-03-17
Connect list of campaigns to single campaign

anafilipadealmeida:
## Description
- Small tweaks to make the transition between list of campaigns and single campaign work.
- Rename PageWrapper as PageWithSidebarWrapper.
- Create a second Switch with campaigns routes inside PageWithSidebar.

## Context & Notes
- If the Sidebar is part of the component being rendered, every time we change the route the sidebar will also be rendered. The idea is to keep both the appshell and the sidebar showing while we only render the content of each campaigns page. A react-router Switch only accepts Route and Redirect components, so we couldn't wrap the multiple campaigns routes with a single PageWithSidebar there. The best option I found was to create a second Switch inside PagesWithSidebar. 

- ~~The transition between profiles and campaigns still looks rough though. The profiles are being rendered with ProfilePage (which in my opinion is too complex for what it should) and weird routing. I could see us in the future moving all the tabs/sidebar click logic to rely only on routes, and all the pages that have a sidebar to use PageWithSidebar.~~ After talking to @karelm , I've decided to create a `<PageWithSidebarWrapper` to render the ProfileSidebar for both campaigns and ProfilePage. Doesn't change the fact we should refactor tabs and how we render the tabs content with a ProfilePage, but helps for now 😄 

- There's some width variation when navigating between the campaigns button and profiles on the sidebar, I'm gonna add it to this existing jira ticket, as I think it might be related with scrolling and the scroll bar: https://buffer.atlassian.net/browse/PUB-2589 

## Checklist

#### Staging Deployment


packages/campaign/components/ViewCampaign/index.jsx
```diff
@@ -49,7 +50,7 @@ const ViewCampaign = ({
   const campaignHasPosts = campaign?.items?.length > 0;
 
   if (isLoading) {
-    return <BufferLoading fullscreen />;
```
karelm: 😍 love the localized loading here 🙌

albennett:
Hey @anafilipadealmeida thanks for adding this change and thinking about our routing. I do want to spend some time looking into our router structure, as it feels like we shouldn't need to reload the sidebar when we click from the queue to campaigns.  (I know you've addressed some of this in in the context & notes 😄 ) 

🚀

anafilipadealmeida:
> Hey @anafilipadealmeida thanks for adding this change and thinking about our routing. I do want to spend some time looking into our router structure, as it feels like we shouldn't need to reload the sidebar when we click from the queue to campaigns. (I know you've addressed some of this in in the context & notes 😄 )
> 
> 🚀

That's a great point @albennett, and you are right!  I see us in the future refactoring how we handle routes in tabs and sidebar, so we can have all the page content being rendered in one place: AppPage. But for now the ProfilePage component is way too complex and attached to the concept of profileId to make the move 😅

PR #1279, opened 2020-03-26
[PUB-2537] Add campaign post actions

anafilipadealmeida:
## Context & Notes
https://buffer.atlassian.net/browse/PUB-2537

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/campaign/components/ViewCampaign/index.jsx
+++ b/packages/campaign/components/ViewCampaign/index.jsx
@@ -102,13 +102,7 @@ const ViewCampaign = ({
             onDeleteConfirmClick={null}
             onEditClick={null}
             onShareNowClick={null}
-            draggable={false}
             type="post"
-            hasFirstCommentFlip={null}
-            isBusinessAccount={null}
-            onPreviewClick={null}
-            serviceId={null}
-            userData={null}
           />
         </React.Fragment>
       ) : (
--- a/packages/campaign/components/ViewCampaign/index.jsx
+++ b/packages/campaign/components/ViewCampaign/index.jsx
@@ -30,19 +30,14 @@ const LoadingContainer = styled.div`
 const ViewCampaign = ({
   campaign,
   campaignPosts,
-  showComposer,
   isLoading,
   hideAnalyzeReport,
   translations,
-  onCreatePostClick,
-  onDeleteCampaignClick,
-  onEditCampaignClick,
   campaignId,
-  fetchCampaign,
-  goToAnalyzeReport,
-  onComposerCreateSuccess,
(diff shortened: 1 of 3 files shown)


packages/campaign/index.js
```diff
@@ -1,24 +1,24 @@
...
-        isManager: true, // temporary value
-        posts: state.campaign.campaign.items,
-        orderBy: 'dueAt',
-      });
-    }
     return {
       campaign: state.campaign.campaign,
-      campaignPosts,
+      campaignPosts: state.campaign.campaignPosts
+        ? formatPostLists({
```
MayaUribe: Hey @anafilipadealmeida I might be missing something, but isn't the `isManager` necessary as a parameter here?

packages/campaign/reducer.js
```diff
@@ -49,14 +79,146 @@ export default (state = initialState, action) => {
         showComposer: false,
       };
     }
+    // Pusher events
+    case queueActionTypes.POST_UPDATED: {
+      const postCampaignId = action?.post?.campaignDetails?.id;
+      if (
+        postCampaignId === state.campaign?.id &&
+        typeof postCampaignId === 'string'
```
MayaUribe: Hey @anafilipadealmeida just curious why we need a typeof validation here?

packages/campaign/components/ViewCampaign/index.jsx
```diff
@@ -94,28 +91,30 @@ const ViewCampaign = ({
               onTabClick={tabId => toggleView(tabId)}
             >
               <Tab tabId="scheduled">{translations.scheduled}</Tab>
-              <Tab tabId="sent">{translations.sent}</Tab>
+              <Tab tabId="sent">
+                {translations.sent}
+                <TabTag type="new" labelName="Coming Soon" />
```
MayaUribe: 😂

packages/campaign/index.js
anafilipadealmeida: Nop @MayaUribe ! 😄 In other queues we pass it here because we can only access it through profile. And what the function does is to pass the `isManager` to each update, to be used in `<Post>`. Since for campaigns we are already passing the` isManager` in each campaign post, no need!

packages/campaign/reducer.js
anafilipadealmeida: The first is to check if the post changed/added/deleted belongs to the campaign currently stored in the `state.campaign`. The second one is replacing 2 validations 😅: instead of checking for `null` and `undefined`, I'm confirming it's a string (to avoid storing the post in campaigns state when postCampaignId=null and state.campaign.id=null)

PR #1284, opened 2020-03-26
Fix `react-dnd` error by not propagating Dropzone drop

hamstu:
## Description

If we don't do this, file drops may propagate up from Dropzone to `react-dnd` and throw errors in the console (see https://github.com/react-dnd/react-dnd/issues/457)

This was happening with the story groups composer when dropping a file on one of the cards. (Though the error wasn't crashing the app; it just added noise to our Bugsnag reporting.)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,4 @@ vendor.css
 *.map
 *.vscode
 dist
+standalone-session.json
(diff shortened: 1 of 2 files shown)


hamstu:
Thanks for the speedy 👀 on this one @karelm 😄

PR #1286, opened 2020-03-27
[PUB-2609] Change campaign title to link in list item

albennett:
## Description
- Add a link to the campaign title in the campaign list. If clicked, redirect to the view campaign page.
- Remove references to `isEvenItem` (no longer being used)

## Context & Notes
Hey Ana, I added a getRoute method to routes. Is the campaignScheduled.routes property mainly for documentation? 

https://buffer.atlassian.net/browse/PUB-2609

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/campaigns-list/components/ListCampaigns/ListItem/index.jsx
+++ b/packages/campaigns-list/components/ListCampaigns/ListItem/index.jsx
@@ -5,6 +5,8 @@ import { Text, Button } from '@bufferapp/ui';
 import ClockIcon from '@bufferapp/ui/Icon/Icons/Clock';
 import ListIcon from '@bufferapp/ui/Icon/Icons/List';
 import CalendarIcon from '@bufferapp/ui/Icon/Icons/Calendar';
+import Link from '@bufferapp/ui/Link';
+import { campaignScheduled } from '@bufferapp/publish-routes';
 
 import {
   ButtonWrapper,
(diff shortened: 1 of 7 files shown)


anafilipadealmeida:
Hey @albennett, that's a great question! 
In fact it's not for documentation, `/campaigns/:id/scheduled/` is a way to write a pathname that react-router understands, and we use inside the <Router> (this one we use here: https://github.com/bufferapp/buffer-publish/blob/master/packages/app-pages/components/PagesWithSidebar/index.jsx#L33). It's how the Router knows that both `/campaigns/id1/scheduled` and `/campaigns/id2/scheduled` should render the same component, ViewCampaign.
The best thing about it is that we are able to get the :id info in our component (example here: https://github.com/bufferapp/buffer-publish/blob/master/packages/campaign/index.js#L25). So if you send someone to `/campaigns/id1/scheduled`, in the ViewCampaign you know `ownProps.match.params.id = 'id1'`, without having to run any extra functions to extract the id from the url  🙂

packages/campaigns-list/components/ListCampaigns/ListItem/index.jsx
```diff
@@ -50,12 +51,18 @@ const ListItem = ({
...
+
   return (
-    <Container isEvenItem={isEvenItem}>
+    <Container>
       <LeftWrapper>
         <Title>
           <Color color={campaign.color} />
-          <Text type="h3">{campaign.name}</Text>
+          <Text type="h3">
+            <StyledLink href={campaignRoute}>{campaign.name}</StyledLink>
```
anafilipadealmeida: @albennett I'm wondering, why is this a link? Would it be better in terms of semantics to consider this a button instead, since this is in-app navigation? If that's the case you no longer need a getRoute, you could use the campaignScheduled.goTo onClick

albennett: Hey @anafilipadealmeida great question. From my knowledge, [links](https://a11y-101.com/design/button-vs-link) should redirect you to a new page or view. Buttons trigger an action (modals, etc). Maybe the [react-router-dom Link component](https://knowbody.github.io/react-router-docs/api/Link.html) would be better? I know we have a pattern of using buttons and `push` to more programmatically navigate. Make we can chat more about it in a call?

PR #1290, opened 2020-03-30
[PUB-2592] Fix double spacebar crash in composer

hamstu:
## Description

On macOS pressing spacebar twice quickly adds a period and a space —
if left 'undhandled' this change in the input can cause draft-js to crash.
(One example is when this is done immediately after you type a link since
that triggers shortening to happen and draft-js / React gets confused
when trying to mutate the DOM since the sudden '.' is unexpected.) 

To solve this we look for when this 'double' character has been entered
and manually add it to the editorState ourselves - thus preventing the
unfortunate crashing.

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/composer/composer/components/Editor.jsx
+++ b/packages/composer/composer/components/Editor.jsx
@@ -1,7 +1,12 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Immutable from 'immutable';
-import { RichUtils, Modifier, EditorState } from '@bufferapp/draft-js';
+import {
+  RichUtils,
+  Modifier,
+  EditorState,
+  SelectionState,
+} from '@bufferapp/draft-js';
 import DraftjsEditor from '@bufferapp/draft-js-plugins-editor';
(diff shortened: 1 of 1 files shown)


packages/composer/composer/components/Editor.jsx
```diff
@@ -412,6 +417,71 @@ class Editor extends React.Component {
...
+      focusKey: selection.getFocusKey(),
+      focusOffset: selection.getFocusOffset() + move,
+    });
+  };
+
+  /**
+   * Handle input before it's processed by draft-js or rendered in React.
+   * See: https://draftjs.org/docs/api-reference-editor#handlebeforeinput
+   */
+  handleBeforeInput = (chars, editorState) => {
```
albennett: ah, so this is similar to how typed :emojis: work.

hamstu: Yeah! I think the plugins do a lot of this type of stuff, but just tucked away into their code. 😄 Definitely learned a lot more about draft-js in the process. The docs are a little terse sometimes, but it was enough, haha.

PR #1299, opened 2020-04-01
Update buffer-publish documentation

MayaUribe:
## Description

Update buffer-publish documentation to get started and coding guidelines:
- Update README.md file
- Adding a CONTRIBUTION.md file

## Context & Notes

In order to provide more context on how to get started and the guidelines you've developed as a team, I updated the documentation for this repo to give more clarity and transparency.

## Checklist

#### Staging Deployment

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -219,7 +219,7 @@ $ yarn run jest <path-to-package>
 $ yarn run jest ./packages/modals/reducer.test.js
 ```
 
-### Debugging
+### Debugging 🕵️‍♂️
 
 To use the `yarn test:debug` script, follow these instructions:
 1. Add a `debugger` statement near the failing line in your test.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,7 +8,8 @@ Here are some guidelines to make it as easy and clear as possible.
   - [Prettier 💁‍♀️](#prettier-💁‍♀️)
   - [Components Styleguide 📦](#components-styleguide-📦)
 - [Styling Styleguide 💅](#styling-styleguide-💅)
-- [Working on RPCs ](#working-on-rpcs)
+- [Working on RPCs](#working-on-rpcs)
+- [Testing 🧪](#testing-🧪)
 - [Reporting bugs 🐛](#reporting-bugs-🐛)
 
 ## Pull Requests
@@ -111,6 +112,37 @@ const Title = styled.div`
 
 ## Working with RPCs
 Refer to this [Notion](https://threads.com/34376693228) for more details on how to use the newest technique when working on RPCs.
+
+## Testing 🧪
+
+### Testing in Publish
+
+1. You can trigger a test in watch mode if you are working on a specific file or package:
+
+```bash
+$ yarn run test-package <path-to-package>
+
+# testing a specific file in watch mode
+$ yarn run test-package ./packages/modals/reducer.test.js
+```
+
+2. You can also test with `jest` directly:
+
+```bash
+$ yarn run jest <path-to-package>
+
+# testing a specific file
+$ yarn run jest ./packages/modals/reducer.test.js
+```
+
+### Debugging
+
+To use the `yarn test:debug` script, follow these instructions:
+1. Add a `debugger` statement near the failing line in your test.
+2. Type `chrome://inspect` in your chrome browser address bar.
+3. Click on "Open dedicated DevTools for Node".
+4. In your terminal run `yarn test:debug <path to test>`
+5. Visit the inspector you opened up, you should see that the debugger has been triggered and the app has paused near the line that is failing.
 
 ## Reporting bugs 🐛
 To report bugs, please feel free to add them in [JIRA](https://buffer.atlassian.net/secure/RapidBoard.jspa?projectKey=PUB)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -96,7 +96,18 @@ The name of the folder is up to you, what's more important is the name of packag
     * Do this when you're customizing your package, and  whenever you change the dependencies in your package or another.
 
 ## Styling Styleguide 💅
-For our styling we use [styled-components](https://styled-components.com/).
+For our styling we use [styled-components](https://styled-components.com/), an example for a styled component:
+
+```js
+const Title = styled.div`
+  display: flex;
+  align-items: baseline;
+  justify-content: center;
+  flex-direction: row;
+`;
+
+<Title>Example</Title>
+```
 
 ## Working on RPCs
 Refer to this [Notion](https://threads.com/34376693228) for more details on how to use the newest technique when working on RPCs.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -15,9 +15,21 @@ Here are some guidelines to make it as easy and clear as possible.
 Please follow the steps for your contribution:
 1. Fork the repo and create your branch from `master`.
 2. Follow the [coding styleguide](#coding-style)
-2. If you've added code that should be tested, add tests.
-3. Ensure the test suite passes.
-4. After submitting your pull request, verify that all status checks are passing.
+3. If you've added code that should be tested, add tests.
+4. Ensure the test suite passes.
+    <details>
+    <summary>What if the status checks are failing?</summary>
+    <p>
+
+    If `continuous-integration/travis-ci/push` is failing, you can try updating the snapshots file by running the following command and committing the resulting snapshot file:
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+// Functional component
+const Welcome = ({name}) => <h1>Hello, {name}</h1>
+
+// Class component
+class Welcome extends React.Component {
+  render() {
+    return <h1>Hello, {this.props.name}</h1>;
+  }
+}
+```
```
karelm, line 66: for me this only applies if you are creating a simple component. If your component is more technical, or has added logic I personally prefer using class components as you can create functions/methods and they aren't executed in the render logic. This is the biggest drawback to functional components to me, if you aren't careful you could be running some expensive code in there. I think that hooks mitigates some of this, but I'm not certain thats a good enough reason / way to deal with state in these particular components

I would suggest we make this statement explicit, that your primary approach should be functional components, but anything more advanced is okay to use a class component.

CONTRIBUTING.md
```diff
@@ -0,0 +1,233 @@
...
+    * Look at other packages for examples!
+
+6. **Connect reducer and/or middleware**
+    * If you have a reducer and/or middleware - don't forget to link those up in `packages/store/index.js` and `packages/store/reducers.js`.
+
+7. **Run `yarn`**
+    * Do this when you're customizing your package, and  whenever you change the dependencies in your package or another.
+
+**Component PropTypes and DefaultProps**
+Avoid using defaultProps for anything that's a required prop.
```
karelm, line 104: should be encourage default props to be passed in via spread operator in the definition instead? 

```const Welcome = ({name = 'John Smith'}) => <h1>Hello, {name}</h1>```

File: CONTRIBUTING.md [hunk 3]
```diff
@@ -0,0 +1,233 @@
...
+Adding packages to a monorepo is slightly different than adding to a standard node package. Common `devDependencies` can be added to the top level `package.json` file.
+
+### Adding A Common Dependencies
+
+This is the most likely scenario you'll face.
+
+in the root directory (`buffer-publish/`) run the follwing commands:
+
+  ```bash
+  $ yarn add -DE some-cool-package
```
karelm, line 117: can this name be changed to something like `@bufferapp/publish-cool-package` to give a more accurate hint at our naming practice?

CONTRIBUTING.md
hamstu: I disagree. (And I know you're in part just expressing your opinion here @karelm, so no offense intended at all! Just wanted to provide an alternative viewpoint. 🤗)

I think we should almost always use functional components. This appears to be the way React is heading, and it feels good to move with them in that direction. (Not saying they'll remove Class components soon, but I wouldn't rule it out in the long-term. There's [almost nothing you can't do with a functional components now](https://reactjs.org/docs/hooks-faq.html#do-hooks-cover-all-use-cases-for-classes).)

One thought from Dan Abramov that I like in his article _(emphasis mine)_:

> Maybe you’ve heard one of them is better for performance. Which one? Many of such benchmarks are flawed so I’d be careful drawing conclusions from them. **Performance primarily depends on what the code is doing rather than whether you chose a function or a class. In our observation, the performance differences are negligible**, though optimization strategies are a bit different.

[[source](https://overreacted.io/how-are-function-components-different-from-classes/)]

But I don't want to come across as overly dogmatic here! I just think _should_ strongly encourage use of functional components with hooks. (And perhaps we can clarify the hooks part here with some links to [the very helpful React docs](https://reactjs.org/docs/hooks-faq.html).

karelm: Thanks for your thought on this @hamstu I'm not in disagreement, I just think that for most situations functional components are the right choice but there will be situations where they won't be. So we should be explicit. I put an example in the thread Maya created, which could be made using hooks, but I still think that having logic in a class component is clearer than having all logic inside a render function with hooks to delay execution. Anyhow, happy to chat more as I know this can be contentious 😅. 

Essentially I agree that functional components should be our default, so we should be explicit, and then call out examples where we might use a class component instead.

hamstu: Thanks so much @karelm! Really appreciate your response. 🙏 

> I still think that having logic in a class component is clearer than having all logic inside a render function with hooks to delay execution

Totally agree we should avoid scenarios where we're losing performance like that - and perhaps that's an easier traps to fall into with functional components and hooks. But if that's indeed happening then I would argue that we are writing our functional components incorrectly. 

For example your ImageUploader example from Threads could look like this:

```jsx
import React, { Fragment, useState } from "react";

function getBinaryImageData(event) {
  return "foo";
}

function ImageUploader() {
  const [image, setImage] = useState(null);
  if (image) {
    return (
      <Fragment>
        <input type="file" onchange={e => setImage(getBinaryImageData(e))} />
        <img src={image} alt="Example" />
      </Fragment>
    );
  }
  return <input type="file" onchange={e => setImage(getBinaryImageData(e))} />;
}

export default ImageUploader;
```

Since `getBinaryImageData` doesn't need to be inside the component (it could even be a method in another module/file) we can just define it as another method outside of the component.

I prefer that approach as I think it can help avoid adding lots of logic into a Class component where it should be more concerned with rendering instead.

In the case that methods need to be defined inside of functional components there are hooks like `useMemo`, or `useCallback` to ensure that they remain well optimized.

> happy to chat more as I know this can be contentious

🙌 Likewise. :) 

I'll readily admit that I have my own preferences and bias' here. I'm super appreciative of your thoughtful original comment and response to mine. This is a fun discussion to have, and at the end of the day I think we agree on 99% of what matters here.

karelm: Thanks @hamstu, I think we are essentially in agreement here 👍 

Thanks for sharing my example but in a more hooks focussed approach 🙌  
We should show that as how you might implement a class component as a functional component 🤔 (but maybe with the getBinaryData function exported into its own file so it can be stubbed. I'm  grateful for your input and totally agree that for 99% of occasions this will be the defacto way we do things. 

This is going to be our approach with functional components and is going to make things more testable, because you can then stub an import for a function like getBinaryImageData with a mock function and export it into its own file, then test the functional component, and the getBinaryImageData functions in isolation of each other.

I'm just wary of people adding code that reads like this instead of your example above:
```
import React, { Fragment, useState } from "react";

function ImageUploader() {
  const [image, setImage] = useState(null);
  const getBinaryImageData = function(event) {
    return "foo";
  };
  if (image) {
    return (
      <Fragment>
        <input type="file" onchange={e => setImage(getBinaryImageData(e))} />
        <img src={image} alt="Example" />
      </Fragment>
    );
  }
  return <input type="file" onchange={e => setImage(getBinaryImageData(e))} />;
}

export default ImageUploader;
```
Then all the function declarations and logic is hard coded into the body of the functional component, we are creating the getBinaryImageData function on every re-render of the component. This adds the potential to add more complexity into the body of the functional component with every piece of additional logic we include, we want to avoid this kind of thing at all costs. This is worlds apart from what we want to achieve as you could see that if we let things like this creep in slowly it could get messy. If we used a class component instead, we could have confidence that the declaration will only happen once when the component is defined and not on every re-render.

Yes, this is a contrived example and functional components with fully isolated functionality would solve this as well. I just wanted to discuss this a bit more explicitly. Maybe we should just go full FP with our components and interaction methods, and be explicit about it. I think it would achieve the same and have the benefit that everything would be so much more "pure" and testable. 

I think that functional components will be the default for 99% of cases, but there will be situations where this won't make sense and we should be concious of that.

anafilipadealmeida: I'm just coming here to say I LOVE seeing us having this type of discussions! My two cents here: I agree functional components allow us to achieve the goals most of the time and that usually the problem is not about what approach we use, but how we use it instead.

I know we've been leaning towards showing how to write code in a good way, but I honestly feel it would be very beneficial to write an example in our documentation with also what we should avoid. We could even grab this example and show 3 situations: the default/preferable way with functional component; the accepted way with class component; the bad way/mistakes you have to avoid.

MayaUribe: This was a very interesting thread to read, and when I included this change in our guidelines:

> For new components, use Functional Components over Class Components

The main reason is due to @hamstu comments, it feels like React is heading that way, so I thought it made sense to add it as a default, however I do understand @karelm's point, and I agree we can be more explicit about it, so I'll add a few examples as @anafilipadealmeida suggested:

> show 3 situations: the default/preferable way with functional component; the accepted way with class component; the bad way/mistakes you have to avoid.

Thank you for all the comments here 👏❤️, this is exactly what I wanted to see, all of your opinion is very important to make a clear and effective guideline 👌

PR #1300, opened 2020-04-02
[PUB-2629] Listen for campaigns pusher events

hamstu:
## Description
Listen and update state with campaigns pusher events. (create campaign, update campaign, delete campaign)

## Context & Notes
https://buffer.atlassian.net/browse/PUB-2629
- Refactored to not store campaigns in other packages, but to reference campaigns from `campaignsList` state. 
- Refactored to only fetch campaigns if needed (if user has campaign feature & campaigns hasn't been fetched yet). This will help with less loading state & performance. 

## Checklist

#### Staging Deployment


albennett:
@hamstu tagging you here to review as I just remembered you created this PR 🤠

packages/pusher-sync/middleware.js
```diff
@@ -130,45 +135,92 @@ const bindProfileStoryGroupEvents = (channel, profileId, dispatch) => {
...
-        );
-        if (channelsByProfileId[profileId].storyGroups) {
-          bindProfileStoryGroupEvents(
-            channelsByProfileId[profileId].storyGroups,
-            profileId,
-            dispatch
-          );
+    switch (action.type) {
+      case 'APP_INIT':
+        dispatch(dataFetchActions.fetch({ name: 'getGlobalOrganizationId' }));
```
karelm, line 208: @hamstu +/ @albennett is GlobalOrganizationId different from GlobalAccountId? If not, then we already make the request to 'globalAccount' which has the core global org id, and would save an extra request. Just FYI :)

hamstu: It is indeed different @karelm. See https://github.com/bufferapp/buffer-web/pull/17001#pullrequestreview-391033048

And since the logic on the backend to get the 'main' organization is quite complex, we need this new API call, since even modifying the account RPC request to get the user's org, could in theory, return the wrong global org ID.

hamstu: Get main org logic: https://github.com/bufferapp/buffer-web/blob/fc8404b579c5fdb37ffb260f68b9adf848fc9176/shared/models/user_model.php#L4364

karelm: Yeah, was just looking into this myself... lots of fun to be had I guess 😂 Thanks for clarifying Hamish 🙌

hamstu: Yeah, all the global stuff is enough to make your head spin! We don't even store the old publish org ID on a campaign as I discovered – just the new global account one, probably to make things easy for cross-product sharing. So it felt right to broadcast this on the global org channel.

I do foresee this being a tricky thing in the future as we continue to migrate. So good for us to all understand :)

albennett: I asked the same question @karelm 😂

PR #1317, opened 2020-04-08
[PUB-2657] Tracking for Campaign Post Added

chayjay:
## Description

Add Segment tracking for when a user adds a post to a campaign.

## Context & Notes
PUB-2657

[https://share.buffer.com/v1uDoDld](https://share.buffer.com/v1uDoDld)
## Checklist

#### Staging Deployment

### DIFF
--- a/packages/composer/composer/utils/TrackingUtils.js
+++ b/packages/composer/composer/utils/TrackingUtils.js
@@ -17,6 +17,9 @@ const getComposerSource = ({ tabId, emptySlotMode }) => {
     case 'analytics':
       sourceName = 'share_again';
       break;
+    case 'campaigns':
+      sourceName = 'campaigns';
+      break;
     default:
       sourceName = tabId;
       break;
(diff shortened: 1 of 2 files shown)


packages/composer/composer/utils/TrackingUtils.js
```diff
@@ -17,6 +17,9 @@ const getComposerSource = ({ tabId, emptySlotMode }) => {
     case 'analytics':
       sourceName = 'share_again';
       break;
+    case 'campaigns':
```
albennett, line 20: Hey @chayjay when testing this in segment, it looks like it never gets to "campaigns" because we only use tabs from the routes with profileID (example: `https://publish.local.buffer.com/profile/5df289f7e65e3d00096d9865**/tab/queue**`. This means the `tabId` while in campaigns will always be the last `tabId` with the profile route. We need to find a way to detect if it's on the campaigns route. How does that sound to you?

packages/composer/composer/utils/WebAPIUtils.js
```diff
@@ -130,6 +130,23 @@ const WebAPIUtils = {
                     metadata,
                   },
                 });
+                if (post.campaign_details) {
+                  if (window.location.pathname.includes('campaigns')) {
```
albennett, line 134: Hey @chayjay , I think this approach works. 😄  I often use the `ownProps.history.location` when dealing with route logic, though it’s quite a lot of work to pass that into the composer.

  I find that adding `const` to logic can make it more readable. A suggestion could be something like: `const onCampaignsPage = window.location.pathname.includes( 'campaigns');`

PR #1325, opened 2020-04-14
Use optional chaining to avoid undefined sent counts

hamstu:
## Description

When post events come through Pusher we're attempting to update state that might not be defined, which is the `state.sent.byProfileId[...].total`. I added some optional chaining to default to `0` in that case.

## Context & Notes

I'm not actually sure if/where we use the total number of sent posts. 🤔  I know we use the 'pending' count in the sidebar.

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/queue/middleware.js
+++ b/packages/queue/middleware.js
@@ -213,8 +213,8 @@ export default ({ dispatch, getState }) => next => action => {
       const state = getState();
       const { profileId } = action;
       const currentCounts = {
-        pending: state.queue.byProfileId[profileId].total,
-        sent: state.sent.byProfileId[profileId].total,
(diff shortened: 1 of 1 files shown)


anafilipadealmeida:
🤔 I'm not sure we use it too @hamstu , I think it's probably not relevant. Do you have any idea why it started failing right now with the pusher events?

hamstu:
@anafilipadealmeida Yeah it's because before we only listened for these events once a profile was selected. (Which triggered fetches and thus populated the 'sent' store) But now we listen on the org channel which will catch events even when we've never selected that profile yet.

But yeah, still don't think we're actually using the sent count 😅

PR #1327, opened 2020-04-15
Use bufflog middleware INFRA-370

erickhun:
## Description
- Use [bufflog](https://www.notion.so/buffer/How-to-log-your-services-6e34e048d1954684a3f8cc33f71666dc) logger and middleware.  
- Originally, the request logs were sent in Kibana. Now the "success" requests are silented by default. I've made sure  the [API error logs](https://github.com/bufferapp/buffer-publish/pull/1327/files#diff-38b470ea439c2e777402cb2df425348fR5) in [Datadog logs](https://app.datadoghq.com/logs). 

## Notes
If we want to see the all request logs, set the `LOG_LEVEL` to `INFO` in the variable environment

## Context & Notes
- We should start using datadog logs instead of kibana. 
- JIRA ticket: https://buffer.atlassian.net/jira/software/projects/INFRA/boards/91?selectedIssue=INFRA-370

## Staging Deployment
https://task-use-bufflog-publish.dev.buffer.com

## Review
@hamstu up for a quick review on that one? It's a small change I don't think will impact any features.  The tests pass locally with `npm test`, and staging looks good. Hope that ok for you! LMK if I might miss something


hamstu:
@erickhun Instead of `npm install` can you add the dependency to the `package.json` manually?

After that you should be able to run `yarn` in the buffer-publish folder to install everything correctly. The short version of why we do that is because we don't use npm in Publish, but instead use yarn with its workspaces feature.

packages/server/middleware.js
```diff
@@ -1,8 +1,8 @@
+const bufflog = require('@bufferapp/bufflog');
+
 module.exports.apiError = (err, req, res, next) => {
   // eslint-disable-line no-unused-vars
-  console.log('-----');
-  console.error(err);
-  console.log('-----');
+  bufflog.error('Publish API error: '.err.message, err);
```
hamstu, line 5: I think there's a typo here, you probably want this:

```suggestion
  bufflog.error(`Publish API error: ${err.message}`, err);
```

erickhun:
Thanks @hamstu , and learnt something with template literals !

hamstu:
So glad to help @erickhun! Thanks again for doing this.

PR #1328, opened 2020-04-15
Remove buttons for team members campaigns

josem:
## Description
Hide actions in campaigns for team members, we use the same logic for hiding reports but in this case we want to hide the option to perform actions like creating a new campaign, and editing and removing existing ones.

## Context & Notes
- https://buffer.atlassian.net/browse/PUB-2669

image

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/campaign/components/ViewCampaign/Header/index.jsx
+++ b/packages/campaign/components/ViewCampaign/Header/index.jsx
@@ -46,12 +46,9 @@ const Header = ({
         label={translations.createPost}
         disabled={isLoading}
         displaySkeleton={isLoading}
-        onSelectClick={selectedItem => {
-          if (typeof selectedItem.selectedItemClick !== 'undefined') {
-            selectedItem.selectedItemClick();
-          }
-          return false;
-        }}
+        onSelectClick={selectedItem =>
+          selectedItem.selectedItemClick(campaignDetails)
+        }
         items={[
           {
             title: translations.createPost,
@@ -59,13 +56,11 @@ const Header = ({
           },
           {
             title: translations.editCampaign,
-            selectedItemClick: () => onEditCampaignClick(campaignDetails.id),
+            selectedItemClick: campaign => onEditCampaignClick(campaign.id),
           },
           {
             title: translations.deleteCampaign,
-            selectedItemClick: () => {
-              onDeleteCampaignClick(campaignDetails);
-            },
+            selectedItemClick: campaign => onDeleteCampaignClick(campaign),
           },
         ]}
       />
--- a/packages/campaign/components/ViewCampaign/Header/index.jsx
+++ b/packages/campaign/components/ViewCampaign/Header/index.jsx
@@ -34,15 +34,73 @@ const Header = ({
   onEditCampaignClick,
   goToAnalyzeReport,
   isLoading,
-}) => {
-  let buttonActionCampaigns;
-
-  if (showCampaignActions) {
-    buttonActionCampaigns = (
+}) => (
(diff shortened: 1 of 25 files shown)


anafilipadealmeida:
Hey @josem, I know this is a draft but I'm curious about why we are choosing to updated property name to reflect broader usage. We had decided to remove the business logic from ui components, that's why we have things like `hideAnalyzeReport`. If we change the business conditions in which we hide the button, instead of changing a bunch of ui components, we only need to change the index.js

josem:
Thanks so much for your comment @anafilipadealmeida, I really appreciate it. 

So my intention was to reflect that we're not only hiding the analyze report button now but also hiding the "create campaign" and also the options to edit and delete campaigns in the list. 

I couldn't come up with a good name for all those cases so that's why I change the variable name but reading your comment I understand now that perhaps a better way would be to have another property called: "hideCampaignManageButtons" or something similar even if the source of information is the same (```state.appSidebar.user.isUsingPublishAsTeamMember```). Would that be more aligned with that vision? 

Sorry for making a mistake there and open to any suggestions really! :)

anafilipadealmeida:
> Thanks so much for your comment @anafilipadealmeida, I really appreciate it.
> 
> So my intention was to reflect that we're not only hiding the analyze report button now but also hiding the "create campaign" and also the options to edit and delete campaigns in the list.
> 
> I couldn't come up with a good name for all those cases so that's why I change the variable name but reading your comment I understand now that perhaps a better way would be to have another property called: "hideCampaignManageButtons" or something similar even if the source of information is the same (`state.appSidebar.user.isUsingPublishAsTeamMember`). Would that be more aligned with that vision?
> 
> Sorry for making a mistake there and open to any suggestions really! :)

No need to apologize, it's not a mistake at all 😄And we are still mixing business logic and ui in most of Publish 😅I'll probably write something about this, as it's not documented anywhere! Something like you suggested, or hideCampaignButtons, or hideCampaignActions, seems good to me 😂

packages/campaign/components/ViewCampaign/Header/index.jsx
```diff
@@ -121,22 +64,100 @@ const Header = ({
...
+              </Details>
+              <Text type="p">
+                <LastUpdated>{campaignDetails.lastUpdated}</LastUpdated>
+              </Text>
+            </Fragment>
+          )}
+        </SubText>
+      </CampaignDetails>
+      <ButtonWrapper>
+        {buttonActionCampaigns}
```
anafilipadealmeida, line 143: Since this button has the same View Campaign default action, and the difference is that in one case we show the list with other campaign actions, another way of doing this would be to use the showCampaignActions prop to determine the isSplit: 
```
<ButtonWithSkeleton
          onClick={onCreatePostClick}
          type="secondary"
          isSplit={showCampaignActions}
          label={translations.createPost}
          disabled={isLoading}
          displaySkeleton={isLoading}
          onSelectClick={selectedItem =>
            selectedItem.selectedItemClick(campaignDetails)
          }
          items={[
            {
              title: translations.createPost,
              selectedItemClick: onCreatePostClick,
            },
            {
              title: translations.editCampaign,
              selectedItemClick: campaign => onEditCampaignClick(campaign.id),
            },
            {
              title: translations.deleteCampaign,
              selectedItemClick: campaign => onDeleteCampaignClick(campaign),
            },
          ]}
        />
```
That would remove the need of having a ButtonActionCampaigns variable and if conditions. 
If we keep the 2 button variations inside the variable, it's probably worth TitleCase its name, as it's returning a react component 🙂

PR #1334, opened 2020-04-17
[PUB-2646] Improve loading experience in Campaigns List

anafilipadealmeida:
## Description
- Add AnimatedList shared-component
- Use AnimatedList in both SkeletonPosts (ViewCampaign) and SkeletonList (ListCampaigns)
- Add Skeleton variation to campaigns List Item
- Transform ListCampaigns loading into skeleton

## Context & Notes
https://buffer.atlassian.net/browse/PUB-2646
One thing I've noticed when building the skeleton variations here is that we have to end up adding more complexity than we would if the feature was built with skeleton/loading state in mind from the beginning. It's a good learning for future features! 

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/shared-components/AnimatedList/index.jsx
+++ b/packages/shared-components/AnimatedList/index.jsx
@@ -21,10 +21,10 @@ const animation = ({ numberItems }) =>
 
 const createCSS = ({ numberItems }) => {
   let styles = '';
-  for (let i = 1; i < numberItems + 1; i += 1) {
+  for (let i = 0; i < numberItems + 1; i += 1) {
     styles += `
      > :nth-child(${i}){
-        animation-delay: ${0.5 * (i - 1)}s  ; 
+        animation-delay: ${0.5 * i}s  ; 
      }
    `;
   }
--- a/packages/shared-components/AnimatedList/index.jsx
+++ b/packages/shared-components/AnimatedList/index.jsx
@@ -42,12 +42,12 @@ const StyledList = styled.span`
   ${createCSS}
 `;
 
-const AnimatedList = ({ numberItems, children }) => (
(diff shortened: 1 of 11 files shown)


packages/shared-components/AnimatedList/index.jsx
```diff
@@ -0,0 +1,54 @@
...
+  > * {
+    animation: ${animation};
+    opacity: 0;
+  }
+  ${createCSS}
+`;
+
+const AnimatedList = ({ numberItems, children }) => (
+  <StyledList numberItems={numberItems}>{children}</StyledList>
+);
```
hamstu, line 47: Instead of requiring the number of children explicitly, you could ask React to figure that out for you:

```jsx
const numberItems = React.Children.count(children);
```
Ref: https://reactjs.org/docs/react-api.html#reactchildrencount

Saves having to write `numberItems={x}` everywhere :)

anafilipadealmeida: Why didn't I know about this??

anafilipadealmeida:
> This is so awesome @anafilipadealmeida! Left one small suggestion, but other than that this is looking solid.
> 
> One other Q though; when trying to click a single campaign on your branch I'm seeing this error:
> image
> 
> It doesn't happen on prod, but I'm guessing it's a dev-api issue and not related to your PR.

I'm not being able to reproduce it, but I agree it can be the dev-api, we've been seeing some issues in staging because of that. But I'll double check on Monday, once I also solve the branch conflicts 😂

packages/shared-components/AnimatedList/index.jsx
hamstu: Just one of the goodies in React's toolbox 😄

PR #1343, opened 2020-04-21
Refactor Publish Server

hamstu:
## Description

Just moving some things around and out of the server code to clean things up. Pre for some work to come. No functional changes. (Only real change is removing Intercom as per the linked Slack thread.)


packages/server/index.html
```diff
@@ -201,7 +201,6 @@
   {{{fullStoryScript}}}
   {{{notificationScript}}}
   {{{showModalScript}}}
-  {{{intercomScript}}}
```
hamstu, line 204: Removing Intercom (See: https://buffer.slack.com/archives/C151RRDL1/p1587491560188200?thread_ts=1587376395.103700&cid=C151RRDL1)

albennett: I think we reference intercom in a few different packages. Will removing it here cause any errors?

hamstu: Good one! I'll clean those up too. 👍

packages/server/lib/embeds/bufferData.js
```diff
@@ -0,0 +1,23 @@
...
+  }
+  if (typeof profiles !== 'undefined') {
+    bufferData.profiles = profiles;
+  }
+
+  return `
+<script>
+  try {
+    window.bufferData = ${JSON.stringify(bufferData)};
+  } catch(e) {}
```
albennett, line 19: Do you think we should log the error for debugging purpose?

hamstu: Yeah, I can add that with the new Bufflogger from Eric! (This was just copied over from the old code, but I like the suggestion.)

packages/server/index.html
hamstu: @albennett Done! ✅

packages/server/lib/embeds/bufferData.js
albennett: Awesome 🙌

PR #1345, opened 2020-04-21
✨ Add standalone mode (look ma' – no Docker!)

hamstu:
## Description

In standalone mode Publish runs directly on your machine, no buffer-dev or Docker containers required. User [authentication is bypassed by statically loading a session](https://github.com/bufferapp/buffer-publish/blob/04c52a7e96c4c59314a19ac8681f40d013532ec3/packages/server/index.js#L96-L107), and all requests are sent directly to the production dev API (`dev-api`).

## Why do we need this?

Right now the primary purpose is to support easy end-to-end testing in CI (Continuous Integration, like in GitHub Actions). When we can run the entire Publish app without other development / local dependencies then it makes simple end-to-end testing with packages like Cypress really easy.

In addition, because of how quick this is to boot up it's also a great way to quickly fix or work on a Publish-only task without the weight or slowness of Docker (this is most noticable on macOS,), or as a way to debug an issue occurring with production services while keeping the React code local.

[Continue reading in README...](https://github.com/bufferapp/buffer-publish/blob/standalone-mode/README.md#standalone-mode)

image

## Checklist

#### Staging Deployment


packages/server/lib/pusher.js
```diff
@@ -1,13 +1,13 @@
...
-});
-
 module.exports = (req, res) => {
+  const pusher = new Pusher({
+    appId: process.env.PUSHER_APP_ID,
+    key: process.env.PUSHER_APP_KEY,
+    secret: process.env.PUSHER_APP_SECRET,
+    cluster: process.env.PUSHER_CLUSTER,
+  });
+
```
karelm, line 10: how come the definition of pusher is being moved into the module.export? I think it was declared outside this scope so that there was only 1 pucher worker running and synchronizing data, but if theres a reason for us to move it thats cool. Just wanted to check

hamstu: Oh good catch, I moved this here so it runs later (after `standalone.loadEnv()`) so we have the env variables we need when running in standalone mode. But I was going to come back and try to refactor the import order so that I didn't need to do that. Will fix it so we're only creating one instance per-server. 👍

hamstu: Fixed this ✅

karelm: image

PR #1360, opened 2020-04-23
[PUB-2243] Add labels to all forms

chayjay:
## Description

Add labels to all forms for increased accessibility in New Publish.

## Context & Notes
PUB-2243

## Checklist

#### Staging Deployment


packages/general-settings/components/GoogleAnalytics/index.jsx
```diff
@@ -154,6 +155,7 @@ const GoogleAnalytics = ({
                 }}
                 name={'utmSource'}
                 placeholder={'bufferapp.com'}
+                ariaLabel='Campaign Source'
```
anafilipadealmeida, line 158: Hey @chayjay, isn't your eslint complaining about the quotes? I remember us using double quotes in this kind of situation 🤔

chayjay: Ah, yes, you're totally right. Apologies for this one, I fixed those linting errors along with the errors caused by the curly braces in this file!

packages/general-settings/components/GoogleAnalytics/index.jsx
```diff
@@ -152,8 +153,9 @@ const GoogleAnalytics = ({
                   value: utmSource,
                   onChange: onChangeUtmSource,
                 }}
-                name={'utmSource'}
-                placeholder={'bufferapp.com'}
+                name="utmSource"
+                placeholder="bufferapp.com"
+                ariaLabel="Campaign Source"
```
anafilipadealmeida, line 158: @chayjay, are you being able to see this aria-label in the DOM tree or listening to it with the voiceOver? I can't seem to find it in neither. You have a comment in the JIRA card for something else ("I have not added a label to those forms because another prop needs to be added in the shared components library for the Input component. Adding a new “aria-label” prop should be all we need to add labels to those forms." ). However this is the same Input component from bufferapp/components...

chayjay: Hey @anafilipadealmeida, good one here. I currently have a change in [this PR](https://github.com/bufferapp/buffer-components/pull/159) to pull together all of these changes! Next steps I believe will be to publish the buffer-components change to npm and then bump the version number in New Publish. I hope that makes sense!

anafilipadealmeida: Makes sense, I'm just curious why we are planning on making the change to our old components repository instead of migrating this inputs to the BDS ones?

chayjay: Ah, gotcha. It sounds like the change that actually needs to be made is to change the BDS Input component to accept a prop for aria label and then change this component to use the BDS Input. I'll discard these changes for now and add a new card for these changes!

packages/composer-user-tags/components/TagInput/index.jsx
```diff
@@ -76,6 +76,7 @@ const TagInput = (
         disabled={inputDisabled}
         help={error && translations.usernameError}
         hasError={error}
+        aria-label="Tag Input"
```
anafilipadealmeida, line 79: Hey @chayjay and @albennett, I don't know what your thoughts are on the aria-label usage, but in places like this, where the input has a visible label, it seems a better practice is to use aria-labelledby https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute. 

Nonetheless I don't think this aria-label is being displayed here, I can't find it in the DOM tree. However in this case I don't even think we needed one. aria-label and aria-labelledby are used when you don't use a html form structure (create the elements with div for instance). In this case we already have a label for="" that is linking to an input. What we are missing here is the id in the input to link to the label

anafilipadealmeida: You can read about the labels here: the https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label. If the input is nested in the label then you don't need the id, which is not the case. I can bring this to be discussed in the BDS meeting today

PR #1367, opened 2020-04-24
[PUB-2342] Cypress E2E Tests

hamstu:
## Description

* Integrate our first Cypress test! 🎉 
* It creates a post and makes sure it appears. (Check out the comment by the Cypress app bot below!)

## To do

- [x] Create mock user / profiles for each test run

## Checklist

#### Staging Deployment


cypress/support/login.js
```diff
@@ -0,0 +1,77 @@
+const email = Cypress.env('PUBLISH_LOGIN_EMAIL') || 'admin@bufferapp.com';
+const password = Cypress.env('PUBLISH_LOGIN_PASSWORD') || 'password';
```
anafilipadealmeida, line 2: @hamstu is there any security implication in exposing this details or is it all good?

package.json
```diff
@@ -13,12 +13,14 @@
...
     "start": "cd ./packages/server/ && yarn run start",
     "start:standalone": "cd ./packages/server/ && yarn run start:standalone",
+    "start:standalone-ci": "cd ./packages/server/ && yarn run start:standalone-ci",
     "storybook": "start-storybook -p 9001",
     "watch": "webpack-dev-server --config ./packages/server/webpack.config.dev.js --colors --display-error-details --progress",
     "build": "webpack --config ./packages/server/webpack.config.prod.js --output-path=`pwd`",
+    "build:ci": "USE_PRECOMPILED_BUNDLES=true webpack --config ./packages/server/webpack.config.prod.js --output-path=dist",
     "build:analyze": "ANALYZE_BUNDLE=true webpack --config ./packages/server/webpack.config.prod.js --output-path=dist --progress",
     "bugsnag:release": "node ./packages/server/scripts/bugsnag-release.js",
-    "rebuild-grpc": "docker exec -it bufferdev_publish_1 sh -c 'cd /usr/src/app && npm rebuild grpc --target_arch=x64 --target_platform=linux --target_libc=musl --update-binary --verbose'"
```
anafilipadealmeida, line 21: 😍

package.json
```diff
@@ -29,7 +31,7 @@
     "url": "https://github.com/bufferapp/project-donut/issues"
   },
   "homepage": "https://github.com/bufferapp/project-donut#readme",
-  "devDependencies": {
+  "dependencies": {
```
anafilipadealmeida, line 34: Hey @hamstu, the previous list had some dependencies as devDependencies 😅  But aren't many of these considered devDependencies as well, like storybook, jest, eslint, enzyme, babel, webpack?

cypress/support/login.js
hamstu: Should be fine! It's only ever applicable to local. Never a production anv. :)

package.json
hamstu: Just a bit of 🧹  😄

package.json
hamstu: Yeah, I made this change since I added `inquirer` as as dep for a script, but since it isn't needed for bundling the code/app I felt like moving everything to to actual 'dependencies' made sense. 

There's been some discussions around this in the NPM community. At the end of the day it doesn't matter until [we start using arguments in our yarn commands to exclude devDeps](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-production-true-false), which is a change I would like to make as a follow up to this.

hamstu: Created a Jira task to track that here: https://buffer.atlassian.net/browse/PUB-2717

anafilipadealmeida: Great @hamstu, thanks for clarifying ! I'm happy with them marked as dependencies, knowing we have another card for tackling the dev dependencies and take the most of out of our builds 😄

cypress/integration/publish.js
```diff
@@ -0,0 +1,23 @@
+const getRandomPostText = () => {
```
MayaUribe, line 1: Hey @hamstu! 🙂 Just a small comment here, because we’re using Cypress mainly for our e2e tests and we can do integration tests with jest and Enzyme / React Testing Library (hopefully 😄), I think it'd be a good one renaming this folder to `e2e` instead of `integration`, to communicate that better, and we can set the integration folder here with this option: `'integrationFolder: 'cypress/e2e'`, in the cypress.json file, What do you think? 😄

cypress/support/index.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Import our custom Cypress commands
+ */
+import './login';
+
```
MayaUribe, line 5: Hey @hamstu! me selling React Testing Library again 😂 a really cool feature about React Testing Library is that it has a plugin we can use for Cypress, by adding `import '@testing-library/cypress/add-commands'` to this file, we can use all the asynchronous selectors from RTL like `findByText` and `findByLabelText`, this prevents having to add a `data-cy` tag for each component we want to access for our e2e tests unless is really necessary (in which case we could use a `data-testid` tag and access it with the `findByTestId`), which makes it easier to test and read 🙂

PS: This is just something I wanted to share 😊

cypress/support/login.js
```diff
@@ -0,0 +1,77 @@
...
+  cy.task('getAccessToken').then(accessToken => {
+    const at = Cypress.env('API_AT') || accessToken; // use env specified token if present
+    cy.fixture('pro-user').then(proUser => {
+      const randomEmail = `cypress-e2e-test-user-${getRandomText()}@buffer.com`;
+      console.log('Creating user with email', randomEmail);
+      const user = {
+        password: getRandomText(),
+        email: randomEmail,
+        ...proUser,
+      };
```
MayaUribe, line 61: Hey @hamstu great work on everything you did here, I'm loving going through this PR 😍
Just a small comment here, maybe we could move this logic for creating a new user to a different file and import it here, like a support/generate.js I think separating this logic would make it a bit easier to read. What do you think? 🙂

MayaUribe: One other thing I wanted to share that I think is kind of cool 😂

This isn’t necessary at all, just wanted to share 😄 there’s this cool package to handle these type of fake data, but maybe is not worth it, because we only need to fake the email and password 🤔 is called `test-data-bot` and it would look something like this:

```
// support/generate.js
const { build, fake, sequence } = require('test-data-bot’);

const userBuilder = build('User').fields({
  password: fake(f => f.internet.password()),
  email: sequence(x => `cypress-e2e-test-user-${x}@buffer.com`),
})

export {userBuilder};
```

We could fake all type of things with that library if we needed to 🙊

PR #1374, opened 2020-04-27
Guidelines | Working with i18n

MayaUribe:
## Description

Update guidelines to default to React.i18n when handling strings in our codebase.

## Context & Notes

This came out after the work for engineering week, since there is a lot of text remaining to move to the translations files, we decided to be more specific on how to work with this and continue this work in a steady way.

## Checklist

#### Staging Deployment

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -180,6 +180,11 @@ const Welcome = ({ name = 'John Smith' }) =>
 ### Strings and i18n
 For i18n, at the moment we have an internal package called `@bufferapp/publish-i18n`, however we are migrating all our String handling to `React.i18n`, so whenever you are working on a component, please defaut to [`React.i18n`](https://react.i18next.com/).
 
+**Adding strings to the translations JSON files:**
+- When adding a new string to the translations/*.json files, make sure to always add them in both files: `en-us.json` and `es-es.json` file, is completely fine if the text is in english for the Spanish json, José, Karël or Maya will make sure to translate it eventually 🙂, but this is mainly to keep consistency among all translation files.
+
+**Some Examples on how to work with React.i18n**
+
 - **Example with Hooks:** Since most of our components are functional components, the more often way to work with translations is with the Hook `useTranslation`:
 
 ```js
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,7 @@ Here are some guidelines to make it as easy and clear as possible.
 - [Coding Styleguide 💻](#coding-styleguide-💻)
   - [Prettier 💁‍♀️](#prettier-💁‍♀️)
   - [Components Styleguide](#components-styleguide)
+  - [Strings and i18n](#strings-and-i18n)
 - [Adding New Dependencies](#adding-new-dependencies)
 - [How Packages Communicate 📦](#how-packages-communicate-📦)
 - [Styling Styleguide 💅](#styling-styleguide-💅)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -176,6 +177,53 @@ const Welcome = ({ name = 'John Smith' }) =>
    <h1>Hello, {name}</h1>;
 ```
 
+### Strings and i18n
+For i18n, at the moment we have an internal package called `@bufferapp/publish-i18n`, however we are migrating all our String handling to `React.i18n`, so whenever you are working on a component, please defaut to [`React.i18n`](https://react.i18next.com/).
+
+**Adding strings to the translations JSON files:**
+- When adding a new string to the translations/*.json files, make sure to always add them in both files: `en-us.json` and `es-es.json` file, is completely fine if the text is in english for the Spanish json, José, Karël or Maya will make sure to translate it eventually 🙂, but this is mainly to keep consistency among all translation files.
```
karelm, line 184: we could use the tool that the android/ios team use to create these translation files for us too @MayaUribe https://crowdin.com/profile? just a little suggestion for things we could try, that way we open up the possibility of translations to the rest of the team too 😄

karelm: I've created a private project here: https://crowdin.com/translate/buffer-publish/6/en-es I can add you

josem: Yes I think that's a good idea although Joe told me that we don't have the budget for that now since it gets a bit expensive but perhaps is something we can explore in the future.

karelm: Good to know... just saw the pricing and it definitely is expensive 😅

diannemcewan: Hey @josem @karelm!   

I am a little worried about us adding this extra load onto the team of needing to take care of an extra json file for Spanish, when we don't yet know that this is something we will be rolling out to our users. 

 If we don't continue to update the 'es-es.json` when adding new copy will anything break? If/when we decide to go live with these translations can we not backfill this copy then?

josem: Not really since our users won't be using Spanish and this is behind a feature flip so it might be ok too.

If the mobile team uses the endpoint to set the language for the API at some point, we might want to change this project so we also check the feature flip before showing the UI in Spanish.

anafilipadealmeida: I agree with @diannemcewan on this one, I'd prefer the KISS approach here, instead of adding extra work keeping the es-es.json file if this is not gonna be displayed to users soon. Also, as @MayaUribe mentioned yesterday, if we'd like to deeply restructure the english file to reuse some strings globally, having to do it in 2 files is gonna make things slow. I'd suggest us no to manually update the Spanish file, if you really want to match it perhaps a script that would do it automatically for us?

karelm: +1 here, I think it would be good to add a note to this readme that says... 
> If we start to adopt i18n more widely in the project, we should make the consious decision to take the following steps. For the time being, please add them to the `en-us.json` file until we make a further decision on this

MayaUribe: > If we don't continue to update the 'es-es.json` when adding new copy will anything break? If/when we decide to go live with these translations can we not backfill this copy then?

Even if the user somehow manages to change the language to Spanish and the translation is not in the file, no it doesn't break anything, it would just display the key for that translation, so if we have something like `t('common.pleaseWait')` and we didn't add that key to es-es.json file, the user would see `common.pleaseWait`, not great, but it wouldn't break it, but in case it's under a feature flip, so nobody will really see it at all.

MayaUribe: I think for the time being we could focus on having a better string structure, and once we have that in place, we can look to replicate it to the es-es.json or any other translation file, but my main point here would be for us to default to adding all strings in the en-us.json file, I'll make the changes for this to be more clear and I'll also add that note you mentioned @karelm 👌

> If we start to adopt i18n more widely in the project, we should make the conscious decision to take the following steps. For the time being, please add them to the en-us.json file until we make a further decision on this

cc: @anafilipadealmeida @josem @karelm @diannemcewan

PR #1390, opened 2020-05-11
Proposal | Use React Testing Library for components and integration tests

MayaUribe:
## Description
This PR is a proposal to start using React Testing Library to test components and integration tests.

### Key benefits for Testing React Library (RTL) 🐙
- No setup required
- Easy and intuitive to start testing
- Helps to avoid implementation details
- Enforces better code practices
- Accessible by default
- Lightweight ([bundle comparison](https://share.buffer.com/rRu9gEGB))

### Changing our mindset when testing 💭
When testing components, is important for us to test in terms of how the software is used rather than testing pieces of code:

> The more your tests resemble the way your software is used, the more confidence they can give you. Kent C. Dodds

Implementation details are important because they impact the behavior of our UI, they can result in DOM changes, state updates, changes in props... **The takeaway here is to test how our UI reacts to those changes, rather than the changes themselves**, and React Testing Library makes it really easy 🙌

### Using our actual Store to test
This test uses our real Redux store instead of a mock-store, Using a real Redux store effectively closes the loop between UI event (input) and UI update (output). In other words, you can fire an event (click, keypress, etc.) then assert that the UI is updated to reflect some new state. With a mock store, you can only assert that actions are dispatched.

Why does this matter? Since this test isn’t concerned with how the UI is updated, you can easily refactor without breaking the test. The elements inside the connected component (its reducer, action creator, etc.) are just implementation details. ([ref](https://peterhrynkow.com/testing/2019/01/01/testing-atoms-and-molecules.html))

### When testing, we can ask ourselves these questions ❓
In our App, there are actions that result in UI changes, there are different types of actions that can take place, so what happens in our app when these actions take place? (from Kent C. Dodds [blog](https://kentcdodds.com/blog/how-to-know-what-to-test)):
- User interactions (using [`fireEvent`](https://testing-library.com/docs/dom-testing-library/api-events#fireevent) from RTL): Is the end-user able to interact with the elements that the component renders?
- Changing props (using [`rerender`](https://testing-library.com/docs/marko-testing-library/api#rerender) from RTL): What happens when the developer user re-renders your component with new props?
- Context changes (using [`rerender`](https://testing-library.com/docs/marko-testing-library/api#rerender) from RTL): What happens when the developer user changes context resulting in your component re-rendering?
- Subscription changes: What happens when an event emitter the component subscribes to changes? (Like firebase, a redux store, a router, a media query, or DOM-based subscriptions like online status)

## Checklist

#### Staging Deployment


packages/utils/test-utils.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
```
MayaUribe, line 1: This is a util test file that helps build our tests easier by wrapping the component with the store and history, so when using `render(<ComponentToTest />)`, we don't need to mock the store every time, and it will work as a production app 👌

.eslintrc.json
```diff
@@ -21,6 +21,7 @@
     "no-underscore-dangle": 0,
     "space-before-function-paren": 0,
     "import/no-extraneous-dependencies": 0,
-    "jsx-a11y/href-no-hash": [0]
+    "jsx-a11y/href-no-hash": [0],
+    "react/jsx-filename-extension": [1, { "extensions": [".js", ".jsx"] }]
```
MayaUribe, line 25: This one is to remove the warning from eslint when creating a component from a .js file.

setupTest.js
```diff
@@ -1,5 +1,10 @@
 import 'jest-styled-components';
+import 'jest-axe/extend-expect';
+import '@testing-library/jest-dom/extend-expect';
+import '@bufferapp/publish-web/components/i18n';
```
MayaUribe, line 4: Importing the i18n file here so we don't have to import it in every test.

packages/campaign-form/index.test.js
```diff
@@ -0,0 +1,163 @@
+import React from 'react';
+import userEvent from '@testing-library/user-event';
+import { axe } from 'jest-axe';
+import { green, purple } from '@bufferapp/ui/style/colors';
+import {
+  render,
+  screen,
+  waitFor,
+} from '@bufferapp/publish-test-utils/test-utils';
```
MayaUribe, line 9: Using the utils file instead of @testing-library/react, since we added the wrapper and re-exported everything.

packages/campaign-form/index.test.js
```diff
@@ -0,0 +1,163 @@
...
+    cancelButton,
+  };
+};
+
+const initialState = {
+  appSidebar: { user: { features: ['campaigns'] } },
+};
+
+describe('CampaignForm | user interaction', () => {
+  test('entering values in create form enables save option', async () => {
```
MayaUribe, line 36: Example of how a user would use Create Campaign:
- The CampaignForm is rendered with the initialState
- The purple color should be selected by default
- The save button should be disabled
- The user types in "Campaign Test" as a campaign name
- The input should now have the text "Campaign Test"
- The user clicks on the "green" color
- The green color should now be selected
- The purple color should no longer be selected
- The save button is no longer disabled
- The user clicks on Save
- The create campaign endpoint was called with "green" and "Campaign Test", only once

packages/campaign-form/index.test.js
```diff
@@ -0,0 +1,163 @@
...
+    render(<CampaignForm />, {
+      initialState: {
+        appSidebar: { user: { features: [''] } },
+      },
+    });
+
+    expect(screen.queryByRole('heading')).toBeNull();
+  });
+
+  test('a11y | campaign form is accessible', async () => {
```
MayaUribe, line 144: This test is a basic example of how to make sure our component is accessible ✨

packages/campaign-form/index.test.js
```diff
@@ -0,0 +1,163 @@
...
+
+  test('user should access create form when having the feature flip', () => {
+    render(<CampaignForm />, {
+      initialState,
+    });
+
+    expect(screen.getByRole('heading')).toHaveTextContent(/create campaign/i);
+  });
+
+  test.skip('user should not be able to access campaigns without having the feature flip', async () => {
```
MayaUribe, line 134: I'm skipping this test because is currently failing due to a bug we have 👀

packages/campaign-form/index.test.js
```diff
@@ -0,0 +1,202 @@
...
+    const campaign = {
+      id: campaignId,
+      _id: campaignId,
+      name: 'Test Campaign',
+      color: green,
+      globalOrganizationId: '1',
+      channels: [],
+    };
+
+    jest
```
MayaUribe, line 90: Mock API response with a campaign.

packages/test/utils/custom-render.js
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
```
MayaUribe, line 1: This is a util test file that helps build our tests easier by wrapping the component with the store and history, so when using `render(<ComponentToTest />)`, we don't need to mock the store every time, and it will work as a production app 👌

MayaUribe:
> This looks really great Maya! LGTM 🚀
> 
> 
> 
> I'm curious how quickly do you think we could replace all the tests that use enzyme? (Not right now of course, in a future PR.) It would be awesome to reduce the number of packages we have required to install. 😄

That's a great one @hamstu!
We currently have **21 tests that use Enzyme**, I did a general review for all of them and for the vast majority we have one or two tests tops, and in many of those the test is `should render`, and 3 of them might be a bit more tricky, so I actually think it _should_ be a quick replacement, I can create a task in JIRA to do this replacement and get rid of Enzyme 😬

jest.config.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
   collectCoverage: true,
   collectCoverageFrom: [
     '**/packages/**/*.{js,jsx}',
+    '!**/packages/test/**',
     '!**/packages/**/story.jsx',
```
anafilipadealmeida, line 31: Since we are changing this file I'm wondering if we can add this as well 😅 
```suggestion
    '!**/packages/**/story.jsx',
    '!**/packages/**/story_test.jsx',
```

packages/campaign-form/index.test.js
```diff
@@ -0,0 +1,224 @@
...
+      name: 'Campaign Test',
+    });
+    expect(rpcCall).toHaveBeenCalledTimes(1);
+  });
+
+  test('entering values in update form enables save option', async () => {
+    const campaignId = '18027';
+    const campaign = {
+      id: campaignId,
+      _id: campaignId,
```
anafilipadealmeida, line 83: Is the _id relevant @MayaUribe ? I might be missing something, but it seems we don't use it in the frontend, so I'm wondering why we keep both `id` and `_id` 🤔

packages/campaign-form/index.test.js
```diff
@@ -0,0 +1,224 @@
...
+    expect(screen.queryByRole('heading')).toBeNull();
+  });
+
+  test('a11y | campaign form is accessible', async () => {
+    const { container } = render(<CampaignForm />);
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+
+  test('renders campaign form and navigates to from path on cancel', async () => {
```
anafilipadealmeida, line 211: Is this test checking if "returns to campaigns when click cancel inside the form"? I'm a bit confused with the current test name 😅

packages/campaign-form/index.test.js
MayaUribe: Yap! you are right! I'll remove it 😊

packages/campaign-form/index.test.js
MayaUribe: Oh my yes! 🤦🏻‍♀️ That name is all wrong now that I re-read it 😄 it is checking that it returns to campaigns when clicking on cancel 😬 I'll change it, thank you for the suggestion 😊

MayaUribe:
> Also, do you know how testing-library performs? Would it be fine to add tests to all packages, or do you think the time running tests would increase a lot?

That's a good one @anafilipadealmeida 😊 for what I've read so far, this library performs really well overall, however, there are a couple of things to consider when testing: 
* Prioritize what we need to test, it is not a great idea to test every component in our app.
* The way we structure our tests is also important, we should try to avoid over-testing because that can lead to having a lot of errors when in fact just one thing is failing and make our tests slower: if for instance, we test the happy path, those assertions shouldn’t be added for the rest of the tests, since we already covered those.

I hope that makes sense 🤔😅

packages/campaign-form/index.test.js
karelm: whats the bug we have @MayaUribe ?

packages/campaign-form/index.test.js
```diff
@@ -0,0 +1,211 @@
...
+      },
+    });
+
+    expect(screen.queryByRole('heading')).toBeNull();
+  });
+
+  test('a11y | campaign form is accessible', async () => {
+    const { container } = render(<CampaignForm />);
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
```
karelm, line 198: 😍

packages/test/utils/custom-render.js
```diff
@@ -0,0 +1,64 @@
...
+  );
+
+  return {
+    ...rtlRender(ui, {
+      wrapper: Wrapper,
+      ...renderOptions,
+    }),
+    // adding `store` and `history` to the returned utilities to allow us
+    // to reference it in our tests (just try to avoid using
+    // this to test implementation details).
```
karelm, line 54: what do you mean by this @MayaUribe ? what would be a "bad" example here vs an allowed example?

anafilipadealmeida:
> > Also, do you know how testing-library performs? Would it be fine to add tests to all packages, or do you think the time running tests would increase a lot?
> 
> That's a good one @anafilipadealmeida 😊 for what I've read so far, this library performs really well overall, however, there are a couple of things to consider when testing:
> 
> * Prioritize what we need to test, it is not a great idea to test every component in our app.
> * The way we structure our tests is also important, we should try to avoid over-testing because that can lead to having a lot of errors when in fact just one thing is failing and make our tests slower: if for instance, we test the happy path, those assertions shouldn’t be added for the rest of the tests, since we already covered those.
> 
> I hope that makes sense 🤔😅

It does @MayaUribe ! I'd love to know how many more seconds this campaign-form test adds to the suite, so I can better understand. Because I can see this in the short run being useful to test hashtag manager, shopgrid  and other campaigns functionalities, but I guess the decision depends not only on the benefit of the tests but also on the time they will take to run 🙂

packages/test/utils/custom-render.js
MayaUribe: Great question @karel! 👌😊

**We should aim to test behavior over implementation details**, here is an example of what I mean.

Let’s assume I want to test that the form isLoading:

**Example with implementation details**
```
test('tests implementation details', () => {
  const { store } = render(<CampaignForm />);
  expect(store.getState().campaign.isLoading).toBe(true);
});
```
- For this example, we are asserting that the state has the value we need, however, the problem with this approach is that because is testing implementation details, if we, later on, decide to make a refactor and decide to use an enum instead of a boolean, this test will fail, and we will need to update it to reflect that change in implementation details (we are now using an enum instead of a boolean to check if the form is loading)

**Example with behavior**
```
test('tests behavior', () => {
  render(<CampaignForm />);
  expect(screen.queryByText(/loading/i)).toBeInTheDocument();
});
```
- Instead, with this approach, were are rather testing the behavior, which is to make sure the loading is displayed, *regardless of how that is implemented*, this test will continue to pass after our refactor from boolean to enum.

I hope that clarifies your question 😊

packages/campaign-form/index.test.js
MayaUribe: @karelm This is the bug that makes this test fail, I actually found it while creating the test 😅 -> [PUB-2727](https://buffer.atlassian.net/browse/PUB-2727)

packages/campaign-form/index.test.js
MayaUribe: @karelm This is the bug that makes this test fail, I actually found it while creating the test 😅 -> PUB-2727

MayaUribe:
> It does @MayaUribe ! I'd love to know how many more seconds this campaign-form test adds to the suite, so I can better understand. Because I can see this in the short run being useful to test hashtag manager, shopgrid and other campaigns functionalities, but I guess the decision depends not only on the benefit of the tests but also on the time they will take to run 🙂

I just merged with master to make sure to have the last version, and ran the tests in master and for this branch if that helps:

Image 2020-05-12 at 4 47 27 PM

Image 2020-05-12 at 4 57 21 PM

To run the test, I temporarily removed this `'!**/packages/**/story_test.jsx'`, to know how much is actually adding in time.

I'm honestly confused 😂 not sure why the time is lower 🤔

PR #1396, opened 2020-05-17
Migrate preferences routes logic to react-router

anafilipadealmeida:
## Description
- Use react-router to deal with preferences routes and tabs
- Replace tabs shared-components in preferences by new `<Nav>` and `<NavLink>`, to use the `<Link>` from react-router
- Update styled-components to v5.1.0  be able to use transient props (https://github.com/styled-components/styled-components/releases/tag/v5.1.0). According to the docs, there are no breaking changes in migrating from v.4 to v.5
- Update jest-styled-components so it works with v.5 styled-components

## Context & Notes
I was trying to understand how much work it would be to keep migrating the routes logic to react-router, and ended up doing it for preferences as a test 😅 
The idea is to replace the Profiles routes to use react router and the new nav components as well

## Checklist

#### Staging Deployment


packages/routes/index.js
```diff
@@ -68,6 +51,35 @@ export const getParams = ({ pathname, route }) => {
   return match?.params || null;
 };
 
+export const goTo = path => push(path);
+
+// Preferences routes
```
MayaUribe, line 56: 🙌

karelm:
I've notice a slight layout shift in the tabs... but it looks to be present in live too... if you switch between `Security` and `Apps & Extras` there is a subtle layout switch in the navigation bar. I don't know if there is anything we can do about that, but thought i'd mention it @anafilipadealmeida

anafilipadealmeida:
> I've notice a slight layout shift in the tabs... but it looks to be present in live too... if you switch between `Security` and `Apps & Extras` there is a subtle layout switch in the navigation bar. I don't know if there is anything we can do about that, but thought i'd mention it @anafilipadealmeida

Yes @karelm, there are two weird movements happening there: one is the content of preferences which moves between the other tabs and the notifications, because notifications page has a scroll. The other one seems to happen because a selected tab weight changes. Didn't address them here because it was getting out of scope 😅

PR #1399, opened 2020-05-26
Remove Edit button from RTs without comments

josem:
## Description
Since the behavior of editing a native retweet without a custom comment is to copy the content of the RT to the composer, this change disables that action to fix the issue.

## Context & Notes
Fix https://buffer.atlassian.net/browse/PUB-1217

## Screenshots

image

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/queue/components/QueueItems/index.jsx
+++ b/packages/queue/components/QueueItems/index.jsx
@@ -99,7 +99,9 @@ const renderPost = ({
     hasPushNotifications,
     onSetRemindersClick,
     hasCampaignsFeature,
+    shouldShowEditButton: post.retweet && !!post.retweetComment,
   };
+
   const PostComponent = Post;
 
   const defaultStyle = {
--- a/packages/shared-components/CardFooter/index.jsx
+++ b/packages/shared-components/CardFooter/index.jsx
@@ -116,6 +116,7 @@ const CardFooter = ({
   isDragging,
   disableBorder,
   hasUserTags,
+  shouldShowEditButton,
 }) => {
   const hideAllButtons = hideButtons || isPerformingAction || messageLink;
   const [isConfirmingDelete, setConfirmingDelete] = useState(false);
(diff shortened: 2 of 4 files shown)


josem:
Hey @MayaUribe! 

I'd love to have your eyes on this when you have some time and I have a couple of questions if that's ok. :slightly_smiling_face: 

I've been a little bit confused by having shared-components/QueueItems and also shared-components/queue/components/QueueItems. Do you know why we have two directories instead of one? :thinking:  It seems we only use the latter for the queue but I'm not totally sure. 

Another question I have is if the name of the property is ok. Not sure if I should pass something like isNativeRetweetWithoutComment or is it better to pass the UI implication of the business logic as I did? (shouldShowEditButton). 

Sorry for all the questions! Thanks so much! :raised_hands:

MayaUribe:
Hey there @josem! thank you for taking care of this one!

> I've been a little bit confused by having shared-components/QueueItems and also shared-components/queue/components/QueueItems. Do you know why we have two directories instead of one? 🤔 It seems we only use the latter for the queue but I'm not totally sure.

I know! it's confusing 😂, we actually have a task in the backlog to refactor this one 👌
At the moment we have two QueueItems:
- The one in `queue/QueueItems` is used only in the main queue tab
- The one in `shared-components` is used everywhere else (campaigns, drafts, etc)

I think the reason was that the main queue was more complicated, and we wanted a "simpler" approach for the rest of the queues, but is really not the best approach and we'll refactor it soon to have a single component 😊

> Another question I have is if the name of the property is ok. Not sure if I should pass something like isNativeRetweetWithoutComment or is it better to pass the UI implication of the business logic as I did? (shouldShowEditButton).

I find the `shouldShowEditButton` to be better, in the case that for some reason we want to hide the edit button for any other reason, the name is generic enough to leave it, if we specify is for a retweet we might need to change it later, I think is fine as it is 😊

> Sorry for all the questions! Thanks so much! 🙌

Thank _you_! 🙌

PR #1402, opened 2020-05-27
PUB-1687 | Adding better support for pagination, and retrieving records, as well as page for requests

karelm:
## Description

Adding better support for pagination, and retrieving records, as well as paging for requests

## Checklist

#### Staging Deployment


packages/profile-page/index.js
```diff
@@ -19,6 +19,11 @@ const requestName = tabId =>
     default: 'queuedPosts',
   }[tabId]);
 
+const getRequestMaxCount = tabId =>
+  ({
+    queue: 300,
```
albennett, line 24: During the bug smash call, a max count number of 500 was decided upon (it looks like Adam recently added a comment in the jira description).

packages/profile-page/index.js
```diff
@@ -36,20 +41,17 @@ export default hot(
...
+      if (state?.[reducerName]?.byProfileId?.[profileId]) {
+        const currentProfile = state[reducerName].byProfileId[profileId];
         return {
-          loading: state[reducerName].byProfileId[profileId].loading,
-          loadingMore: state[reducerName].byProfileId[profileId].loadingMore,
-          moreToLoad: state[reducerName].byProfileId[profileId].moreToLoad,
-          page: state[reducerName].byProfileId[profileId].page,
-          posts: state[reducerName].byProfileId[profileId].posts,
-          total: state[reducerName].byProfileId[profileId].total,
+          loading: currentProfile.loading,
```
albennett, line 47: Much cleaner 💯

packages/profile-page/index.js
```diff
@@ -36,20 +41,17 @@ export default hot(
           : tabId;
       if (tabId === 'awaitingApproval' || tabId === 'pendingApproval')
         reducerName = 'drafts';
-      if (
-        state[reducerName] &&
-        state[reducerName].byProfileId &&
-        state[reducerName].byProfileId[profileId]
-      ) {
+      if (state?.[reducerName]?.byProfileId?.[profileId]) {
```
albennett, line 44: Is optional chaining necessary for state? ex `state?.`

karelm: I added this because of the fact we are using the [reducerName] here, which gives me slight concern, otherwise I would have left it without the optional chaining at that level. If you think its better to remove that, then I'm happy to do it... I'm just wary of using the `[reducerName]` there

packages/profile-page/index.js
karelm: Thanks for raising that... I think that the 300 limit we are using is actually a reasonable amount for us to use, especially as that ends up downloading upto about 500Kb in updates in total 😅  (if a user has 300 updates). The issue we actually have is that theres a mismatch between how we fetch data, and display data (I've added a note to the ticket about that, and suggest we do another piece of work as follow up).

I will comment on the ticket about why 300 is likely appropriate for the time being, but I don't want to cause excess fetching in the UI, or under fetching either... but we are now using the count in the pagination logic, which should fetch 300 per page scroll vs the 120 limit being fetched before due to the backend limitation. So we should get a more consistent experience in the queue going forward, even if the limits and fetched data don't quite align.

albennett: That makes sense @karelm , thanks for leaving a comment explaining the decision for Anne & Adam 🙌

packages/profile-page/index.js
albennett: Ah, that makes sense. I agree, I think this case is fine to use optional chaining.

PR #1404, opened 2020-05-28
Add tests to the main queue

MayaUribe:
## Description

Add tests to the main queue and fake data builder.

## Context & Notes

I decided to add a couple of tests to the main Queue since we don't have any for the component itself at the moment.

## Checklist

#### Staging Deployment


packages/queue/index.test.js
```diff
@@ -0,0 +1,194 @@
...
+  publishProfiles: [profile],
+  user: { isOnBusinessTrial: false },
+};
+
+// eslint-disable-next-line react/prop-types
+const _TestContextContainer = ({ children }) => <>{children}</>;
+
+const queueComponents = () => {
+  const whatToShareInput = screen.getByRole('button', {
+    name: /what would you like to share?/i,
```
albennett, line 55: I wonder if we can lean towards grabbing text from translations. This way, the tests won't fail when we change copy in components.

packages/test/generate-data/index.js
```diff
@@ -0,0 +1,171 @@
+import { build, fake, sequence } from '@jackfranklin/test-data-bot';
+
+const schedules = [
```
albennett, line 3: I love that there's an easier way to generate data, as I've felt we've been duplicating code in tests across packages. 💯 Thanks for getting this started.

package.json
```diff
@@ -64,6 +64,7 @@
     "@bugsnag/js": "^6.5.2",
     "@bugsnag/plugin-react": "^6.5.0",
     "@hot-loader/react-dom": "16.9.0",
+    "@jackfranklin/test-data-bot": "^1.3.0",
```
albennett, line 67: Neat, I haven't heard of this package yet. If there's ever maintenance issues in the future, we could build out a data generator ourselves. 😄

packages/queue/index.test.js
MayaUribe: Hey, @albennett great question! 🙌 the reason why I query by the text directly is that we also want to make sure the translations are being applied correctly, I'm using the default locale which is en-US

I do get it can be a bit of a pain if we change the copy, we will need to change the test, but this is a quick way to make sure our translations component is working correctly 😊

More details on this approach [here](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-querying-by-text) 🙂

PR #1407, opened 2020-06-01
Load organizations user

josem:
## Description
Fetch organizations for the current logged in user. 

This is part of a bigger task where we'll be able to select different organizations so it "simulates" that we select one organization and filter the profiles by it but in reality, we just return all the profiles like before until the feature is completed. 

JIRA ticket: https://buffer.atlassian.net/browse/PUB-2779.

## Context & Notes
This is mostly based on @karelm's investigation :)  

Since https://github.com/bufferapp/buffer-web/pull/17098 is deployed I think we can safely deploy it this one as well once it's reviewed.

## Checklist

#### Staging Deployment


packages/profile-sidebar/utils/index.js
```diff
@@ -8,3 +8,10 @@ export const shouldGoToProfile = (profile, prevProps) => {
     (isADifferentProfile(profile, prevProps) || prevProps.profileId === null)
   );
 };
+
+export const filterProfilesByOrg = (profiles, organization) => {
+  return profiles;
```
josem, line 13: I return all the profiles for the time being since we're not really selecting any organization (and we shouldn't filter by the first one).

packages/bootstrap/middleware.js
```diff
@@ -1,14 +1,28 @@
...
+      dispatch({ type: 'INIT_MODALS' });
+      dispatch({ type: 'INIT_USER' });
+      dispatch({ type: 'INIT_ORGANIZATIONS' });
+      dispatch({ type: 'INIT_APPSHELL' });
+      dispatch({ type: 'INIT_FEATURES' });
+      dispatch({ type: 'INIT_PUSHER' });
+      dispatch({ type: 'INIT_CHECK_BOOKMARKLET' });
+      dispatch({ type: 'INIT_STRIPE_DETAILS' });
       break;
+    case `organizations_${dataFetchActions.FETCH_SUCCESS}`: {
```
karelm, line 16: instead of using the `organizations_${...}` approach, I was wondering if it would make more sense introducing an `ORGANIZATIONS_INITIALIZED` event instead inside of the organizations middleware? Then we can be purposeful in our actions, and it becomes clear of our actions and intentions? I'm not 100% sure, but just wondered what you thought to that instead of a fetch_success event here?

josem: Ah, great point! 

Sorry for the silly question since I don't have enough context on the project :pray: Do you mean capturing the fetch success event in the organizations' middleware and then dispatch a new event called ORGANIZATIONS_INITIALIZED that can be captured by others?

karelm: Sorry, I should have been clearer. Yes that is precisely what I mean. This way we arent using the "underlying" mechanisms, but instead allow each module to signal its state, instead of reading signalled state from http state. This way if we want to add more checks, filtering, etc... to our organization processing, we don't need to update all locations in the code relying on this.

josem: Thank you @karelm! That makes total sense :+1:

josem: I've just made the change. Please, let me know if you think it's fine now, happy to adjust anything! :slightly_smiling_face:

karelm: Hey @josem sorry about this, but you haven't updated this event to use the ORGANIZATIONS_INITIALIZED property, would you be able to update this one please?

packages/organizations/reducer.js
```diff
@@ -0,0 +1,26 @@
+const orgData =
+  typeof window !== 'undefined' &&
+  typeof window.bufferData !== 'undefined' &&
+  typeof window.bufferData.organizations !== 'undefined' &&
+  window.bufferData.organizations;
```
karelm, line 5: I think this can be simplified to just be something like this instead 😉 
```suggestion
const orgData = window?.bufferData?.organizations;
```

packages/organizations/utils/index.js
```diff
@@ -0,0 +1,11 @@
+export const getSelectedOrganization = organizations => {
```
karelm, line 1: Could you add some tests around this please? Happy to help with this if needed, I know a lot of this code came from my little spike in investigating so appreciate it might not be in the perfect state

packages/server/parsers/src/orgParser.js
```diff
@@ -0,0 +1,9 @@
+module.exports = orgData => ({
+  id: orgData.id,
+  locked: orgData.locked,
+  name: orgData.name,
+  ownerId: orgData.owner_id,
+  ownerEmail: orgData.owner_email,
+  plan: orgData.plan,
+  planCode: orgData.plan_code,
```
karelm, line 8: After speaking with Ana yesterday I think there might be more data we need, but if we don't know what that is at present its fine for us to add to this later. But if you do know what extra data we need, please could you add it to the org parser please? 😄

packages/bootstrap/middleware.js
karelm, line 16: could you change this to use the ORGANIZATIONS_INITIALIZED method please?
```suggestion
    case 'ORGANIZATIONS_INITIALIZED': {
```

josem:
Thanks so much @karelm! I think I've addressed all the changes now. 

I'll probably add more things to the parser as we need them but I think for now it might be ok. Sorry for not adding tests to the utils file before! :pray:

PR #1421, opened 2020-06-12
Queues Refactor | Integration tests for Queues

MayaUribe:
## Description

Add integration test for all our Queues
- Main queue
- Stories
- Past Reminders
- Analytics
- Drafts
- Campaigns

## Context & Notes

This is the first step for the Multiple Queue refactoring.
JIRA Card -> PUB-2841

## Screenshots
N/A

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/app-pages/index.test.js
+++ b/packages/app-pages/index.test.js
@@ -50,20 +50,18 @@ const _TestContextContainer = ({ children }) => <>{children}</>;
 
 const campaign = buildCampaign();
 
-const day1 = new Date('2020-01-01T11:00:00.000Z');
 const queuedPost1 = buildPostWithImage({
   overrides: {
     profileId: profileIG.id,
-    scheduledAt: getTime(day1),
+    scheduledAt: getTime(new Date('2020-01-01T11:00:00.000Z')),
     campaignDetails: campaign,
   },
 });
 
-const day2 = new Date('2020-01-02T11:00:00.000Z');
 const queuedPost2 = buildPostWithImage({
   overrides: {
     profileId: profileIG.id,
-    scheduledAt: getTime(day2),
+    scheduledAt: getTime(new Date('2020-01-02T11:00:00.000Z')),
   },
 });
 
--- a/packages/app-pages/index.test.js
+++ b/packages/app-pages/index.test.js
@@ -36,7 +36,7 @@ const profileIG = buildIGProfile({
   },
 });
 
-const appInitialState = {
+const initialState = {
   user: buildUser(),
   profileSidebar: {
     selectedProfileId: profileTwitter.id,
@@ -213,14 +213,12 @@ describe('AppPages | user interaction', () => {
     jest.clearAllMocks();
   });
 
-  test('non IG account should render proper tabs', () => {
+  it('should render proper tabs for non IG account', () => {
     render(
       <TestDragDropContainer>
         <AppPagesWithRouter />
       </TestDragDropContainer>,
-      {
-        initialState: appInitialState,
-      }
+      { initialState }
     );
     const {
       queueTab,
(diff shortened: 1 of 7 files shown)


packages/campaign/index.test.js
```diff
@@ -1,27 +1,211 @@
...
+  });
+};
+
+const campaignDetailsFields = async () => {
+  const totalScheduled = await screen.findByText(
+    `${campaign.scheduled} Scheduled`
+  );
+  const totalSent = await screen.findByText(`${campaign.scheduled} Scheduled`);
+  const dateRange = await screen.findByText(campaign.dateRange);
+  const campaignName = screen.getByText(campaign.name);
```
karelm, line 83: @MayaUribe would it make more sense to pass in `campaign` and `screen` as a parameter to the campaignDetailsFields function, then the campaign data + screen we are using is passed in by the caller instead of based on global data? Then if we add more tests, we can extract this accordingly, or is that an anti-pattern in this testing approach? - Asking from a functional perspective, as well as a learning perspective 😉

packages/campaign/index.test.js
```diff
@@ -1,27 +1,211 @@
...
+
+  return {
+    totalScheduled,
+    totalSent,
+    dateRange,
+    campaignName,
+  };
+};
+
+const campaignsListFields = async () => {
```
karelm, line 93: similarly to the above @MayaUribe, do you want to pass in screen instead to this function and access it from the params instead of from the global object?

packages/campaign/index.test.js
MayaUribe: Very interesting point @karelm 🤔

Specifically for the screen part, I wouldn't say it's an antipattern, but I would say is not really necessary because of the way screen behaves, if you see the test, I'm using `campaignDetailsFields()` twice in the same test, the first time for campaigns list and later on to re-check in campaign view (after the user clicked on a campaign), screen automatically refreshes any changes that happened.
One small disadvantage I can think of is that because we are sending the screen as a param, we wouldn't be able to see the autocompleted suggestions with all the possible queries while coding. (findBy*, getBy*, queryBy*, ...)

Does that make sense? 🤔

karelm: yeah, this is definitely not a blocker... just more of a discussion around usage. Whether to go fully functional or not with these methods, and have their props determine the outcome 👍

packages/app-pages/index.test.js
```diff
@@ -0,0 +1,352 @@
...
+  render,
+  screen,
+} from '@bufferapp/publish-test-utils/utils/custom-render';
+import { getTime } from '@bufferapp/publish-utils/date';
+import userEvent from '@testing-library/user-event';
+import {
+  buildUser,
+  buildProfile,
+  buildIGProfile,
+  buildPostWithImage,
```
albennett, line 13: This makes me happy, as it will save time in getting states in a certain order to test 💯

packages/app-pages/index.test.js
```diff
@@ -0,0 +1,352 @@
...
+
+    const {
+      queueTab,
+      analyticsTab,
+      awaitingApprovalTab,
+      draftsTab,
+      settingsTab,
+    } = tabMenuOptions();
+
+    const storiesTab = screen.getByRole('link', { name: /stories/i });
```
albennett, line 229: Have you been finding it fairly easy to target elements? I remember we use dynamic classes, so targeting through classes isn't an option.

MayaUribe: Yes! so far I haven't had major issues querying elements, which is a good sign from a11y perspective 🙌

MayaUribe:
> This is great @MayaUribe 🙌 I had some questions come up when reviewing this PR around unit vs integration tests and whether we should be writing unit tests first. I then read this article that was really interesting: https://kentcdodds.com/blog/write-tests . I'm excited to get deeper into the course and thought it was interesting how he recommends to "stop mocking so much stuff."

That is a **great** article, 🏆 he often talks about "Write tests. Not too many. Mostly integration", and I think that is spot on, in a single integration test we can cover a lot more meaningful use cases than with multiple unit tests (not saying unit tests are not necessary) but integration tests give us a lot more confidence.

PR #1427, opened 2020-06-16
[PUB-2795] Add Org Switcher to Publish UI

hamstu:
## Description

https://buffer.atlassian.net/browse/PUB-2795

## Context & Notes

- [x] Only show switcher for users with feature flip
- [x] Only show when num. orgs >= 2
- [x] Bug: Reloading after switching orgs does not work right (defaults to first org.)

## Checklist

#### Staging Deployment


packages/organizations/middleware.js
```diff
@@ -32,39 +36,56 @@ export default ({ dispatch, getState }) => next => action => {
...
+        p => p.organizationId === action.id
+      );
+      const firstProfileInOrganization = profilesInOrganization[0];
+      const newPath = profilePages.getRoute({
+        profileId: firstProfileInOrganization.id,
+      });
+      dispatch(push(newPath));
+      dispatch(
+        profileActions.selectProfile({ profile: firstProfileInOrganization })
+      );
```
hamstu, line 84: I don't love this logic here. Can't help but wonder if there's a better way to do it, but I think that all comes back to a rewrite/overhaul in our routing. (I.e., Ideally we could just do the push/navigate and everything else would work, but without the `selectProfile` it doesn't.)

hamstu: It's also worth noting that things don't **not** work on page reload yet. For example, if you switch to a new org/profile and reload the URL remains correct but the profile queue doesn't ever load. 

Will add to the to dos on this PR.

anafilipadealmeida: Hey @hamstu, I'm curious, why isn't your org switcher itemClick `switchOrganization` not just `organization.goTo({orgId})` from routes ? In theory that should be enough to trigger everything else in the app.

packages/app-shell/index.js
```diff
@@ -2,9 +2,13 @@ import { connect } from 'react-redux';
 import {
   newBusinessTrialists,
   preferencesGeneral,
+  organization,
 } from '@bufferapp/publish-routes';
 import { actions as modalActions } from '@bufferapp/publish-modals';
 import { getURL } from '@bufferapp/publish-server/formatters/src';
+
+// import { actions as orgActions } from '@bufferapp/publish-data-organizations';
```
anafilipadealmeida, line 10: Would we be able to delete this? 🙂

packages/organizations/middleware.js
```diff
@@ -2,9 +2,11 @@ import {
   actions as dataFetchActions,
   actionTypes as dataFetchActionTypes,
 } from '@bufferapp/async-data-fetch';
+import { organization } from '@bufferapp/publish-routes';
```
anafilipadealmeida, line 5: I believe we might not need this one anymore

PR #1429, opened 2020-06-18
e2e | Add an await for auth in e2e test

MayaUribe:
## Description

Add an await for auth in e2e test before clicking in the composer button.

## Context & Notes

- Added Testing Library Cypress plugin
- Included a `wait` for `/pusher/auth` to make sure the page is loaded before clicking on the composer button

## Additional comments
The main reason why I added the Testing Library Cypress plugin, is because it makes it a lot easier to query elements in the DOM without the need to add additional tags in the elements we need to test (`data-cy`), and as we get more familiar with the queries from React Testing Library, we'll be able to easily use the same queries for our e2e tests without having to switch too much context.

## Checklist

#### Staging Deployment

### DIFF
--- a/cypress/e2e/publish.js
+++ b/cypress/e2e/publish.js
@@ -9,18 +9,20 @@ describe('Publish', () => {
     cy.server();
     cy.route({
       method: 'POST',
-      url: '/pusher/auth',
+      url: '/rpc/queuedPosts',
       status: 200,
-    }).as('auth');
+    }).as('getQueuedPosts');
 
     cy.login();
     cy.visit('/');
-    cy.wait('@auth');
+    cy.wait('@getQueuedPosts');
     cy.get('[data-cy=open-composer-button]').click();
 
     const randomPostText = getRandomPostText();
     cy.get('[data-cy=composer-text-zone]').type(`${randomPostText}`);
     cy.findByText(/add to queue/i).click();
+    cy.wait('@getQueuedPosts');
+
     cy.findByText(randomPostText).should('exist');
   });
 });
--- a/cypress/e2e/publish.js
+++ b/cypress/e2e/publish.js
@@ -6,13 +6,21 @@ const getRandomPostText = () => {
 
 describe('Publish', () => {
   it('creates a new post', () => {
+    cy.server();
+    cy.route({
+      method: 'POST',
+      url: '/pusher/auth',
+      status: 200,
+    }).as('auth');
+
     cy.login();
     cy.visit('/');
+    cy.wait('@auth');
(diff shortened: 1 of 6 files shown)


cypress/e2e/publish.js
```diff
@@ -6,13 +6,21 @@ const getRandomPostText = () => {
...
+    cy.server();
+    cy.route({
+      method: 'POST',
+      url: '/pusher/auth',
+      status: 200,
+    }).as('auth');
+
     cy.login();
     cy.visit('/');
+    cy.wait('@auth');
```
MayaUribe, line 18: This one makes it work for me locally, I tried adding something like this later on:
```
cy.route({
  method: 'POST',
  url: '/rpc/queuedPosts',
  status: 200,
}).as('createUpdate');
```
But this doesn't work for some reason 🤔

hamstu: Ah, I think it might be because Cypress can't intercept `fetch()` calls.
image
(Source: https://docs.cypress.io/guides/guides/network-requests.html#Testing-Strategies)

hamstu: There are a few ways to approach fixing this in the linked GitHub issue. I think there's one where you just delete `window.fetch` in a Cypress config file somewhere so that your fetching library replaces it with a polyfill using XHR instead. (This _should_ work for us, since the RPC client uses `isomorphic-fetch > whatwg-fetch`, [which seems to do that](https://github.com/github/fetch/blob/74e4fe8a4731b4e233ac7da42c960a351fd3400f/fetch.js#L460).)

Are you up for giving that a try here? 😄

cypress/support/index.js
```diff
@@ -1,3 +1,5 @@
+import '@testing-library/cypress/add-commands';
```
hamstu, line 1: 😍

MayaUribe:
> Hey @MayaUribe - thanks for putting this together. So your theory is that we're not getting connected to Pusher properly? I have seen these tests fail a lot more lately, so I really appreciate you taking a look at this.

That's what I was thinking at first (at least locally) but I think mainly because the pusher was the only request I was seeing in Cypress, but I changed it to queuedPosts now (after adding up your suggested changes)
 
> I left one comment on why `cy.route()` is not working. Let's keep working on this one; really keen to get rid of the flakiness here! 🙏
Thank yoouu! 🙌

cypress/support/index.js
MayaUribe: Oooh that was it!!! Thank you so much! 🎉 😍 🥇

PR #1456, opened 2020-07-13
[PUB-2904 ] Inject orgs data in user

anafilipadealmeida:
## Description
- Temporarily Inject orgs data in user
- Temporarily dispatch INIT_USER when selecting new org, to inject new org data in the user data. To be removed after org_switcher rollout.

## Context & Notes
Once we rollout the org switcher the plan is to stop relying on user data to get plan/org data and rely on organization data instead.
I've only injected the plan_code from orgs in this PR, as this was just to set the mechanism. The full org data injection is happening in a different PR.
https://buffer.atlassian.net/browse/PUB-2904

## Checklist

#### Staging Deployment


packages/organizations/middleware.js
```diff
@@ -95,6 +95,11 @@ export default ({ dispatch, getState }) => next => action => {
       break;
     }
 
+    // Temporarily dispatch user data, to inject new org data in it. To be removed after org_switcher rollout
+    case `setCurrentOrganization_${dataFetchActionTypes.FETCH_SUCCESS}`:
+      dispatch({ type: 'INIT_USER' });
```
karelm, line 100: @anafilipadealmeida will this trigger a refetch?

anafilipadealmeida: Yes, it's not ideal but it's the safest way of doing it, as it was not possible to make all user references rely only on user.state

packages/server/rpc/user/index.js
```diff
@@ -3,14 +3,37 @@ const { method } = require('@bufferapp/buffer-rpc');
...
+      userData &&
+      userData.features &&
+      userData.features.includes('org_switcher');
+
+    // Temporarily injecting org plan data in users. To be removed after org switcher rollout.
+    if (hasOrgSwitcher) {
+      const orgs = await PublishAPI.get({
+        uri: `1/user/organizations.json`,
+        session,
+      });
```
karelm, line 24: is it faster to just always do this query and the userData query simultaneously instead?

e.g. something along these lines
```
Promise.all([publishApi.getUser(), publishApi.getOrg()])
.then(([user, org]) => { return [userParser(user), orgParser(org)]; }).then(([user, org]) => ({...user, plan_code: org.plan_code}));
```

packages/server/rpc/user/index.js
```diff
@@ -3,14 +3,39 @@ const { method } = require('@bufferapp/buffer-rpc');
...
+        uri: `1/user.json`,
+        session,
+        params: {
+          includes: 'avatar',
+        },
+      }).catch(PublishAPI.errorHandler),
+      PublishAPI.get({
+        uri: `1/user/organizations.json`,
+        session,
+      }),
```
karelm, line 18: close @anafilipadealmeida, you want to catch the error on the organizations method, because if it fails, you want the user data to continue to return safely. 

```
PublishAPI.get({
        uri: `1/user/organizations.json`,
        session,
      }).catch(() => {return {};});
```

packages/server/rpc/user/index.js
```diff
@@ -3,14 +3,39 @@ const { method } = require('@bufferapp/buffer-rpc');
...
-      uri: `1/user.json`,
-      session,
-      params: {
-        includes: 'avatar',
-      },
-    })
-      .then(user => parsers.userParser(user))
-      .catch(PublishAPI.errorHandler)
+  async (_, { session }, res, { PublishAPI, parsers }) => {
+    return Promise.all([
```
karelm, line 7: you could actually make this more async/await-y by instead of doing return Promise.all do the following:

const [user, org] = await Promise.all([publishApi.getUser, publishApi.getOrg]);
then move your code to a more sync approach.

Again, this is just a thought, don't mean to keep making you rewrite this code 😅

packages/server/rpc/user/index.js
```diff
@@ -3,14 +3,39 @@ const { method } = require('@bufferapp/buffer-rpc');
...
+      PublishAPI.get({
+        uri: `1/user.json`,
+        session,
+        params: {
+          includes: 'avatar',
+        },
+      }).catch(PublishAPI.errorHandler),
```
karelm, line 14: you probably want to remove this from here and instead do this after the last bracket of the Promise.all(...) call instead to be more consistent with other places

packages/server/rpc/user/index.js
```diff
@@ -3,14 +3,41 @@ const { method } = require('@bufferapp/buffer-rpc');
...
+        },
+      }),
+      PublishAPI.get({
+        uri: `1/user/organizations.json`,
+        session,
+      }).catch(() => {
+        return [];
+      }),
+    ])
       .catch(PublishAPI.errorHandler)
```
karelm, line 22: the thing with having this before you process the data is that the errorHandler will run and hand over to the then call below, I would suggest moving the catch as the last thing that happens in this promise chain if possible. This way all error states will get consumed by the error handler

karelm: essentially:
```
        return user;
      }).then(PublishAPI.errorHandler);
```

PR #1474, opened 2020-07-22
[PUB-2811] Hide Org Admin paths for non admin users

anafilipadealmeida:
## Context & Notes
https://buffer.atlassian.net/browse/PUB-2811

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/server/parsers/src/userParser.js
+++ b/packages/server/parsers/src/userParser.js
@@ -85,6 +85,8 @@ module.exports = userData => ({
   showUpgradeToProCta:
     userData.billing_plan_base === 'free' && !userData.is_business_team_member,
   analyzeCrossSale: userData.is_analyze_customer,
+  canManageSocialAccounts: true,
+  hasAccessTeamPanel: !userData.billing_plan_base === 'free',
 
   // Org data
   plan:
--- a/packages/server/rpc/user/index.js
+++ b/packages/server/rpc/user/index.js
@@ -65,6 +65,8 @@ module.exports = method(
               hasShareNextFeature: planBase !== 'free',
               hasUserTagFeature: planBase !== 'free',
               analyzeCrossSale: user.analyzeCrossSale && ownerId === user.id, // to do: return organization products array
+              canManageSocialAccounts: isAdmin,
+              hasAccessTeamPanel: planBase === 'business' && isAdmin,
             };
           }
         }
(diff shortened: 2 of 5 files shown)


packages/server/rpc/user/index.js
```diff
@@ -65,6 +65,8 @@ module.exports = method(
               hasShareNextFeature: planBase !== 'free',
               hasUserTagFeature: planBase !== 'free',
               analyzeCrossSale: user.analyzeCrossSale && ownerId === user.id, // to do: return organization products array
+              canManageSocialAccounts: isAdmin,
+              hasAccessTeamPanel: planBase === 'business' && isAdmin,
```
MayaUribe, line 69: @anafilipadealmeida so I should be able to access the team panel if I'm admin (not necessarily the owner) and a business plan? just confirming 🙂

anafilipadealmeida: Yep 😄 That's right!

packages/server/parsers/src/userParser.js
```diff
@@ -85,6 +85,8 @@ module.exports = userData => ({
   showUpgradeToProCta:
     userData.billing_plan_base === 'free' && !userData.is_business_team_member,
   analyzeCrossSale: userData.is_analyze_customer,
+  canManageSocialAccounts: true,
```
MayaUribe, line 88: @anafilipadealmeida curious if this is temporary? 🙂

packages/server/rpc/user/index.js
anafilipadealmeida: Giving a bit more context: users with admin access can add/manage social accounts or team members; only business accounts (starting with premium business w/ 2 users) have team members 😄

packages/server/parsers/src/userParser.js
anafilipadealmeida: Yes it is! So the plan is to override this value for now in the rpc/user by injecting the org data. Once we rollout the org switcher, all these fields that are being overridden by orgs will be migrated to the orgs and removed from the user parser 😄

anafilipadealmeida: In this particular case, without the org switcher we always display the manage social accounts option. With the org switcher, `canManageSocialAccounts: isAdmin` so we will only display it if the user is an admin in the selected org

MayaUribe: I see, yes that makes perfect sense, thank you for clarifying! 😊

PR #1494, opened 2020-07-30
Upgrading buffer-publish chart to 1.1.1

kiriappeee:
Upgraded the chart `buffer-service` used by `buffer-publish` from version 1.0.3 to 1.1.1


kiriappeee:
heyya @hamstu . Nothing much here. Just an automated PR created to upgrade our chart :) . Would you be up for reviewing it and if all is well, merging it too? Nothing much has changed here. Just a bunch of sensible defaults which your values already override.

hamstu:
@kiriappeee I'm off until next week but will take care of this asap when I return 😄

PR #1513, opened 2020-08-31
Fast Front-ends: Separate RPC server and index.html

hamstu:
## Checklist

#### Staging Deployment


packages/profiles/reducer.js
```diff
@@ -1,17 +1,11 @@
...
-  typeof window.bufferData !== 'undefined' &&
-  typeof window.bufferData.profiles !== 'undefined' &&
-  window.bufferData.profiles;
-
 export const actionTypes = keyWrapper('PROFILES', {
   FETCH_SINGLE_PROFILE: 0,
 });
 
-export default (state = profilesData || [], action) => {
+export default (state = [], action) => {
```
anafilipadealmeida, line 8: @hamstu, if we are no longer using state.publishProfiles, then perhaps we can remove this entire reducer.  I took a look at the profile-sidebar reducer and it seems it's handling the `singleProfile_${dataFetchActionTypes.FETCH_SUCCESS}`, so in theory it's fine

hamstu: Good one! Yeah – I should be able to remove it 👍

anafilipadealmeida (approved):
Looks great to me, I'll be approving it but just a heads up the tests seem to be failing! My hunch is that the mocked data has changed and perhaps we miss some props used in the LoadingGate to properly run the tests?

packages/profiles/middleware.js
```diff
@@ -5,32 +5,16 @@ export default ({ dispatch }) => next => action => {
...
+      dispatch(
+        dataFetchActions.fetch({
+          name: 'profiles',
+          args: {
+            // Passing this prevents the profiles FETCH_SUCCESS from affecting the
+            // rest of the publish app.
+            forAnalyze: action.forAnalyze || false,
+          },
+        })
+      );
```
karelm, line 17: Hey @hamstu, I'm just wondering how come you've completely removed the statically loaded data option? This solution allowed for both solutions, so if we decide to keep the server loaded data, or perhaps decide to use a webworker to improve load times later and statically set the data on window.bufferData we can still do that, but we can also have the server fetched approach too.

hamstu: Hey @karelm! I may have answered this with [my comment in Slack](https://buffer.slack.com/archives/C151RRDL1/p1599139021111300?thread_ts=1599123910.102900&cid=C151RRDL1) but I can see your point in keeping the static data checks, as we may be able to leverage it in the future even without a traditional server backing the site.

I think the decision came down to just removing unneeded code as we don't presently have a clear idea of how we're going to optimize the fetching and rendering for Fast Front-ends, but it's coming up soon, so we'll certainly be revisiting this.

Let me know if that helps – and thank you so much for your keen eyes on this! ❤️

karelm: Thanks for clarifying, I did figure that might be the case. I'm still slightly concerned about premature optimization with the impact of the server side load time Vs client side load times for our users. But hopefully they will all improve once we start caching in the cdn 🤞

PR #1515, opened 2020-09-01
[PUB-2999] Delete references to feature flips 100% rolled out or not used

anafilipadealmeida:
## Description
Delete references to feature flips 100% rolled out or not used:
 - hasIGLocationTaggingFeature: 100% rolled out
- hasIGDirectVideoFlip: 100% rolled out
- has_simplified_free_plan_ux: nobody has it

## Context & Notes
https://buffer.atlassian.net/browse/PUB-2999

## Checklist

#### Staging Deployment


anafilipadealmeida:
This is a great comment @hamstu ! There are some things I found out with this PR:
1) I don't think any free users will get the share next button 😂 I've made some tests with my account after making the changes and the behavior was still the same, so I went to the code and realized that this is where the share next is being handled: https://github.com/bufferapp/buffer-publish/blob/master/packages/composer/interfaces/buffer-publish.jsx#L50-L51. With that in mind, I made the decision to not spend more time trying figure out why that code in the composer utils exists, as it's a different matter.
2) I'm not sure free users who are not team members have it. I've impersonated some and none has it. Last month I've changed `has_simplified_free_plan_ux: userData.features.includes('has_simplified_free_plan_ux' )` to `has_simplified_free_plan_ux: false` but if it existed it would still come in the features array. I've also ran the query https://app.redash.io/buffer/queries/431672/source for it and nothing showed. I do agree it would be less confusing to address it not as a feature flip, but it was supposed to be received in the features array and nobody seems to have it 🤷‍♀️

One thing I can do is to keep track of these things (since they are out of the scope for the org switcher/user parser clean up) and by the end of the cleanup story bring them to the team so we can decide which ones should be converted to jira tickets. How do you feel about it? Anything I might be missing? 😅

hamstu:
Thanks for the great response @anafilipadealmeida!

> so I went to the code and realized that this is where the share next is being handled [...]

Haha, I should have known with the composer. It's like Inception, too many layers deep. 😂  Good find!

> but if it existed it would still come in the features array

As far as I can tell there's never been a feature `has_simplified_free_plan_ux`, do you have a referenc to somewhere in the code where we search for a feature with that name? 

> One thing I can do is to keep track of these things (since they are out of the scope for the org switcher/user parser clean up) and by the end of the cleanup story bring them to the team so we can decide which ones should be converted to jira tickets. How do you feel about it? Anything I might be missing? 😅

Agreed! I don't want to bikeshed on this little thing too much. Let's table it for now and we can revisit in future clean ups. 😄

anafilipadealmeida:
@hamstu I went back some months in github to make sure it wasn't something I had changed by mistake: https://github.com/bufferapp/buffer-publish/blob/b0f2e1dd18430306d6403bac4a7d07b9f2b878a9/packages/server/parsers/src/userParser.js#L52-L54
Maybe it never worked? 😅

hamstu:
Ooh, good find! 😲 

> Maybe it never worked? 😅

Yeah! I can't see any evidence that it would have worked. Well all the more reason to remove it now!

PR #1518, opened 2020-09-03
Fast Front-ends: Async Bugsnag

hamstu:
## Description

Identify the user with Bugsnag async.

## Checklist

#### Staging Deployment


packages/server/webpack.config.prod.js
```diff
@@ -41,6 +41,8 @@ const plugins = [
   }),
   new webpack.EnvironmentPlugin({
```
hamstu, line 42: This plugin basically rewrites references to `process.env.VAR_NAME` to whatever that env var is set to at build time. The string after the key is the default value if no env var was found defined when buidling.

packages/web/components/ErrorBoundary/index.jsx
```diff
@@ -6,25 +6,29 @@ import BoundaryFallback from './errorComponent';
...
 
-let BugsnagErrorBoundary;
-
 /**
  * Setup Bugsnag
  * https://docs.bugsnag.com/platforms/javascript/react/#installation
  */
-if (window._bugsnagConfig) {
+const getReleaseStage = () => {
+  if (window.location.href.match(/local\.buffer/)) {
```
albennett, line 14: I didn't realize we had bugsnag working locally. If we ever get close to our quota, we could consider turning this off locally. 🤔

packages/thirdParty/middleware.js
```diff
@@ -39,6 +39,16 @@ export default ({ dispatch, getState }) => next => action => {
       dispatch({ type: actionTypes.FULLSTORY, result: action.result });
       dispatch({ type: actionTypes.APPCUES, result: action.result });
       dispatch({ type: actionTypes.ITERATE, result: action.result });
+      dispatch({ type: actionTypes.BUGSNAG, result: action.result });
+      break;
+
+    case actionTypes.BUGSNAG:
```
albennett, line 45: These Codecov warnings are a bit annoying when reading code 😅

packages/web/components/ErrorBoundary/index.jsx
```diff
@@ -6,25 +6,29 @@ import BoundaryFallback from './errorComponent';
...
+  if (window.location.href.match(/dev\.buffer/)) {
+    return 'staging';
+  }
+  return 'production';
+};
+
+let BugsnagErrorBoundary;
+
+if (process.env.NODE_ENV === 'production') {
+  const BUGSNAG_KEY = '6d235b284e8baf3c8d669a1991844969';
```
albennett, line 26: Don't we get the bugsnag key from process.env? Is this ok to be public?

packages/web/components/ErrorBoundary/index.jsx
hamstu: Well, the earlier code will stop it from running locally (`if (process.env.NODE_ENV === 'production') { ...`) so this would just run in production or staging. (I did want to test locally temporarily, thus the regex which has local, but it will not run there normally.)

packages/web/components/ErrorBoundary/index.jsx
hamstu: It was public before so it's all good 👍  – it has to be since you create this on the client side. Thanks for double checking! 🦅 👁️

packages/thirdParty/middleware.js
```diff
@@ -39,6 +39,16 @@ export default ({ dispatch, getState }) => next => action => {
...
+      if (window && window.bugsnagClient) {
+        window.bugsnagClient.user = {
+          id: action.result.id,
+          adminLink: `https://buffer.com/admin/user/${action.result.id}`,
```
albennett, line 49: is `adminLink` new? 😍

packages/thirdParty/middleware.js
hamstu: Agreed! DM'd Dan since we're supposed to be getting rid of it because we have Code Climate.

packages/thirdParty/middleware.js
hamstu: Nope, I added this a while back, it should be in the user tab of Bugsnags (but just the ones from Publish dashboard, not Publish API for example)
image

albennett (approved):
Hey @hamstu I left a few comments mainly to just understand it better. I'm excited about all the faster front end work 🚀 🙌

packages/thirdParty/middleware.js
albennett: Ah, you're right!

PR #1521, opened 2020-09-08
Fast Front-ends: Check for modals and notification query params client side

hamstu:
## Description

Logic is the same, just moving the checks client side. For simplicity still putting the modal and notification data into global vars as it felt more lightweight than creating an entire new store/reducer.

## Checklist

#### Staging Deployment


packages/modals/middleware.js
```diff
@@ -18,6 +18,20 @@ import {
 export default ({ dispatch, getState }) => next => action => {
   next(action);
   switch (action.type) {
+    case 'APP_INIT': {
+      let modal = null;
```
albennett, line 22: I was curious about how you were going to handle the modal key params. 😄 

We rely quite a lot on requests on APP_INIT. As the list grows, we could question whether all the requests need to be initially called, and whether some can be called after we get key data for visual purposes. Though I may be misunderstanding something! 😄

packages/thirdParty/middleware.js
```diff
@@ -10,16 +10,6 @@ import * as FullStory from '@fullstory/browser';
 
 import { actionTypes } from './reducer';
 
-const checkExtensionInstalled = () => {
```
albennett, line 13: Hm, is this no longer needed?

packages/server/lib/generateIndexHtml.js
```diff
@@ -13,14 +11,7 @@ const getDatadogRumScript = require('./embeds/datadogRum');
...
 
-const getHtml = ({
-  staticAssets,
-  isProduction,
-  isStandalone,
-  notification,
-  modalKey,
-  modalValue,
-}) => {
+const getHtml = ({ staticAssets, isProduction, isStandalone }) => {
```
albennett, line 14: 😍

albennett (approved):
Great job @hamstu 🙌 Before long, allll the scripts will be gone.

packages/thirdParty/middleware.js
hamstu: Yeah it's not used any more 👍

packages/modals/middleware.js
hamstu: Good one! The reason I use APP_INIT is that we might trigger navigation / redirects once the React app renders, so we need to pull the query strings first, before that (potentially) happens.

We could also dispatch another event that comes early, but it felt simple enough to piggyback on `APP_INIT` for this one. Let me know if that makes sense!

hamstu:
Thank you @albennett! And yes, we're getting close to a simplified experience here I hope! 😄

PR #1556, opened 2020-09-21
[PUB-3050] Prevent customers from using their personal instagram accounts

albennett:
## Description
Prevent customers from using their personal instagram accounts until they convert / enable direct scheduling
https://buffer.atlassian.net/browse/PUB-3050
- Only show notification component for personal ig profiles in tab content.
- Update banner button copy with better nomenclature of action
- Refactor direct posting modal to modal package

Originally, we wanted to leave out the nav tabs and import it once in the profile-page package, however, after feedback from Advocacy, we decided to allow access to the Settings tab so a user can still pause their queue. Honestly, it was painful having to add this logic to every tab package. It's temporary, though 😅 
## Context & Notes
I noticed a few other improvements while working on this task. As this PR is already long and time-sensitive, I'll add the improvements in separate PRs.
- The `isInstagramLoading` button is part of the queue state and is imported there. However, it was being used across tabs, which led to having to redirect to the queue before displaying the modal. This state should live in the direct-posting-modal and be imported in ProfilePage package. 

Questions:
- Do we know why we keep the nav tabs for the locked profile notification? 
- Are we still expected to add other buffer package dependencies to package.json files since we're no longer publishing to npm?

https://share.buffer.com/z8uZNODN
## Checklist

#### Staging Deployment


hamstu (approved):
Great stuff @albennett! Sorry you had to go through all those tabs. 😭 

> The isInstagramLoading button is part of the queue state and is imported there. However, it was being used across tabs, which led to having to redirect to the queue before displaying the modal. This state should live in the direct-posting-modal and be imported in ProfilePage package.

I see what you mean, yeah. I guess if we didn't worry about showing this loader we could maybe get rid of it altogether? In fact, won't we be able to remove this in the future since only direct-posting enabled IG profiles will be connected?

> Do we know why we keep the nav tabs for the locked profile notification?

I think you and I already talked about this and I still can't recall any specific reason. Maybe it's similar to your situation where we wanted to keep the settings available? 🤔 

> Are we still expected to add other buffer package dependencies to package.json files since we're no longer publishing to npm?

This one is a bit tricky, but the short answer is **yes, we should try** to keep doing that for consistency and future-proofing if we change things in the future. Running `yarn` / `yarn install` will _hoist_ all the dependencies into the root `node_modules` – which means if Package A and Package B both require Package C, but Package A doesn't have Package C in it's package.json, **everything still works** because Package C was installed for Package B and hoisted, and so it's still found when webpack looks for it at build time. Hopefully that makes sense! (And with that in mind, if you were to say, pull Package A into it's own repo, _then_ you would start seeing issues with missing dependencies.)

albennett:
> I see what you mean, yeah. I guess if we didn't worry about showing this loader we could maybe get rid of it altogether? In fact, won't we be able to remove this in the future since only direct-posting enabled IG profiles will be connected?

Great point, I was thinking we'd still need the direct posting flow for the cases where users remove their connected fb page etc. However, that wouldn't change the db service_type, so a faq link in the failed post would prob suffice. 

And similarly, with the locked notifications, I remember seeing in Core where we'll get rid of locked profiles soon, too.

PR #1565, opened 2020-09-23
Lighthouse CI

hamstu:
## Description
Lighthouse CI will start running at every push to any branch. 
https://buffer.atlassian.net/browse/PUB-3105

## Checklist

#### Staging Deployment


lighthouse-puppeteer-login.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * @param {puppeteer.Browser} browser
```
albennett, line 2: Curious about puppeteer. Is there a reason we use this instead of cypress?

.github/workflows/lighthouse.yml
```diff
@@ -0,0 +1,49 @@
...
+          URL: ${{ fromJSON(steps.wait-deploy.outputs.status).target_url }}
+        run: |
+          until $(curl --output /dev/null --silent --head --fail $URL); do
+            printf '.'
+            sleep 5
+          done
+
+      - name: run Lighthouse CI
+        env:
+          LHCI_COLLECT_URL: ${{ fromJSON(steps.wait-deploy.outputs.status).target_url }}
```
albennett, line 40: to confirm, this will be the staging url? I read in the docs that LI will run against it 3 times for each url. 🤔 Do you know why this? If there's no advantage, I may change it to 1.

albennett, line 40: I think I'm a bit confused where we're actually sending the value of the staging urls that LH should run against 🤔

hamstu: [Loom answer](https://www.loom.com/share/c2293651288b46b8944c837fd36f5858?from_recorder=1).

tl:dnw; This the line that does that. It extracts the result of a past action step (id: `wait-deploy`) [which contains an output called `status`](https://github.com/hamstu/github-action-wait-for-commit-status/blob/master/src/index.js#L37), which is a JSON object with the `target_url` in it.

lighthouse-puppeteer-login.js
hamstu: [Loom answer](https://www.loom.com/share/c2293651288b46b8944c837fd36f5858?from_recorder=1).

This is just a file used to setup or prepare Lighthouse CI to run the tests. So it's using puppeteer since that's the automation library Lighthouse CI uses itself to run the audits under the hood. 

Cypress itself wouldn't work here, as it's designed as it's own full blown testing software, and so it's more of an apple and oranges sort of thing here, if that makes sense?

albennett (approved):
Hey @hamstu thanks for all the work on the Lighthouse CI. 🙌 This is looking great.

I'm still a bit wary on whether we should run this for every push to a branch. However, I think it makes sense to start out this way and we can reiterate based on feedback (potentially deploy when we merge to master, and have a specific command to run if we want to test against a branch 🤔 ). 

I'll deploy this later today. 🚀

albennett:
Actually, a few more comments first 😅 

I noticed LH runs 3 times by default against a branch. I may configure it to run 1 time unless there's a benefit of running it more than once? 

Also, where is the link to see the LH results? I may try to get this to post somewhere more transparent.

hamstu:
@albennett At least three times is encouraged as it helps smooth out any inconsistencies in the run results and will overall make things less 'spiky', if that makes sense. (Things like CPU usage and memory can vary, network conditions might change, etc, so that's why it's run a few times.). I believe running Lighthouse in Chrome does this as well. So I'd strongly advocate for keeping it.

Results are available in our LHCI dashboard; which you can read about here: https://github.com/bufferapp/lhci-heroku 😄

PR #1576, opened 2020-09-30
add Zendesk widget to Publish dashboard via Help Menu item in App Shell

mickmahady:
## Description
Adds the Zendesk support widget to the Publish dashboard. It is triggered from a new Help Menu item in the AppShell. We also suggest Help Center articles to the user based on the path they are currently on.

Will deploy once Advocacy wants to roll out the widget. Likely, mid-October.

Peek 2020-09-30 17-19

## Checklist

#### Staging Deployment


packages/server/index.ejs
```diff
@@ -223,6 +223,15 @@
     }}();
   </script>
 
+  <!-- Zendesk Widget: A -->
+  <script id="ze-snippet" src='https://static.zdassets.com/ekr/snippet.js?key=<%= process.env.ZENDESK_WIDGET_KEY %>'></script>
```
hamstu, line 227: Would you be up for adding `async` to this `script` tag? It should help prevent blocking parsing / loading the rest of the page. [[ref](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#attr-async)]

hamstu (approved):
Thanks so much for taking care of all this work @mickmahady! LGTM - when you're ready! 🚀 

Quick Q; do you have prettier / eslint working in your setup? It all looks good to me, so I'm only asking in case there's places where it's not formatted. Not a blocker by the way! Just thought I'd ask.

mickmahady:
Thanks for the review, @hamstu! I do have both prettier and es-lint. I think the formatting should be good. Prettier pointed out a few issues such as missing semi-colons which I resolved :grin:

packages/server/index.ejs
mickmahady: @hamstu when I add the async tag, the widget icon flashes on screen before it is eventually hidden by code triggered later on. I'm wondering if there is another approach I could take here? The flashing icon doesn't feel great :grimacing:

packages/thirdParty/components/RouteMonitor/index.jsx
```diff
@@ -19,6 +20,11 @@ function RouteMonitor({ pathname, appCues }) {
         window.Appcues.track('Viewed Shop Grid Tab');
       }
     }
+
+    // Let Zendesk Widget know when we've changed pages
+    if (zendeskWidget && zendeskWidget.loaded && window && window.zE) {
+      updateHelpCenterSuggestions(pathname);
```
mickmahady, line 26: @hamstu it looks like this code is only triggered when a route changes. I'm also hoping to call `updateHelpCenterSuggestions()` on the initial page load, passing the current path as an argument. Any advice on how/where to do this, since it can't be done in the RouteMonitor? :pray:

packages/server/index.ejs
```diff
@@ -222,5 +222,16 @@
       analytics.load("<%= process.env.SEGMENT_KEY %>");
     }}();
   </script>
+
+  <!-- Zendesk Widget: P -->
+  <% if (process.env.NODE_ENV === 'production') { %>
+    <script id="ze-snippet" src='https://static.zdassets.com/ekr/snippet.js?key=<%= process.env.ZENDESK_WIDGET_KEY %>'></script>
```
hamstu, line 228: Since the `ZENDESK_WIDGET_KEY` doesn't differ between prod and dev you can just hard code it in here and remove it from the webpack files.

packages/thirdParty/components/RouteMonitor/index.jsx
hamstu: Hmm, it does run on first load, so I'm guessing this is when you have `async` on the script that it fails? (Since one of those vars, like `window.zE`, won't be defined yet.)

Let's leave out `async` for now. It sounds like [better optimizations are on ZenDesk's radar](https://support.zendesk.com/hc/en-us/community/posts/360004214608-Web-Widget-Performance-FAQs), and it's only going to overcomplicate things for ya.

On a misc. related note, checking for `window` here would not be necessary as it's always defined in the context of the browser.

hamstu (approved):
Thanks so much @mickmahady! tl:dr; let's leave out `async` if it causes issues. LGTM when you're ready and have made the changes I mentioned.

packages/server/index.ejs
hamstu: Closing the loop here, per my last review let's remove `async` for now then 😄

packages/thirdParty/components/RouteMonitor/index.jsx
mickmahady: Thanks, @hamstu! It seems to fail without the async tag too. On first load, `updateHelpCenterSuggestions()` is not called until the path changes :thinking: I don't think it's a big enough issue to block me from rolling out but it would be awesome to figure it out and deploy a fix later :sweat_smile:

mickmahady: It does seem to run on page load but both `zendeskWidget.loaded` and `window.zE` are falsey on the first page load, it seems. I figured they'd be loaded in time, like AppCues above it, but that doesn't seem to be the case :sweat_smile:

PR #1583, opened 2020-10-06
[PUB3006] Replace org/plans references in profiles by orgs data

anafilipadealmeida:
## Description
Replace org/plans references in profiles by orgs data:
- Delete no longer needed `shouldShowGridPreview`
- Delete unused `hasOrganizationMembers` prop in profile parser
- Delete unused `isInstagramContributor` and `isInstagramManager` methods in composer
- Delete unused profile-filtering utils in composer
- Delete `isBusinessProfile` in composer
- Delete `business` prop in profile parser

## Context & Notes
https://buffer.atlassian.net/browse/PUB-3006
The last orgs clean up task 😃 

## Checklist

#### Staging Deployment


packages/composer-user-tags/components/UserTags/index.jsx
```diff
@@ -89,9 +89,7 @@ const UserTags = ({
     }
     setShowInput(false);
     selectedChannels.forEach(channel => {
-      if (channel.isBusinessProfile) {
```
anafilipadealmeida, line 92: User Tagging is available for Pro and up. @albennett, I'm wondering, why were we checking if each profile was in a business plan before tracking?

hamstu: Ah perhaps this was _meant_ to check if this was an Instagram business profile, since you can only add user tags for those. @albennett might remember better.

So maybe this `if` should be checking `isInstagramBusiness` instead? 🤔

hamstu (approved):
So much great clean up! 🙏  LGTM, assuming we can resolve the confusion around the IG profile check.

packages/composer-user-tags/components/UserTags/index.jsx
albennett: Great one! Honestly, I'm having trouble remembering. We originally were back and forth around whether we want to make it available to pro and up or business. There was trickiness with the [omnibox state](https://share.buffer.com/eDuP9ny2), where we chose to still show the user tags option regardless of the channel plan type. It's likely I didn't change it here when we decided to do pro, too. 🙈 I'll let Julian know. So glad we've solved some of this omnibox trickiness with the org switcher.

albennett: Just saw your comment @hamstu . That might also be the case, as user tagging wasn't supported yet on mobile (though I think they might be adding it now 🤔 ). I'll check with Andy, but let's change it to `isInstagramBusiness` for now.

anafilipadealmeida: Thank you both! I'll change it to `isInstagramBusiness` for now! I @albennett please ping me in case it was supposed to check the plan and I'll pass the org feature instead 🙂

albennett: Sounds great, thanks @anafilipadealmeida 🙌 If we do allow it for reminders now, I'll add a separate jira task for it.

albennett (approved):
The last org switcher clean up task 🙌 🎉 Great job.

packages/composer-user-tags/components/UserTags/index.jsx
albennett: This also makes me think of the point Karël brought up during code critique around channel features. When we do the feature flip work, it'd be great to document the channel type feature differences and whether it makes sense to return a features array on the profile endpoint, too. (though I think the ig feature differences will be less frequent with the personal deprecation)

packages/composer-user-tags/components/UserTags/index.jsx
```diff
@@ -110,7 +113,10 @@ const UserTags = ({
     const globalTags = removeClientXY(tags);
     saveGlobalTags(globalTags);
     selectedChannels.forEach(channel => {
-      if (channel.isBusinessProfile) {
+      if (
+        channel.service?.name === 'instagram' &&
```
anafilipadealmeida, line 117: We don't have the `isInstagramBusiness` prop here, so instead of passing it through the composer I just took the traditional approach

PR #1639, opened 2020-11-25
[PUB-3161] Add emoji picker plugin to slate

anafilipadealmeida:
## Description
Add emoji picker plugin to slate:
- Adjust webpack.config to be able to use emoji-mart css
- We're using the same library, styles and code as in the hack week (apart from the draft js plugin part)

Bonus: Fix snapshot tests from hot fix: https://github.com/bufferapp/buffer-publish/commit/9ec58f020a294400225d71bdb89fb4789d97c059

## Context & Notes
I've tried to add some tests but they don't run apparently without updating jest, but then storybook doesn't seem to work as well, so I'm delaying that and adding a task for it.
https://buffer.atlassian.net/browse/PUB-3161

https://share.buffer.com/jkuYonwn
The side scroll when opening the emoji picker will not be an issue when we replace draft js by this one

## Checklist

#### Staging Deployment


packages/composer-editor/components/Editor/index.jsx
```diff
@@ -7,13 +7,7 @@ import styled from 'styled-components';
...
 import { fontSize, fontFamily } from '@bufferapp/ui/style/fonts';
 
-const Editor = styled(Editable)`
-  color: ${grayDarker};
-  padding: 8px;
-  line-height: 24px;
-  font-size: ${fontSize};
-  font-family: ${fontFamily};
-`;
+import EmojiPicker from '../Plugins/EmojiPicker';
```
hamstu, line 10:
```suggestion
import EmojiPicker from '../plugins/EmojiPicker';
```

Can we use lowercase here for directories unless it's a component folder? I think that would be a little more consistent.

hamstu (approved):
One quick thing, otherwise this is great! 💯

packages/composer-editor/components/Editor/index.jsx
anafilipadealmeida: Yes, you are right! It took me a couple of commits because I was in a non-case-sensitive system 😅

PR #1644, opened 2020-12-02
[PUB-3166] Add overCharLimit decorator

anafilipadealmeida:
## Description
Add overCharLimit decorator and highlight decorator for mentions, hashtags and links.

## Context & Notes
~~Updating our testing libraries next so we can add tests~~  With react testing-library I'm only able to render the component, but not to interact with it properly. After a lot of digging and adding myself to their Slack channel, it seems interactions only work with Cypress. @hamstu, how are we with it? 
https://buffer.atlassian.net/browse/PUB-3166

Screenshot 2020-12-02 at 19 42 22

https://share.buffer.com/eDuw9O95

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/composer-editor/components/Editor/index.jsx
+++ b/packages/composer-editor/components/Editor/index.jsx
@@ -1,13 +1,12 @@
 import React, { useMemo, useCallback, useState } from 'react';
 import { Text, createEditor } from 'slate';
 import { withHistory } from 'slate-history';
-import { Slate, Editable, withReact } from 'slate-react';
+import { Slate, withReact } from 'slate-react';
+import twitterText from 'twitter-text';
 
-import styled from 'styled-components';
-import { grayDarker } from '@bufferapp/ui/style/colors';
-import { fontSize, fontFamily } from '@bufferapp/ui/style/fonts';
-
-import Leaf from '../Leaf';
+import Editor from './styles';
+import renderLeaf from '../Leaf';
+import renderElement from '../Element';
 import EmojiPicker from '../plugins/EmojiPicker';
 
 const initialValue = [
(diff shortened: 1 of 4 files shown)


packages/composer-editor/components/Editor/index.jsx
```diff
@@ -51,10 +31,50 @@ const ComposerEditor = () => {
...
+      const charLimit = 20; // Temporary static data
+      const { text } = node;
+
+      if (!Text.isText(node)) {
+        return ranges;
+      }
+
+      // Add text decorator to highlight words with links, hashtags and mentions
+      const linksArray = twitterText.extractUrlsWithIndices(text) || [];
+      const hashtagsArray = twitterText.extractHashtagsWithIndices(text) || [];
```
anafilipadealmeida, line 46: I had a regex for hashtags and mentions first, but ended up using the twitter-text library instead, as it's more accurate and the output is the same as with links, which means we can use the same function !

packages/composer-editor/components/Editor/index.jsx
```diff
@@ -51,10 +31,50 @@ const ComposerEditor = () => {
    */
   const editor = useMemo(() => withHistory(withReact(createEditor())), []);
 
-  /**
-   * Setup custom renderer for elements
-   */
-  const renderElement = useCallback(props => <Element {...props} />, []);
```
anafilipadealmeida, line 57: I've tried to understand why we were using the useCallback hook here (slate has examples with it but I couldn't see any explanation in their docs), especially when the array of dependencies is empty, which means this would always return the same memoized data.
Here's my best guess so far: perhaps they use the useCallback in these render functions to avoid firing unnecessary side effects and renders in their parent Editable component, to which we pass renderElement, renderLeaf and decorate.
If we declare both renderElement and renderLeaf outside the ComposerEditor context, we already guarantee that they will keep the same references and don't cause any re-renders, so we don't need the useCallback.
That leaves us with the useCallBack only in the decorate 🙂

albennett:
Ok, I know I said I'd review this today but I may push it to tomorrow since you'll be on holiday.

packages/composer-editor/components/Editor/index.jsx
albennett: Great job thinking through this @anafilipadealmeida . Your interpretation makes sense to me. We could potentially look closer into the performance of the editor with our dev tools if something feels off.

albennett (approved):
Looks great @anafilipadealmeida Thanks for walking us through this code and some of the nuances during code critique.

PR #1673, opened 2021-01-21
[PUB-3204] Redirect channel management from Org Admin to Core Channels

albennett:
## Description
Redirect channel management from Org Admin to Core Channels
https://buffer.atlassian.net/browse/PUB-3204

## Context & Notes
There are quite a few places where we redirect to connect/reconnect/manage our channels. I'm going to split these publish changes into two PRs. The next PR will cover more connect urls. 

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/globalAccount/reducer.js
+++ b/packages/globalAccount/reducer.js
@@ -2,14 +2,18 @@ import { actionTypes as asyncDataFetchActionTypes } from '@bufferapp/async-data-
 
 export default (state = {}, action) => {
   switch (action.type) {
-    case `globalAccount_${asyncDataFetchActionTypes.FETCH_SUCCESS}`:
+    case `globalAccount_${asyncDataFetchActionTypes.FETCH_SUCCESS}`: {
+      const featureFlips = action.result.featureFlips || [];
       return {
         ...state,
         email: action.result.email,
         _id: action.result._id,
         productSolutionName: action.result.productSolutionName,
-        featureFlips: action.result.featureFlips || [],
+        shouldRedirectToAccountChannels: featureFlips.includes(
+          'sharedChannels'
+        ),
       };
+    }
     default:
       return state;
   }
--- a/packages/globalAccount/reducer.js
+++ b/packages/globalAccount/reducer.js
@@ -8,6 +8,7 @@ export default (state = {}, action) => {
         email: action.result.email,
         _id: action.result._id,
         productSolutionName: action.result.productSolutionName,
+        featureFlips: action.result.featureFlips || [],
       };
     default:
       return state;
--- a/packages/globalAccount/reducer.js
+++ b/packages/globalAccount/reducer.js
@@ -1,6 +1,10 @@
 import { actionTypes as asyncDataFetchActionTypes } from '@bufferapp/async-data-fetch';
 
-export default (state = {}, action) => {
+const initialState = {
+  shouldRedirectToAccountChannels: false,
+};
+
+export default (state = initialState, action) => {
   switch (action.type) {
     case `globalAccount_${asyncDataFetchActionTypes.FETCH_SUCCESS}`: {
       const featureFlips = action.result.featureFlips || [];
--- a/packages/onboarding/middleware.js
+++ b/packages/onboarding/middleware.js
@@ -2,7 +2,7 @@ import { actions as dataFetchActions } from '@bufferapp/async-data-fetch';
 import { getURL } from '@bufferapp/publish-server/formatters/src';
 import { actionTypes } from './reducer';
 
-export default ({ dispatch }) => next => action => {
+export default ({ dispatch, getState }) => next => action => {
   next(action);
   switch (action.type) {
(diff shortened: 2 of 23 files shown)


packages/onboarding/middleware.js
```diff
@@ -16,9 +16,14 @@ export default ({ dispatch }) => next => action => {
         })
       );
       break;
-    case actionTypes.MANAGE_SOCIAL_ACCOUNT:
-      window.location = getURL.getManageSocialAccountURL();
+    case actionTypes.MANAGE_SOCIAL_ACCOUNT: {
```
albennett, line 19: Hey @hamstu and @anafilipadealmeida I'm keen to get your opinion on where we call `window.location`. 😄 While working on this task, I found we take two approaches in publish around redirecting a url (that's not in-app). We either redirect in the index method or we dispatch an action and redirect in the middleware. My preference is to keep it in the index file if it's just a redirect vs adding changes in more files to add it to the middleware. If it's more logic heavy, like dispatching a notification or a data fetch action, then I add to the middleware.

I'm curious about your thoughts!

anafilipadealmeida: To be honest I'd go for the index method as well and reduce the amount of actions when all we need is to redirect the user. That's what I've been doing lately 🙂

packages/profile-sidebar/index.test.jsx
```diff
@@ -1,19 +1,69 @@
...
+import ProfileSidebar, {
+  reducer,
+  actions,
+  actionTypes,
+  middleware,
+} from './index';
 
 describe('ProfileSidebar', () => {
-  it('should export reducer', () => {
+  test('should export reducer', () => {
```
anafilipadealmeida, line 16: It's been a while so I'm not sure anymore, did we decide to move forward at some point with `test` instead of `it`?

packages/globalAccount/reducer.js
```diff
@@ -2,13 +2,18 @@ import { actionTypes as asyncDataFetchActionTypes } from '@bufferapp/async-data-
...
   switch (action.type) {
-    case `globalAccount_${asyncDataFetchActionTypes.FETCH_SUCCESS}`:
+    case `globalAccount_${asyncDataFetchActionTypes.FETCH_SUCCESS}`: {
+      const featureFlips = action.result.featureFlips || [];
       return {
         ...state,
         email: action.result.email,
         _id: action.result._id,
         productSolutionName: action.result.productSolutionName,
+        shouldRedirectToAccountChannels: featureFlips.includes(
```
anafilipadealmeida, line 12: Should we have an initial state in globalAccount with `shouldRedirectToAccountChannels` as false, just in case this request fails or takes longer than expected?

anafilipadealmeida (commented):
Hey @albennett, thanks for doing this work! I just have a couple of questions 😃

packages/profile-sidebar/index.test.jsx
albennett: Ah, great question. I thought we were leaning towards the react-testing library going forward, though maybe this hasn't been decided yet 🤔 I can bring it up in retro. Or better yet, the next frontend gathering since there have been discussions around standards. ✨ 

I'll leave it as test for now if that feels ok to you.

anafilipadealmeida: Sure! 🙂 I know in jest they are alias, so both work the same way. In React Testing Library I don't even know. But sounds good to decide this and add to notion!

packages/globalAccount/reducer.js
albennett: It technically should still be falsy as null, though I think it's great to be more explicit with an initialState. I'll add this change, thanks Ana.

packages/onboarding/middleware.js
hamstu: I'm all for simplicity, so +1 to that approach! If the only purpose of a `dispatch` is to run some small side-effect that we could reasonably do inline, then let's do inline instead. 💯

anafilipadealmeida (approved):

packages/profile-sidebar/index.test.jsx
hamstu: React Testing Library doesn't affect or provide `test` or `it` since it's just helpers we use with the Jest, [which provides both of those](https://jestjs.io/docs/en/api.html#testname-fn-timeout). (And yep they're the exact same method 👌🏻 as Ana said.)

So in the end it's just a style choice, and IMO it doesn't really matter. Though I suppose one could argue for consistency if you wanted to be really pedantic. 😂

packages/profiles-disconnected-modal/index.test.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react';
```
hamstu, line 1: Should this be *.jsx just for consistency? Since it has React code.

hamstu (approved):
Great stuff @albennett! LGTM 🚀

PR #1675, opened 2021-01-26
CORE-1366 | Deprecate channels drag and drop

MayaUribe:
## Description
Deprecate channels drag and drop and prevent displaying locked channels in the profile sidebar.

## Context & Notes

CORE-1366: This work is related to Shared Channels.

## Checklist

#### Staging Deployment


packages/profile-sidebar/reducer.js
```diff
@@ -49,63 +48,6 @@ const moveProfileInArray = (arr, from, to) => {
...
-  const { profileLimit, hoverIndex, dragIndex } = action;
-  const reorderedProfiles = moveProfileInArray(profiles, dragIndex, hoverIndex);
-  /* The reducer will return an object with 3 properties, each of them an array of profiles.
-  For each profile reduced, we will need to spread the ACC object,
-  changing only 1 property, i.e., adding the profile to only 1 array */
-  const {
-    enabledProfiles,
-    lockedProfiles,
-    blockedProfiles,
-  } = reorderedProfiles.reduce(
```
anafilipadealmeida, line 67: I believe this was the first array reduce I wrote in Publish 😂

karelm (approved):
LGTM 🚀

anafilipadealmeida (commented):
Hey @MayaUribe, those changes look good to me! 
I have a couple of questions now that we'll no longer display locked profiles in the sidebar: 
- do we need to adjust our profiles array (perhaps filter the profiles array in the sidebar reducer instead) or adjust our routes, to make sure we handle the navigation to a profile id of a locked profile in a graceful way?
- Did we already log in a ticket a task to remove the locked-profile-notification component?
- What will happen if the owner of an org downgrades the plan, since we no longer display the queue with the locked profiles? What will contributors see?

albennett:
Great job @MayaUribe 🙌 
❓ to get more clarity: 
- Will these changes be released once Shared Channels is 100% rolled out?
- If a user downgrades and goes over the limit, will we be removing the channels that were most recently added? 
- Will we just "hide" these channels or are we removing them (and updates etc) from the db?

MayaUribe:
Hey! @anafilipadealmeida thank you for the questions!
> * do we need to adjust our profiles array (perhaps filter the profiles array in the sidebar reducer instead)

Yes, we should adjust the profiles array in the reducer (I might have missed removing it there 😅)

> or adjust our routes, to make sure we handle the navigation to a profile id of a locked profile in a graceful way?

Because at the moment we are displaying the locked notification component I didn't think of it as a blocker for the moment, but I'm thinking that perhaps we can create a card to redirect to an unlocked profile instead, so it's not confusing for users?


> * Did we already log in a ticket a task to remove the locked-profile-notification component?

Good one! we should create a ticket to remove this component 👍

> * What will happen if the owner of an org downgrades the plan, since we no longer display the queue with the locked profiles? What will contributors see?

Good question, are you thinking of a particular scenario here?
The sidebar won't display the locked profiles, and if we do something as I mentioned above, we would redirect to an enabled profile?
I'm thinking what this should look like if that was the only profile that the user had access to before the downgrade 🤔

MayaUribe:
Thank you for your questions @albennett 😊

> * Will these changes be released once Shared Channels is 100% rolled out?

Yes! ideally, we want to release these changes only after we have the lock/unlock functionality in core.

> * Will we just "hide" these channels or will we eventually remove them (and updates etc) from the db?

At the moment, we are only hiding them, we are aiming for a "quick" approach first, but I think it makes sense to think about removing these from the db in the future, I can confirm this!

MayaUribe:
Thank you so much for your comments @anafilipadealmeida @albennett I added a few comments so that I can confirm with Core and give you a better response to your questions 🤗

hamstu (approved):
Ah, I love the smell of deleted code in the morning. ☕ 🗑️  


LGTM! 🚀

albennett (approved):
This looks great Maya! 🚀 🧹

packages/profile-sidebar/components/ProfileList/index.jsx
```diff
@@ -1,27 +1,23 @@
...
+    <ProfileListItem
       avatarUrl={profile.avatarUrl}
       type={profile.type}
       handle={profile.handle}
       pendingCount={profile.pendingCount}
       selected={profile.id === selectedProfileId}
-      locked={profile.disabled}
       disconnected={profile.isDisconnected}
       onClick={() => onProfileClick(profile)}
       profileLimit={profileLimit}
```
albennett, line 20: I think we can remove `profileLimit` here, too, as it was a part of the drag n drop logic 💅

MayaUribe:
Heya @msanroman! 👋
I'm adding you to this PR, because of what we discussed during the standup, on hiding this deprecation behind a feature flip, I'm thinking that if I go for the feature flip approach I might have to create a separate PR and leave this one to merge after as a cleanup work since this one deletes bit of code 😅

anafilipadealmeida:
Hey @MayaUribe

If you filter the profiles in the profileSidebar reducer, the profileSidebar state will store it. That will be enough to handle navigation, as we use the profileSidebar state for it (https://github.com/bufferapp/buffer-publish/blob/main/packages/app-pages/index.js#L17). Meaning that if you navigate to the url of a locked profile, you'll be redirected to another url, as if the channel doesn't exist (either to a default page or other channel, depending on the situation).

> What will happen if the owner of an org downgrades the plan, since we no longer display the queue with the locked profiles? What will contributors see?
> The sidebar won't display the locked profiles, and if we do something as I mentioned above, we would redirect to an enabled profile?

With the org switcher, if the owner downgrades, nobody will have enabled profiles in that org. If profiles are already filtered in the state.profileSidebar, the users will see the `DefaultPage`. Currently, non-owners see 
image
. My suggestion would be to create a new copy for when we are in the case of downgrade.

MayaUribe:
> If you filter the profiles in the profileSidebar reducer, the profileSidebar state will store it. That will be enough to handle navigation, as we use the profileSidebar state for it 

Oh yes, great point! I'll make the adjustments in the reducer 🙌
 
> > What will happen if the owner of an org downgrades the plan, since we no longer display the queue with the locked profiles? What will contributors see?
> > The sidebar won't display the locked profiles, and if we do something as I mentioned above, we would redirect to an enabled profile?
> 
> With the org switcher, if the owner downgrades, nobody will have enabled profiles in that org. If profiles are already filtered in the state.profileSidebar, the users will see the `DefaultPage`. Currently, non-owners see
> 
> image
> 
> and owners see
> image
> My suggestion would be to create a new copy for when we are in the case of downgrade.

That sounds like a good suggestion! We should create a task to add a new copy for this particular scenario.

PR #1684, opened 2021-02-03
CORE-1366 | Hide drag and drop and locked channels under a feature flip.

MayaUribe:
## Description
Deprecate channels drag and drop and prevent displaying locked channels in the profile sidebar.
This PR hides the drag and drop and locked channels for an account with the shared channels feature flip. (Related PR: https://github.com/bufferapp/buffer-publish/pull/1675)

⚠️ **This PR hides the logic under a Feature Flip so we can deploy this now, and the other PR if for the cleanup work.**

## Notes to Reviewers
❓ I have a question regarding the channel limits, specifically for the `ProfileConnectShortcut` component, it feels like a product question, but I'm hoping to get some opinions

**For users with shared channels feature flip, what should be the logic here?:**
As you know, we have these 3 buttons in the profile sidebar as shortcuts to connect one of these three types of channels: 

With the work on buffer account to connect channels, I'm starting to think, these buttons don't make a lot of sense anymore, since the logic to display them is related to the channel limits, and the channel limits is something we will be handling from core side, so there are two options here:
- Redirect to buffer account to connect a channel (would be exactly the same route as in the `Manage Channels` button, which is where I'm not sure if it makes sense to have the same link (maybe it does?).
- Remove these three buttons since we won't redirect to connect to those networks directly.

Tagging you here to know your thoughts @karelm @albennett @anafilipadealmeida @hamstu @letsbleachthis

## Context & Notes

**CORE-1366:** This work is related to Shared Channels:
- Deprecate the drag and drop functionality from the profile's sidebar
- Only display enabled channels

## Next Steps
- After the rollout, we can merge this related PR https://github.com/bufferapp/buffer-publish/pull/1675 to clean up the unused code: profileLimit, ProfileDragWrapper, logic for locked profiles.

## Checklist

#### Staging Deployment


karelm (approved):
LGTM 🚀

albennett:
> With the work on buffer account to connect channels, I'm starting to think, these buttons don't make a lot of sense anymore, since the logic to display them is related to the channel limits, and the channel limits is something we will be handling from core side

Great thoughts, Maya! I think of the buttons as more of "discovery" for connecting channels. It's not based solely on account limits but also whether a user has connected specific social accounts yet. I agree, it does feel a bit strange to redirect vs connecting in-app (maybe we can rethink the UI?). It does sound like Core will eventually create an endpoint for direct connections (I asked about it in the redirect jira task).

hamstu:
@MayaUribe I do think this is a good product question. I would encourage you to ask @letsbleachthis in Slack to get her thoughts (she's understandably not in GitHub much, but you can just share this PR directly). 

My opinion is if we can't link directly to OAuth flows from those buttons then we should probably just remove them. I think they are helpful for encouraging users to connect accounts so perhaps there's an argument to be made for keeping them, but it's not a great experience if they don't do exactly what they say. 🤷🏻

letsbleachthis:
Hey y'all, thanks for the ping! Good catch, @MayaUribe. I agree with @hamstu's sentiment:

> My opinion is if we can't link directly to OAuth flows from those buttons then we should probably just remove them. I think they are helpful for encouraging users to connect accounts so perhaps there's an argument to be made for keeping them, but it's not a great experience if they don't do exactly what they say.

Clicking them to be redirected to the /connect page for shared channels wouldn't be an ideal experience. Plus, now that we're focusing more broadly on all types of SMBs, these promoted options (IG, Twitter, FB) feel slightly off. They're still our most popular channel types, but they feel more aligned with a DTC strategy. I vote we remove them 👍

MayaUribe:
> Hey y'all, thanks for the ping! Good catch, @MayaUribe. I agree with @hamstu's sentiment:
> 
> > My opinion is if we can't link directly to OAuth flows from those buttons then we should probably just remove them. I think they are helpful for encouraging users to connect accounts so perhaps there's an argument to be made for keeping them, but it's not a great experience if they don't do exactly what they say.
> 
> Clicking them to be redirected to the /connect page for shared channels wouldn't be an ideal experience. Plus, now that we're focusing more broadly on all types of SMBs, these promoted options (IG, Twitter, FB) feel slightly off. They're still our most popular channel types, but they feel more aligned with a DTC strategy. I vote we remove them 👍

Sounds great! thank you all for your comments and thoughts ❤️ I think we can create a card to remove these buttons in a different PR then, the thing to keep in mind is that these 3 buttons are not only displayed in the profile sidebar but are also displayed in a couple more of places (@anafilipadealmeida pointed that out today) so we might want to be aware of that

packages/profile-sidebar/index.js
```diff
@@ -19,14 +19,25 @@ export default hot(
   connect(
     (state, ownProps) => {
       const cta = 'publish-app-sidebar-addProfile-1';
-      const { shouldRedirectToAccountChannels } = state.globalAccount;
+      const {
+        shouldRedirectToAccountChannels,
```
albennett, line 23: 💭 We could rethink the naming of this if we keep needing to use it for different reasons.

packages/profile-sidebar/reducer.js
```diff
@@ -160,15 +163,16 @@ export default (state = initialState, action) => {
       };
 
     case `profiles_${dataFetchActionTypes.FETCH_SUCCESS}`: {
+      const profileList = getEnabledProfiles(action.result);
```
albennett, line 166: Hmm 🤔 does this need to be under the feature flip, too?

albennett:
> the thing to keep in mind is that these 3 buttons are not only displayed in the profile sidebar but are also displayed in a couple more of places
I know we reference the ProfileSidebar in the Disabled Queue page (which has these buttons). Are there others? 🤔  

Thanks Maya for bringing this up! 🌮

packages/profile-sidebar/reducer.js
MayaUribe: Oh my yes! Thank you for pointing it out 😂

anafilipadealmeida (commented):
Hey @MayaUribe! All the changes seem good to me, my only question here is: to whom are we giving the feature flip during the rollout? I don't feel comfortable rolling that out to certain users without addressing the issue first 👇 

I used my team members account (the owner downgraded) and the message is wrong. As I stated in the [clean up PR](https://github.com/bufferapp/buffer-publish/pull/1675), we should have a different copy for this scenario, as this as it is will create confusion for users and advocates.
Screenshot 2021-02-15 at 10 27 01

MayaUribe:
> Hey @MayaUribe! All the changes seem good to me, my only question here is: to whom are we giving the feature flip during the rollout? I don't feel comfortable rolling that out to certain users without addressing the issue first 👇
> 
> I used my team members account (the owner downgraded) and the message is wrong. As I stated in the [clean up PR](https://github.com/bufferapp/buffer-publish/pull/1675), we should have a different copy for this scenario, as this as it is will create confusion for users and advocates.
> Screenshot 2021-02-15 at 10 27 01

Hey @anafilipadealmeida I realize your concern here, but as we start the rollout of shared channels, we don't want users to be able to reorder or have the ability to lock/unlock profiles from Publish side, I agree that this error is misleading and I'd suggest we create a Jira Card and address it in a different PR, what do you think?

anafilipadealmeida (approved):
Cool @MayaUribe, if this is only be rolled out to 5% of users this week I'm good with it! I'll create a new ticket, I talked to Dianne to grab this one so you can finish what you're doing and come back to the team faster 😊

PR #1651, opened 2020-12-14
[PUB-3118] Add org id check to analytics middleware

chayjay:
## Description

This change adds check for organization selected to analytics middleware to ensure that org IDs are tracked in all events.

## Context & Notes
PUB-3118

## Checklist

-   [x] My code follows the code style and guidelines of this project. 
-   [x] I have added tests to cover my changes.
-   [x] All new and existing tests passed.
-   [ ] I have tested different user stories. 
-   [x] I have identified similar or related features and tested they are still working.
-   [x] I have performed a self-review of my own code.
-   [ ] I have tested my changes/additions in the latest Chrome, Firefox, and Safari.
-   [ ] I have commented my code, particularly in hard-to-understand areas.
-   [ ] I kept accessibility in mind by following the [a11y checklist.](https://www.notion.so/buffer/Workflow-Checklist-e64d86eb795140bcbfdc16d1c72e573f)
-   [ ] I have considered [security, abuse & compliance](https://www.notion.so/buffer/Engineering-Wiki-f34142d290304c35bebadf76cc9cc89e#cc6dcc7617184227b77da2e1b262a563) implications of my changes.

#### Staging Deployment

To visit the URL of the staging deployment, please click "Show all checks" at the bottom of this PR and click "details" next to `bufferbotbrains/cicd-buffer-publish-legacy`

PR #1693, opened 2021-02-10
[EN-469] Make sure 'organizationId' is included on all tracked events

ssvmvss:
## Description

This PR has the changes needed to: 

**- Make sure that the correct `organizationId` value in included as part of all tracked events:**
There were a couple of events that would send an `organizationId` value as part of their payload (like `Campaign Deleted` or `Campaign Report Viewed`) which on some scenarios was `null` or `undefined`. Since our tracking function would build the tracking payload this way:

```
{
   organizationId,
   ...action.payload
}
```

The `organizationId` value that was correctly set was immediately overriden by sometimes wrong values if present on the action payload.

**- Make sure that the `eventQueue` is cleared after all present messages were sent:**
When an event cannot be tracked because there is no organization selected yet, we save that event on an `eventQueue` array to wait until they can be sent. But the queue was not cleared after events were processed so they were re-sent every time a new organization was selected.

**PS:** If you check the commit history you will see me adding some `console.logs` here and there and then removing them. I did that because I don't have Publish running locally and I had no idea `analytics.debug()` existed! 🙈 

## Context & Notes

Jira ticket: https://buffer.atlassian.net/browse/EN-469

## Checklist

#### Staging Deployment


packages/analytics-middleware/middleware.js
```diff
@@ -14,28 +14,30 @@ export default ({ dispatch, getState }) => next => action => {
...
       const orgId = getState().organizations.selected?.globalOrgId;
-      if (!orgId) {
-        eventQueue.push(action);
-      }
-      if (orgId && window.analytics) {
-        window.analytics.track(action.eventName, {
+
+      if (orgId) {
+        window.analytics && window.analytics.track(action.eventName, {
+          ...action.payload,
```
ssvmvss, line 20: Copy the event payload first, as a base, and then make sure we complete the data with `product`, `clientName` and `organizationId`.

packages/analytics-middleware/middleware.js
```diff
@@ -14,28 +14,30 @@ export default ({ dispatch, getState }) => next => action => {
...
       eventQueue.forEach(event => {
         dispatch(
           actions.trackEvent(event.eventName, {
-            ...event.payload,
+            ...event.payload
           })
         );
       });
+
+      eventQueue = [];
```
ssvmvss, line 40: Make sure we clear the array after processing all comments!

packages/campaign/middleware.js
```diff
@@ -30,11 +30,9 @@ export default ({ dispatch, getState }) => next => action => {
 
     case actionTypes.GO_TO_ANALYZE_REPORT: {
       const { id, name } = action.campaign;
-      const { organizationId } = state.profileSidebar.selectedProfile;
       const metadata = {
```
ssvmvss, line 33: Removed this from both `Campaign Report Viewed` and `Campaign Deleted` since we don't need it anymore! Also, the value that we were getting from `profileSidebar` was `null` or `undefined` sometimes.

albennett (approved):
Hey @ssvmvss thanks for helping us investigate and fix these issues. 🙌 Can you remind me again why we need an eventQueue? 

also, @hamstu does this solve the issue you brought up the other day around the work Char did on tracking?

ssvmvss:
Heya @albennett! 👋 Great question! I did not add the `eventQueue` myself so I might be wrong but as far as I could see [here](https://github.com/bufferapp/buffer-publish/pull/1651) it was added because the router's `LOCATION_CHANGE` action is fired the first time **before** there is an organization selected. So, as far as I understand that queue was added as a buffer, to keep the events saved until an organization is selected and then they can be sent.

I think @hamstu might have more context about that! 🙂

hamstu (approved):
This looks great @ssvmvss! Thanks for working on this.

And yep, it solves the issue which I had suspected around the eventQueue not being cleared. (I didn't have time to dig in too much but I suspected it was a bug.) Great catch 💯 

LGTM! 🚀

ssvmvss:
Perfect! I'll merge this tomorrow and I'll start monitoring 🙂 Thanks so much @albennett and @hamstu! ❤️ 🌮

hamstu:
@ssvmvss One thing I noticed after this was merged is that it doesn't line up with our Prettier/ESLint settings (so there's some style warnings in `main` now).

image

No worries, since it's easy to miss if your editor isn't configured for either of those! Just bringing it up for the future! (And this makes me think we should all really get better at standardizing these things soon 😅 

Thanks again for working on this!

PR #1708, opened 2021-03-02
Revert "Revert "[PUB-3200] Implement global appshell""

anafilipadealmeida:
Reverts bufferapp/buffer-publish#1707
Final 🤞 implementation of the org switcher. We've asked advocates to try our staging branch this time, to make sure there are no weird edge cases we might be missing 😅


packages/app-pages/components/AppPages/index.jsx
```diff
@@ -1,45 +1,70 @@
...
+  const switchOrganization = useOrgSwitcher();
+
   // If org coming from route doesn't match the last org stored, select and store the new value
   useEffect(() => {
-    if (needsToSetCurrentOrg) {
-      setCurrentOrganization(currentOrgId);
+    if (needsToSelectNewOrgInAppShell) {
+      switchOrganization(currentOrgId);
     }
+    storeSelectedOrg(currentOrgId);
```
anafilipadealmeida, line 54: Only difference from the previous approach: just dispatching an event to update the redux store. This seems to be enough to solve our state issues with orgs

albennett (approved):
🚀 🙌  🤞 

I tried testing in staging but it looks like the staging url may have been reset (default backend - 404) 🤔

PR #1619, opened 2020-11-03
[PUB-2477] Add pinterest link as source url

anafilipadealmeida:
## Description
- When user creates a Pinterest post with links on the description and no source url, we pick the first link as the source url

## Context & Notes
Pinterest doesn't post the text description if it contains links and no source url is defined
I'm using twitter-text to get urls, as it's the same we use to highlight links in the composer text
https://buffer.atlassian.net/browse/PUB-2477

## Checklist

-   [ ] My code follows the code style and guidelines of this project. 
-   [ ] I have added tests to cover my changes.
-   [ ] All new and existing tests passed.
-   [ ] I have tested different user stories. 
-   [ ] I have identified similar or related features and tested they are still working.
-   [ ] I have performed a self-review of my own code.
-   [ ] I have tested my changes/additions in the latest Chrome, Firefox, and Safari.
-   [ ] I have commented my code, particularly in hard-to-understand areas.
-   [ ] I kept accessibility in mind by following the [a11y checklist.](https://www.notion.so/buffer/Workflow-Checklist-e64d86eb795140bcbfdc16d1c72e573f)
-   [ ] I have considered [security, abuse & compliance](https://www.notion.so/buffer/Engineering-Wiki-f34142d290304c35bebadf76cc9cc89e#cc6dcc7617184227b77da2e1b262a563) implications of my changes.

#### Staging Deployment

To visit the URL of the staging deployment, please click "Show all checks" at the bottom of this PR and click "details" next to `bufferbotbrains/cicd-buffer-publish-legacy`

PR #1710, opened 2021-03-03
[PUB-3167] Add validation to Pinterest source url

anafilipadealmeida:
## Description
- Undo https://github.com/bufferapp/buffer-publish/pull/1619, that automatically picks the first link as the source url when Pinterest posts have links.
- Make source field mandatory for Pinterest when there are links in the post.
- Since Pinterest marks most url shorteners as spam, block users from inserting buff.ly and bit.ly source urls.
- **NEW:** Unshorten Pinterest links when moving from omni to individual composer

## Context & Notes
Pinterest blocks most url shorteners. Our Pinterest profiles don't have the option to select link shorteners in their settings. If users are composing to multiple profiles at the same time, and one of them has link shortening available, all the profiles including Pinterest will get it in the frontend - we then remove the link shortener for Pinterest in the backend.
https://buffer.atlassian.net/browse/PUB-3167

Screenshot 2021-03-03 at 16 38 24

## Checklist

#### Staging Deployment

### DIFF
--- a/packages/composer/composer/stores/ComposerStore.js
+++ b/packages/composer/composer/stores/ComposerStore.js
@@ -251,7 +251,8 @@ const ComposerStore = {
         const isPinterestSourceUrlUsingLinkShortener =
           isPinterest &&
           (draft.sourceLink?.url.includes('bit.ly/') ||
-            draft.sourceLink?.url.includes('buff.ly/'));
+            draft.sourceLink?.url.includes('buff.ly/') ||
+            draft.sourceLink?.url.includes('j.mp/'));
 
         const contentText = contentState.getPlainText();
         const linksInText = twitterText.extractUrls(contentText);
--- a/packages/composer/composer/stores/ComposerStore.js
+++ b/packages/composer/composer/stores/ComposerStore.js
@@ -247,6 +247,18 @@ const ComposerStore = {
           draft.sourceLink === null ||
           twitterText.isValidUrl(draft.sourceLink.url, true, false);
 
+        const isPinterest = draft.service.name === 'pinterest';
+        const isPinterestSourceUrlUsingLinkShortener =
+          isPinterest &&
+          (draft.sourceLink?.url.includes('bit.ly') ||
+            draft.sourceLink?.url.includes('buff.ly'));
+
+        const contentText = contentState.getPlainText();
+        const linksInText = twitterText.extractUrls(contentText);
+
+        const hasPinterestLinkWithoutSourceUrl =
+          isPinterest && linksInText.length > 0 && !draft.sourceLink;
+
         const hasRequiredText = !draft.service.requiresText || hasText;
 
         let validationResultVideo = new ValidationSuccess();
@@ -269,7 +281,9 @@ const ComposerStore = {
           !hasRequiredText ||
           !isSourceUrlUnrequiredOrValid ||
           validationResultVideo.isInvalid() ||
-          validationResults.isInvalid();
+          validationResults.isInvalid() ||
+          hasPinterestLinkWithoutSourceUrl ||
+          isPinterestSourceUrlUsingLinkShortener;
 
         if (isInvalid) {
           let messages = [];
@@ -317,9 +331,13 @@ const ComposerStore = {
             messages.push('Please include at least some text or an attachment');
           }
 
-          if (!isSourceUrlUnrequiredOrValid) {
+          if (hasPinterestLinkWithoutSourceUrl) {
+            messages.push(`Please include a source url`);
+          } else if (!isSourceUrlUnrequiredOrValid) {
+            messages.push('Please include a valid source url');
+          } else if (isPinterestSourceUrlUsingLinkShortener) {
             messages.push(
-              'Please include a valid source url or leave the source url blank'
+              'Please include a source url without a link shortener'
             );
           }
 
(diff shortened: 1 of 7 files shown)


packages/composer/composer/stores/ComposerStore.js
```diff
@@ -247,6 +247,18 @@ const ComposerStore = {
           draft.sourceLink === null ||
           twitterText.isValidUrl(draft.sourceLink.url, true, false);
 
+        const isPinterest = draft.service.name === 'pinterest';
+        const isPinterestSourceUrlUsingLinkShortener =
+          isPinterest &&
+          (draft.sourceLink?.url.includes('bit.ly/') ||
```
albennett, line 253: 💭 There are other less common link shorteners. Does Pinterest document which link shorteners they check? Maybe other link shorteners are a small enough edge case that it's ok if it fails.

albennett, line 253: I notice `j.mp` as an option in our link shortener. Should we check for it, too?

packages/composer/composer/stores/ComposerStore.js
```diff
@@ -317,10 +331,15 @@ const ComposerStore = {
...
           }
 
-          if (!isSourceUrlUnrequiredOrValid) {
+          if (
+            hasPinterestLinkWithoutSourceUrl ||
+            isPinterestSourceUrlUsingLinkShortener
+          ) {
             messages.push(
-              'Please include a valid source url or leave the source url blank'
+              `Please include a source url without a link shortener`
```
albennett, line 339: 💭 We may want to think about matching our language more with pinterest. For example, they call "source url" a "destination link". A nomenclature audit could be a good one to bring up to product.

packages/composer/composer/utils/WebAPIUtils.js
```diff
@@ -800,28 +799,13 @@ function getFormattedAPIData(serviceName, unformattedData) {
...
+        let sourceUrl = sourceLink.url;
+        if (sourceUrl.indexOf('http') !== 0) sourceUrl = `http://${sourceUrl}`;
 
-      // If Pinterest has text description with links and without a source url,
-      // we need to define one of the links as a source to be able to post the description
-      if (hasPinterestWithoutSourceUrl) {
-        const linksInText = twitterText.extractUrls(serviceDraftText);
-        if (linksInText.length > 0) {
-          sourceUrl = linksInText[0];
-        }
```
hamstu, line 817: Re: main comment; wondering if we can somehow keep this behavior if we can properly unshorten the first URL. Which should work unless the user shortened it themselves, but we could catch that in the validation step earlier.

hamstu (changes requested):
Hey @anafilipadealmeida! 

I'm wondering if we considered using the unshortened version of the link as the source URL? We can call [`ComposerStore.getCanonicalUrl(url)`](https://github.com/bufferapp/buffer-publish/blob/0dc80d2bb7dfeb2e87c70421e3086d03b73fc749/packages/composer/composer/stores/ComposerStore.js#L377-L384) to get the original URL if it was shortened during the composer session.

Left some more comments inline! Let me know if I can clarify or if I'm missing any other context here.

anafilipadealmeida:
Great thoughts in here @albennett and @hamstu !

anafilipadealmeida:
Great thoughts in here @albennett and @hamstu !

anafilipadealmeida:
Great thoughts in here @albennett and @hamstu! 
1. "destination URL" makes a lot of sense! I had no idea the names didn't match Pinterest vocabulary 😅 I did a quick search in our help center and there are 2 articles mentioning it. Mobile is also using source url.  I'll leave a note to let advocates and mobile know about this!
2. Pinterest is not very clear about which link shorteners they mark as spam (at least bit.ly is). We are not sure if users with custom domains see their shortened links marked as spam, and we don't have an easy way to identify all those cases in the code, so the idea was to flag the most used ones in Buffer and monitor from there.
3. I'm not a fan of the previous fix I did for the destination URL and I'd prefer not to keep the logic. The reason why is because it doesn't seem right to me to have us filling fields automatically for users, especially when those fields are visible in the UI and there's no indication that we are doing it and why.
4. The main source of confusion with link shortening is the fact users can see links being shortened in our UI for some Pinterest profiles (when they used the multi-channel composer functionality), even though we don't allow it and end up undoing it in the backend. Our frontend should be a reflection of the end result. Let me see if we could use the `ComposerStore.getCanonicalUrl(url)` there. It's not the scope of this bug but it would improve the experience a lot.
5. 
> I also think the message shouldn't say 'without a link shortener' unless the user has specifically typed/pasted one in themselves that is actually using a link shortener.

This is a great topic! My first commit had a different message depending on whether you had typed a link with link shortener or not. However Dianne flagged that telling users in advance that there's the limitation might reduce friction. I can totally understand that side, as first we tell them to insert a link under certain circumstances, they do it, and then we tell them that the link shortened is not valid. However, if we can make the adjustments I mentioned in 4., the number of users inserting shortened urls would decrease a lot (as most of them just copy the link they inserted in the description of the post) so I'd be happy to separate the error messages!

Thank you so much for all this ❤️

hamstu:
Thanks for the great response @anafilipadealmeida! Let me know how it goes with trying to improve the UX with `getCanonicalUrl` — I totally agree that it could likely improve things for the vast majority of cases.

hamstu (approved):
This is fantastic @anafilipadealmeida! Thanks for taking all the feedback/thoughts so graciously and working hard to improve the UX here. ❤️ 

LGTM! 🚀

PR #1716, opened 2021-03-09
[PUB-3420] Revised user menu

hamstu:
## Checklist

#### Staging Deployment


packages/app-shell/index.js
```diff
@@ -55,8 +59,9 @@ export default connect(
...
       dispatch(organization.goTo({ orgId: organizationId }));
     },
+    manageChannels() {},
   })
 )(AppShell);
 
-export reducer from './reducer';
-export middleware from './middleware';
+export { default as reducer } from './reducer';
+export { default as middleware } from './middleware';
```
hamstu, line 67: This has the same effect, but doesn't get flagged as invalid syntax in VS Code. We should progressively migrate our code to this style.

For more background, I think the version on the left is not actually approved in any spec, but is supported in Babel / webpack which is why it builds and runs fine. But I feel better about sticking with the more common and approved pattern.

packages/app-shell/components/userMenuItems/index.jsx
```diff
@@ -1,61 +1,49 @@
...
-  };
-  const extraItems = [];
-  if (showManageTeam) {
-    extraItems.push(userMenuItems.manageTeam);
-  }
-  if (showSwitchPlan) {
-    extraItems.push(userMenuItems.switchPlan);
-  }
-  return [...userMenuItems.top, ...extraItems];
+  ].filter(Boolean);
```
hamstu, line 46: The `.filter(Boolean)` is a handy trick to remove `falsy` elements from the Array. So for example if `showManageTeam` isn't truthy a `false` value will be in the returned array, but this filter removes it.

This can also be written as `.filter(item => item)`, and has the same effect.

albennett: Ah yes, I've used that trick a few times

packages/app-shell/index.js
```diff
@@ -3,51 +3,52 @@ import {
...
-     * Needs organizations and profiles.
-     */
-    canSeeOrgSwitcher: state.organizations.canSeeOrgSwitcher,
-    organizations: getOrgsAlfabeticalOrder(state.organizations.list) || [],
-    selectedOrganizationId: state.organizations.selected?.id,
-    profiles: state.profileSidebar.profileList,
-    isImpersonation: state.appShell.isImpersonation,
-  }),
+  state => {
+    const { shouldRedirectToAccountChannels } = state.globalAccount;
```
albennett, line 16: So Karel recently made a change to get this from the org level instead of account level: https://github.com/bufferapp/buffer-publish/pull/1717/files#diff-4875c3e53264eb807cd85d65f042532b63fbc1fdfdd8b9178aaf7aa88e87e9faR10

We'll need to change it here, too.

packages/app-shell/index.js
```diff
@@ -3,51 +3,52 @@ import {
...
+        state.onboarding.canSeeOnboardingPage &&
+        state.router.location.pathname === newBusinessTrialists.route,
+      enabledProducts: state.appShell.enabledProducts,
+      featureFlips: state.appShell.featureFlips,
+      /**
+       * Org Switcher
+       * Needs organizations and profiles.
+       */
+      canSeeOrgSwitcher: state.organizations.canSeeOrgSwitcher,
+      organizations: getOrgsAlfabeticalOrder(state.organizations.list) || [],
```
albennett, line 39: You may have some merge conflicts depending on who deploys first, as @anafilipadealmeida removes a lot of these props here: https://github.com/bufferapp/buffer-publish/pull/1708/files

packages/app-shell/components/userMenuItems/index.jsx
```diff
@@ -1,61 +1,48 @@
...
       onItemClick: () => {
         window.location.assign(`${getURL.getManageTeamURL()}`);
       },
     },
-    returnToClassic: {
-      id: 'returnToClassic',
-      title: t('app-shell.returnToClassic'),
-      icon: <InvertedReturnIcon />,
+    {
+      id: 'upgrade',
```
albennett, line 37: We show upgrade regardless of their plan type. Could showing upgrade when they're on a business plan cause more potential downgrades 🤔

hamstu: This is a great one! Thank you so much. I think I had planned to do that and then got distracted by the UI changes and forgot. 🙈

PR #1719, opened 2021-03-10
[PUB-3421] Include the free plan in the drafts paywall

albennett:
## Description
Include the free plan in the drafts paywall. 
Includes: 
- Tracking for the free plan paywall action
- Display paywall nav tag on drafts if org is on free or pro plan

https://buffer.atlassian.net/browse/PUB-3421
## Context & Notes
Free & Pro plan: 
Screen Shot 2021-03-15 at 1 12 25 PM

Pro plan with business trial access:
Screen Shot 2021-03-15 at 1 13 59 PM

## Checklist

#### Staging Deployment


packages/drafts/reducer.js
```diff
@@ -92,7 +93,7 @@ const draftReducer = (state, action) => {
 const draftsReducer = (state = {}, action) => {
   switch (action.type) {
     case `draftPosts_${dataFetchActionTypes.FETCH_SUCCESS}`: {
-      const { drafts } = action.result;
+      const { drafts = {} } = action.result;
```
albennett, line 96: I noticed when we run tests, the rpc drafts success runs twice and the second time drafts are null. The tests still pass but an error shows up in the test logs. This issue only happens with tests, so I think it's an issue with the rpc mock. I went ahead and added an error handler for now.

anafilipadealmeida (commented):
Hey @albennett, it might be worth confirming with Lauren, but I thought we were dropping the plan names from any CTAs. In this case, I would see the button always with the copy "Upgrade" and pointing to the plans page (and not the billing page in case of pro plans). Before some of the pro plans were going to the billing page because we only had a switch modal for free users, with the plans page available we don't need it.

packages/constants/index.js
```diff
@@ -75,6 +75,7 @@ const STORIES_PREVIEW_COMPOSER_ADD_NOTE =
   'publish-stories-composerPreview-addNote-1';
 const STORIES_IMAGE_ASPECT_RATIO_UPLOADED =
   'publish-stories-composer-imageAspectRatioUploaded-1';
+const UPGRADE_PATH_DRAFTS_FREE = `publish-drafts-freeUserPaywall-upgrade-1`;
```
anafilipadealmeida, line 78: Not a big deal but looking at the naming of the other variables, it would be more consistent as `DRAFTS_FREE_UPGRADE`

packages/server/parsers/orgParser.js
```diff
@@ -64,7 +64,8 @@ module.exports = orgData => ({
 
   // Upgrade/ Trial Paths
   showUpgradeToProCta: orgData.planBase === 'free',
-  showShowDraftsPaywall: orgData.planBase === 'pro',
+  showProPaywall: orgData.planBase === 'pro',
```
anafilipadealmeida, line 67: From now on I'd use `hasDraftsFeature`, that already exists in the orgParser, to decide if we should should the paywall or not. If we always send the users to the plans page, there's no need for the frontend to know if it's free or pro

packages/constants/index.js
```diff
@@ -15,12 +15,14 @@ const SMALL_PLAN_ID = '10';
 // segment naming convention: APP-VIEW-LOCATION-BUTTON-VERSION
 // https://www.notion.so/buffer/CTA-Parameter-Conventions-Call-to-Action-WIP-0ae87bdd99574e3888afb2fe6b75035a
 const APP_SHELL_PRO_UPGRADE = 'publish-appShell-menu-proUpgrade-1';
+const APP_SHELL_USER_MENU_UPGRADE = 'app-shell-userMenu-upgrade';
```
albennett, line 18: I moved the app shell segment name to our constants. 

I don't like having to import from two separate packages for tracking. Maybe a future refactor could be moving the segment names to the analytics-middleware package. ✨

albennett:
Thanks for the context around billing and cta names @anafilipadealmeida . I went ahead and simplified the logic!

packages/constants/index.js
anafilipadealmeida: 😍 Thank you so much for this! I agree, at this point the constants file is getting too big just with segment names!

anafilipadealmeida (approved):
Looking great! Thank you so much for doing this @albennett 🚀

PR #1728, opened 2021-03-17
Remove missing access page

albennett:
## Checklist

#### Staging Deployment

### DIFF
--- a/packages/app-pages/components/AppPages/index.jsx
+++ b/packages/app-pages/components/AppPages/index.jsx
@@ -67,10 +67,6 @@ const AppPages = ({
   };
   const hasProfiles = profiles && profiles.length > 0;
 
-  if (!canAccessPublishing) {
-    return <MissingAccessPage />;
-  }
-
   return (
     <Switch>
       <Route path={preferencesPage.route} component={Preferences} />


karelm:
Hey @albennett , I can see that removing the component fixes the issue... But should the component be there? Are we fixing the symptom, or the cause. I'm okay fixing the symptom to get us closer to a solution, but we should also fix the cause of the issue if possible too since I would imagine we would like that page there in some scenarios. But I might be wrong. 👍

karelm:
Is the fundamental issue that this user doesn't have a default current organisation and therefore this code didn't work as expected here? If so, perhaps we should do what we did with the old org switcher and select the default org if one doesn't exist, this would then address the fundamental issue we are seeing 🤔

albennett:
Hey @karelm , you're right. We do need to fix the underlying issue. We do get the user data at this point from the app-shell but it looks like currentOrganization from the appShell isn't set, so setting the first one ourselves could fix it for these cases.  We still have the unauthorized redirect logic in place, so it should be ok to leave this logic as-is while I work on the solution.

albennett:
@karelm And thank you for your help on this! I'll take some time going over the currentOrganization logic and how it works and will make sure this problem isn't causing issues with how we're storing the selected org in our redux store.

PR #1739, opened 2021-03-24
[PUB-3203] Move missing access page to loading gate

albennett:
## Description
Render missing access pages from loading gate. 
https://buffer.atlassian.net/browse/PUB-3203

## Context & Notes

- We need to render the pages in the LoadingGate because the `ready` condition would never be successful as it's based on the publish api success. 

- I removed the `ownerEmail` logic, as this is a change we need from Core and I confirmed with Design that it's not a blocker to the missing access pages. When Core adds this change, we can include the value.

- I looked into using the `TranslationReplacer` component for the description, however, it didn't fit the needs for the mark-up in the strings. We should build out something similar to the translationReplacer if we add a translation hook. It also feels like a lot of work to use the json file with more complicated strings. It could be a good one to rethink our approach 🤔 

- Ok, last one. I timeboxed an hour on trying to get the useUser hook working in our tests. I got closer but I couldn't get the tests to pass correctly. I've added a jira task for this: https://buffer.atlassian.net/browse/PUB-3484 

## Checklist

#### Staging Deployment


packages/app-shell/package.json
```diff
@@ -11,7 +11,7 @@
     "@bufferapp/async-data-fetch": "2.0.0",
     "@bufferapp/publish-server": "2.0.0",
     "@bufferapp/publish-data-organizations": "2.0.0",
-    "@bufferapp/app-shell": "2.1.0-orgswitcher.10"
+    "@bufferapp/app-shell": "2.6.10-beta.1"
```
albennett, line 14: I'll publish in the app-shell add update this PR before deploying.

packages/missing-access-page/components/MissingAccessPage/index.jsx
```diff
@@ -73,18 +38,15 @@ const MissingAccessPage = ({
       imageSrc="https://buffer-publish.s3.amazonaws.com/images/illustration-publishing.png"
       altText={t('missing-access-page.altText')}
       header={t('missing-access-page.heading')}
-      description={description}
+      description={<span dangerouslySetInnerHTML={{ __html: description }} />}
```
anafilipadealmeida, line 41: I'm not a fan of setting HTML, as it's a security risk for cross-site scripting (XSS) attacks 🤔 I think it will not be a problem in this case, but we should think about either avoiding this, or sanitizing the html.

anafilipadealmeida (commented):
Happy to see the unauthorized-redirect go away 😂 Just left a suggestion and one comment!

packages/missing-access-page/components/MissingAccessPage/index.jsx
albennett: Hey @anafilipadealmeida my understanding was that it was only dangerous if the data wasn't static. I do see why we'd want to avoid, as someone could mirror the logic without understanding when to not use it. Let me see if there's another solution that isn't overly complicated.

albennett: Actually, we are passing a dynamic org name, so another solution is ideal.

anafilipadealmeida (approved):
Looking good to me @albennett 😍 😍  Thank you so much for all the hard work with this one!

PR #1741, opened 2021-03-25
Trigger Segment group call on org selected event

djfarrelly:
## Description
For the Pendo integration as well as other integrations like Mixpanel, we need to track users at at the group level. This allows us to group organizations in both tools.

## Context & Notes
https://paper.dropbox.com/doc/Pendo-Implementation--BHhJHecLpbyrW0m53elsTVLRAg-gZGagoSuk3nokHZx15zAB

This `group()` call should execute when:
- The app loads - it should send the current viewed org
- The org switcher changes organizations - it should send the newly selected org.

## Checklist

#### Staging Deployment


packages/analytics-middleware/middleware.js
```diff
@@ -29,6 +29,14 @@ export default ({ dispatch, getState }) => next => action => {
       break;
     }
     case orgActionTypes.ORGANIZATION_SELECTED: {
+      try {
```
djfarrelly, line 32: I'm not sure if we can guarantee the payload of the `action` to contain `selected` - I'm rusty on my Redux and didn't want a fatal error to prevent loading the app. If we can guarantee that `selected` and the correct properties are populated, we can remove this `try/catch`.

packages/analytics-middleware/middleware.js
```diff
@@ -29,6 +29,14 @@ export default ({ dispatch, getState }) => next => action => {
...
+        window.analytics && window.analytics.group(action.selected.globalOrgId, {
+          organizationId: action.selected.globalOrgId,
+          organizationName: action.selected.name,
```
djfarrelly, line 35: This is always "My Organization" during my tests

packages/analytics-middleware/middleware.js
ssvmvss: Heya @djfarrelly! 👋 

By looking at the code [here](https://github.com/bufferapp/buffer-publish/blob/main/packages/organizations/middleware.js#L36:L41), we can see that the action will always contain a `selected` property but it could potentially be `undefined` (if for any reason `list` has an `undefined` value so the default `[]` is not set) so I think keeping the `try/catch` here or any other check is a good idea.

ssvmvss (approved):

packages/analytics-middleware/middleware.js
ssvmvss: @djfarrelly Yes, "My Organization" is the default. Some users (like me!) have that value on Prod as well.

PR #1744, opened 2021-03-31
Revert "Use the core-auth svc from dev instead of production"

colinscape:
Reverts bufferapp/buffer-publish#1736

Thought this had already been changed, but apparently not. 😓 

Back to pointing at production.


karelm (approved):
LGTM 🚀

PR #1749, opened 2021-04-04
Dev pods to "normal" nodes INFRA-991

erickhun:
This will make the staging pods to be created on the "normal" nodes, where all pods are CPU limited and could be throttled.

I believe they don't need to be CPU unbounded and they're cluttering the nodes where latency-sensitive pods are and need to run fast

Screenshot 2021-04-04 at 19 46 10


colinscape (approved):
Looking good - I like keeping latency sensitive nodes for production. 👍 🌟

PR #1812, opened 2021-06-23
[PUB-3232] Update the "what would you like to post" CTA (& A/B test)

daltino:
### Description

Added a new Create Post button which would show close to the Calendar button group in the header in queue view. The “What would you like to share” is gone; replaced by this blue “Create Post” button based on a feature flip, `create_post_button_queue` assigned on an org level.

### Context & Notes
Related PR: https://github.com/bufferapp/buffer-web/pull/17748

As a Buffer customer, when I’m my queue view it should be really clear and compelling for where to click to create a cross-platform post

https://buffer.atlassian.net/browse/PUB-3232

Create Post Button Visible
Onclick of the button

### Best Practices Checklist

### ⚠️ Pre-deploy Checklist

image


packages/server/formatters/entitlements.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   CUSTOM_BITLY: 'customBitly',
   SENT_POSTS_30_DAYS: 'sentPosts30Days',
   CALENDAR: 'calendar',
+  CREATE_POST_BUTTON: 'createPostButton',
```
MayaUribe, line 6: Hey, @daltino I don't think this should be an entitlement, this is why:
- Entitlements are "features" an organization is entitled to, these are org level and **depend on the billing plan**, when we use an entitlement, this means it's somehow bound to the plan.
- A feature flip on the other hand is used to roll out a new feature, these are user level, and we use them to experiment or do a roll-out by percentage, and to be able to turn things on and off.

packages/server/parsers/orgParser.js
```diff
@@ -111,4 +111,7 @@ module.exports = orgData => ({
     !orgData.isOneBufferOrganization ||
     orgData.planBase === 'one-buffer-individual',
   shouldShowAppShellDropdownUpgradeCta: !orgData.isOneBufferOrganization,
+
+  // Experiments
+  hasCreatePostButtonFeature: orgData.ownerFeatures.includes(entitlements.CREATE_POST_BUTTON),
```
MayaUribe, line 116: @daltino in here you are checking against `orgData.ownerFeatures` which are the featureFlips for the organization owner, and this entitlement in any case would be in the organization level

MayaUribe: For percentage rollouts (10, 50%, etc) we use the feature flips, the problem is that we do this at the user level, and we are aiming to do this at the organization level, this is something we don't currently have implemented in the backend.

However this should be transparent in the frontend, I'm thinking this should look something like `orgData.features.includes(features.CREATE_POST_BUTTON)`

MayaUribe (changes requested):
Hey @daltino! Thank you for your work!
I left some comments over here! I'll be taking a look at the backend one now :D

packages/server/formatters/entitlements.js
daltino: Yh I thought about this as well, do we have a file where we store feature flips in code? @MayaUribe

MayaUribe: @daltino as I mentioned the other day in our chat, you can use the calendar feature as an example, because the features are user-based, you need to look at the `userParser` instead:
https://github.com/bufferapp/buffer-publish/blob/c62eac7d6dda59de71b0becd289cf284614d1e94/packages/server/parsers/userParser.js#L38

daltino: Ok so its the `userParser` 😄  thanks for reminding me on this @MayaUribe

packages/server/parsers/orgParser.js
daltino: I agree on this one, when we pair we can discuss in detail @MayaUribe

MayaUribe:
Hey @daltino this one is 99% there! when you get the tests fixed this one should be ready to go 😊🚀

packages/queue/middleware.js
```diff
@@ -220,6 +222,22 @@ export default ({ dispatch, getState }) => next => action => {
       break;
     }
 
+    case actionTypes.OPEN_COMPOSER_VIA_TEXTBOX: {
+      const ctaProperties = getCtaProperties(
+        SEGMENT_NAMES.SHARE_TEXTBOX
+      );
+      dispatch(analyticsActions.trackEvent('Composer Opened Via Share Textbox', ctaProperties));
```
MayaUribe, line 229: Hey @daltino I think we should confirm here because, on @jwinternheimer's PR, he mentioned: 

> Initially I said that I'd add two separate events to track clicks of each, but I thought that both could be covered with a generic Composer Opened event.

But here you are tracking it as two: `Composer Opened Via Share Textbox` and `Composer Opened Via Create Post Button`, I think the CTA properties look fine, which is what he also mentioned on his PR.

daltino: Ok so both should have same name `Composer Opened` @MayaUribe

packages/shared-components/QueueItems/Header/index.jsx
```diff
@@ -65,6 +74,17 @@ const Header = ({
           {renderCalendarButtons && (
             <CalendarButtons onCalendarClick={onCalendarClick} />
           )}
+
+          {renderCreatePostButton && (
+            <CreatePostButtonWrapper>
+              <Button
+                type={'primary'}
```
MayaUribe, line 81: A very tiny one, but the curly braces aren't necessary for this line:
```suggestion
                type="primary"
```

packages/shared-components/QueueItems/Header/index.jsx
```diff
@@ -65,6 +74,17 @@ const Header = ({
...
+                onClick={onCreatePostClick}
+                label="Create Post"
+                size="small"
```
MayaUribe, line 84: A styling detail where we could use @letsbleachthis's help 😊

The size of the button doesn't look like the one on the image in Jira:


How it is now, it looks smaller:


And if we use the use the "normal" size, it'd look like this:


I'd say the one that looks slightly bigger would be better?

MayaUribe (commented):
I left a couple of comments @daltino and one question for @letsbleachthis 😄

packages/shared-components/QueueItems/Header/index.jsx
letsbleachthis: > A styling detail where we could use @letsbleachthis's help 😊
> 
> The size of the button doesn't look like the one on the image in Jira:
> 
> 
> How it is now, it looks smaller:
> 
> 
> And if we use the use the "normal" size, it'd look like this:
> 
> 
> I'd say the one that looks slightly bigger would be better?

Ya, I'd agree @MayaUribe! One day, we hopefully will replace the old button group with the [new segmented control](https://www.figma.com/file/e98Oo1Z8lVI74Oks8zqr6f/BDS-Segmented-Control?node-id=1%3A2) component that Mw just built, and then they'll match in size, but for now — let's use the "normal" primary button size that is slightly bigger.

daltino: Ok @MayaUribe  @letsbleachthis I will use the "normal" size

PR #1850, opened 2021-07-26
Preview campaign bar component

daltino:
### Description

Add campaign bar to post previews component

### Context & Notes

It adds the campaign bar to post previewsl.

https://buffer.atlassian.net/browse/PUB-3714

Preview with Campaign Bar

### Best Practices Checklist

### ⚠️ Pre-deploy Checklist

image


packages/calendar/components/PostPreview/index.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+const PostPreview = ({ post }) => {
+  return (
+    <PostPreviewWrapper>
+      {post.campaign && (
+        <CampaignWrapper
+          campaignColor={post.campaign?.color}>
+          <ChartLine
+            color="white"
+            role="img"
```
anafilipadealmeida, line 19: I recommend this article on accessibility in svg (https://css-tricks.com/accessible-svgs/#example-2-standalone-icon-decorative). You might have gotten the inspiration for adding the role to the icon from `<PostItem>` but these 2 cases are slightly different. In the case of the PostItem, the icon is a Standalone Icon, Meaningful. That's why for accessibility we use a combination of title and role. In this case you have the name of the campaign, so the icon is just decorative and doesn't need it.

packages/calendar/components/PostPreview/style.js
```diff
@@ -0,0 +1,31 @@
...
+export const CampaignWrapper = styled.div`
+  border-radius: ${borderRadius} ${borderRadius} 0px 0px;
+  display: flex;
+  background-color: ${props => props.campaignColor};
+  padding: 8px;
+`;
+
+export const CampaignText = styled(Text)`
+  display: flex;
+  margin: -2px 0 0 8px;
```
anafilipadealmeida, line 28: I'm not sure what you're trying to achieve here but I'd avoid using negative margins if possible

packages/calendar/components/PostPreview/style.js
```diff
@@ -0,0 +1,31 @@
...
+  color: ${white};
```
anafilipadealmeida, line 29: UI Text components accept a color prop, so you don't need to add css for that: https://bufferapp.github.io/ui/#/ui/ui/text

packages/calendar/components/PostPreview/style.js
```diff
@@ -0,0 +1,31 @@
...
+
+export const CampaignWrapper = styled.div`
+  border-radius: ${borderRadius} ${borderRadius} 0px 0px;
+  display: flex;
+  background-color: ${props => props.campaignColor};
+  padding: 8px;
+`;
+
+export const CampaignText = styled(Text)`
+  display: flex;
```
anafilipadealmeida, line 27: Flexbox is used for layout, to align multiple elements. If the `<Text>` doesn't have any children, you don't need it. I'd recommend reading the css tricks guide on [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/)

anafilipadealmeida (changes requested):
Hey @daltino, I wonder if we could also add `text-overflow: ellipsis;` as we spoke in our call, so we can handle longer names nicely in the ui

packages/calendar/components/PostPreview/index.jsx
daltino: Ok got it.

packages/calendar/components/PostPreview/style.js
daltino: Ok good point!

packages/calendar/components/PostPreview/style.js
daltino: Ok noted!

packages/calendar/components/PostPreview/style.js
daltino: I wanted to align the icon and text correctly, but I can remove this.

daltino:
@anafilipadealmeida regarding the `text-overflow: ellipsis;` the `type='label'` handles from looking at the label styling, what do you think?

anafilipadealmeida:
> @anafilipadealmeida regarding the `text-overflow: ellipsis;` the `type='label'` handles from looking at the label styling, what do you think?

🤔 The thing here is that you are using an incorrect semantic element, as labels are for inputs, selects and textareas. The most appropriate element to use here is a type="p", which I believe you'll have to adjust in styles (it doesn't come with text-overflow)

daltino:
> > @anafilipadealmeida regarding the `text-overflow: ellipsis;` the `type='label'` handles from looking at the label styling, what do you think?
> 
> 
> 
> 🤔 The thing here is that you are using an incorrect semantic element, as labels are for inputs, selects and textareas. The most appropriate element to use here is a type="p", which I believe you'll have to adjust in styles (it doesn't come with text-overflow)

Ok I will make use type="p" and apply the styling.

packages/calendar/components/PostPreview/index.jsx
```diff
@@ -1,13 +1,32 @@
...
 
-const PostPreview = () => {
-  return <PostPreviewWrapper></PostPreviewWrapper>;
+const PostPreview = ({ post }) => {
+  return (
+    <PostPreviewWrapper>
+      {post.campaign && (
+        <CampaignWrapper campaignColor={post.campaign?.color?.value}>
+          <ChartLine color="white" />
+          <CampaignText type="p" color="white">
```
anafilipadealmeida, line 12: Screenshot 2021-08-02 at 10 52 40
 These are the colors campaigns can have at the moment. If the name of the campaign is always white, against background colors like yellow it will not be accessible, as it fails the color contrast test. I'd encourage you to reach out to Lauren to ask which campaign colors should have dark icon and text instead

packages/calendar/components/PostPreview/style.js
```diff
@@ -1,12 +1,27 @@
...
   display: flex;
+  flex-direction: column;
   border: 1px solid ${grayLight};
   box-shadow: ${grayShadow};
   box-sizing: border-box;
-  align-items: center;
+`;
+
+export const CampaignWrapper = styled.div`
+  border-radius: ${borderRadius} ${borderRadius} 0px 0px;
```
anafilipadealmeida, line 16: To align both the icon and text you need `align-items: center;`

anafilipadealmeida (changes requested):

packages/calendar/components/PostPreview/style.js
daltino: Ok good point!

packages/calendar/components/PostPreview/index.jsx
daltino: Ok will reach out to lauren on this. Thanks ana!

daltino: I added some logic to switch the campaign text or icon color, the only concern is how do we maintain the list of lighter colors that require dark text. @anafilipadealmeida

daltino: Another option could be if we choose to SCSS lightness($color) function to check the campaign color and infer the text or icon color.

daltino: I confirmed with Lauren, there are only two colors `pinkLighter, yellow` that need dark text.

packages/calendar/components/PostPreview/index.jsx
```diff
@@ -1,13 +1,39 @@
...
 import PropTypes from 'prop-types';
-import { PostPreviewWrapper } from './style';
+import { ChartLine } from '@bufferapp/ui/Icon';
+import { pinkLighter, yellow } from '@bufferapp/ui/style/colors';
+import { CampaignText, CampaignWrapper, PostPreviewWrapper } from './style';
 
-const PostPreview = () => {
-  return <PostPreviewWrapper></PostPreviewWrapper>;
+const PostPreview = ({ post }) => {
+  const lighterCampaignColors = [pinkLighter, yellow];
```
anafilipadealmeida, line 8: Hey @daltino, even though this works for now, it's gonna be difficult to maintain. Imagine someone adds or removes colors from campaigns, this will no longer work properly, and nobody will know we have to update a piece of code inside the calendar. I believe the best way is to create a utils function [example1](https://stackoverflow.com/questions/35969656/how-can-i-generate-the-opposite-color-according-to-current-color), [example2](https://gomakethings.com/dynamically-changing-the-text-color-based-on-background-color-contrast-with-vanilla-js/) that we can call here to generate the text color

daltino: @anafilipadealmeida yh i pointed out this question yesterday in the conversation. I will follow the example util functions thanks 😄

packages/calendar/utils/getContrastColor.js
```diff
@@ -0,0 +1,24 @@
+const getContrastColor = (hexColor) => {
+  if (hexColor === undefined)
+    return null;
+
+  if (hexColor.slice(0, 1) === '#') {
+    hexColor = hexColor.slice(1);
+  }
```
anafilipadealmeida, line 7: Hey @daltino, this is not following our prettier and eslint rules, do you have your setup 100% working?

packages/calendar/utils/getContrastColor.js
```diff
@@ -0,0 +1,24 @@
...
+    }).join('');
+  }
+
+  var r = parseInt(hexColor.substr(0, 2), 16);
+  var g = parseInt(hexColor.substr(2, 2), 16);
+  var b = parseInt(hexColor.substr(4, 2), 16);
+
+  var yiq = ((r * 299) + (g * 587) + (b * 114)) / 1000;
+
+  return (yiq >= 128) ? 'black' : 'white';
```
anafilipadealmeida, line 21: I'd use the colors from ui repo. Also, if you take a look at the figma file you'll see we don't use black for text, but a grayDarker instead.

anafilipadealmeida (changes requested):

packages/calendar/utils/getContrastColor.js
daltino: For my setup I switch Format on Save in VSCode I will double check this once more @anafilipadealmeida

packages/calendar/utils/getContrastColor.js
daltino: Got it will update.

packages/calendar/utils/getContrastColor.js
```diff
@@ -0,0 +1,28 @@
...
+      .join('');
+  }
+
+  const r = parseInt(hexColor.substr(0, 2), 16);
+  const g = parseInt(hexColor.substr(2, 2), 16);
+  const b = parseInt(hexColor.substr(4, 2), 16);
+
+  const yiq = (r * 299 + g * 587 + b * 114) / 1000;
+
+  return yiq >= 128 ? black : white;
```
anafilipadealmeida, line 25: if you take a look at the figma file you'll see we don't use black for text, but a grayDarker instead.

anafilipadealmeida (changes requested):

packages/calendar/utils/getContrastColor.js
daltino: Ah ok got it

packages/calendar/components/PostPreview/index.jsx
MayaUribe: @daltino Sorry for the ceiling-cat here, but wanted to mention that we already have a function in place for getting the color contrast, in case is helpful 👀

https://github.com/bufferapp/buffer-publish/blob/dfd7325b0aff0b4bf7988a37c8fe4452943187aa/packages/shared-components/ColorPicker/utils/HexValidations.js#L3

packages/calendar/components/PostPreview/story.jsx
```diff
@@ -14,7 +22,7 @@ const Wrapper = styled.div`
 
 storiesOf('Post Preview', module)
   .addDecorator(withA11y)
-  .add('simple', () => (
+  .add('simple w/ Campaign', () => (
```
MayaUribe, line 25: Hey @daltino what does w/ means on this line? I'm not sure it's too clear 🤔 I think is best to default to the full words to make the code super clear and non-ambiguous

packages/calendar/utils/getContrastColor.js
```diff
@@ -0,0 +1,28 @@
...
+  }
+
+  const r = parseInt(hexColor.substr(0, 2), 16);
+  const g = parseInt(hexColor.substr(2, 2), 16);
+  const b = parseInt(hexColor.substr(4, 2), 16);
+
+  const yiq = (r * 299 + g * 587 + b * 114) / 1000;
+
+  return yiq >= 128 ? grayDarker : white;
+};
```
MayaUribe, line 26: @daltino as I mentioned on the other comment, we already have this logic in place -> https://github.com/bufferapp/buffer-publish/blob/dfd7325b0aff0b4bf7988a37c8fe4452943187aa/packages/shared-components/ColorPicker/utils/HexValidations.js#L3

I would avoid duplicating this logic 🙏

MayaUribe (changes requested):
Hey @daltino I added a couple of comments, but basically, we already have some logic in place for the contract color, we use it in the Shop Grid section where users can basically pick any color they want, please make sure not to duplicate the already existing logic 🙏
Thank you!

packages/calendar/components/PostPreview/story.jsx
daltino: We use it to mean `with` we have it in other stories as well, I presume it's to keep the names short in storybook @MayaUribe

anafilipadealmeida:
> Can you point me to this existing code just to be sure @MayaUribe thanks.
Hey @daltino, @MayaUribe did it, please read the comments: 
Screenshot 2021-08-04 at 16 42 47

We have two options here, since the other utils function returns black and white, and we want grayDark and white. We either keep both utils and live with it, or tweak the old one to support both options (preferable)

packages/calendar/components/PostPreview/index.jsx
daltino: Thanks

packages/shared-components/ColorPicker/utils/HexValidations.js
```diff
@@ -7,7 +8,7 @@ export const getColorContrast = hex => {
   const b = parseInt(hexColor.substr(4, 2), 16);
   const yiq = (r * 299 + g * 587 + b * 114) / 1000;
 
-  return yiq >= 160 ? '#000000' : '#ffffff';
+  return yiq >= 160 ? grayDarker : white;
```
anafilipadealmeida, line 11: Please move this back to black and white, if you change the shared-component you'll be causing a UI regression (as there are other places that use black and not grayDark). See this in combination with the comment in: packages/calendar/components/PostPreview/index.jsx

packages/calendar/components/PostPreview/index.jsx
```diff
@@ -1,13 +1,39 @@
...
+  const textAndIconColor = getColorContrast(post.campaign?.color?.value);
+
+  return (
+    <PostPreviewWrapper>
+      {post.campaign && (
+        <CampaignWrapper campaignColor={post.campaign?.color?.value}>
+          <ChartLine color={textAndIconColor} />
+          <CampaignText
+            type="p"
+            color={textAndIconColor === grayDarker ? 'grayDarker' : 'white'}
```
anafilipadealmeida, line 18:
```suggestion
            color={textAndIconColor === black ? 'grayDarker' : 'white'}
```

anafilipadealmeida (changes requested):

packages/shared-components/ColorPicker/utils/HexValidations.js
anafilipadealmeida: Please see the comment/suggestion in: packages/calendar/components/PostPreview/index.jsx, as I believe that's the extra logic you are mentioning. There's no need for a flag in the utils function, we can just override the darker color in PostPreview

packages/calendar/components/PostPreview/index.jsx
daltino: Ok @anafilipadealmeida I discussed this with Lauren last week just so you know and she's aware of this change we decided to go with

anafilipadealmeida: @daltino, where can I find that discussion?

packages/shared-components/ColorPicker/utils/HexValidations.js
daltino: Ok your suggestion makes sense

anafilipadealmeida (approved):
I pushed a change to make the icon color match the text color. It's good to go

PR #1867, opened 2021-08-16
[PUB-3572] Task/global create button

guldenmw:
### Description

Add a global "Create Post" button on the Calendar page in the navigation header. 
The button should open the Composer and also attempt to sync channels selected in the Channel filter with pre-selected Omnibox channels. When closing the Composer, changing selected channels in the Channel filter and reopening the Composer, it mimics selecting and unselecting channels in the Composer as closely as possible.

### Context & Notes

There is some messy logic required to determine whether we're switching between a page that only views a single channel vs a page that can view multiple channels when opening the composer. This is because when opening the Composer on a page like a single channel queue, it will only open with one selected profile ie the queue you are currently viewing, however, when on a page where you can view multiple channels at once, the Composer needs to know that multiple channels need to be pre-selected. I added sufficient comments to describe the logic.

image

### Best Practices Checklist

### ⚠️ Pre-deploy Checklist

image


packages/composer-popover/components/ComposerWrapper/index.jsx
```diff
@@ -17,7 +17,11 @@ const ConnectedComposer = props => (
   </Suspense>
 );
 
-const getProfiles = (state, selectedProfileId) => {
+const getProfiles = (
+  state,
+  selectedProfileId, // Currently selected and open profile ID
```
anafilipadealmeida, line 22: I believe the descriptions you added in the interfaces file are easier to read 🙂 
```suggestion
  selectedProfileId, // Provided when in the context of a single profile (e.g. queue)
```

packages/composer-popover/components/ComposerWrapper/index.jsx
```diff
@@ -17,7 +17,11 @@ const ConnectedComposer = props => (
...
+  selectedProfileIds // All selected profile IDs
```
anafilipadealmeida, line 23:
```suggestion
  selectedProfileIds // Provided when in the context of multiple profiles (e.g. calendar)
```

anafilipadealmeida (commented):
Left some suggestions there @guldenmw . The only thing I wasn't sure about was using the queue state in the calendar. I do see the benefits of not having code duplication by using the redux actions and logic, but at the same time, it seems the type of thing in the app that is much more local state rather than global state. By reusing this we are creating an explicit queu dependency in the calendar, I'd love to hear your thoughts on it

packages/calendar/index.js
```diff
@@ -1,13 +1,18 @@
 import { connect } from 'react-redux';
 import { replace } from 'connected-react-router';
 import { calendarWeek, calendarMonth } from '@bufferapp/publish-routes';
+import { actions } from '@bufferapp/publish-queue';
```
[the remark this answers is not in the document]
guldenmw: @anafilipadealmeida Great comment! Thanks!
Honestly, the queue dependency was added purely because of my ignorance of how the composer works. 😅
I was kind of expecting some reviews based on that fact, so will happily refactor a little to remove the dependency! I definitely prefer to avoid creating one between queue and calendar if we can avoid it.

packages/composer-popover/components/ComposerWrapper/index.jsx
guldenmw: Ah great, will update it then 😄 Glad at least one of the descriptions are useful!

guldenmw:
Created a separate reducer and actions in the composer-popover package to be used instead of queue composer popover actions to avoid creating an unnecessary dependency between packages. The new reducer and actions will eventually replace the ones used in other packages as well.

packages/composer-popover/reducer.js
```diff
@@ -0,0 +1,61 @@
...
+export const actionTypes = keyWrapper('COMPOSER_POPOVER', {
+  OPEN_COMPOSER: 0,
+  OPEN_COMPOSER_VIA_BUTTON: 0,
+  OPEN_COMPOSER_VIA_TEXTBOX: 0,
+  HIDE_COMPOSER: 0,
+});
+
+export const actions = {
+  handleCreatePostClick: () => ({
+    type: actionTypes.OPEN_COMPOSER_VIA_BUTTON,
```
anafilipadealmeida, line 13: Can't believe I lost the entire comment 😅 
If you look at this: https://github.com/bufferapp/buffer-publish/blob/main/packages/queue/middleware.js#L233-L239, we are tracking the event of clicking. The name of the event is here: https://github.com/bufferapp/buffer-publish/blob/main/packages/constants/index.js#L21.  if we keep this as it is, we'll be messing the tracking results, as it's gonna track calendar buttons as being in the queue.

This might be what you are looking for:

```suggestion
    type: actionTypes.OPEN_COMPOSER,
```
I'd remove any reference to `OPEN_COMPOSER_VIA_BUTTON` and `OPEN_COMPOSER_VIA_TEXTBOX` from this file, and handle those in the further refactor

amooabeebadesina (approved):

PR #1873, opened 2021-08-24
add tracker to sidebar button

jezebelbaal:
### Description

This PR adds tracker to calendar sidebar button

### Context & Notes

[https://buffer.atlassian.net/browse/PUB-3766](https://buffer.atlassian.net/browse/PUB-3766)

### Best Practices Checklist

### ⚠️ Pre-deploy Checklist

image


packages/profile-sidebar/index.js
```diff
@@ -68,6 +70,12 @@ export default hot(
         dispatch(campaignsPage.goTo());
       },
       onCalendarButtonClick: () => {
+        const metadata = {
+          source: 'sidebar',
+          calendarVersion: 'new',
+          view: 'week',
```
anafilipadealmeida, line 76: I'm feeling a bit doubtful about this one. I understand why adding a view "week", as the calendar selected by the default is the week. At the same time, the action itself is not done with the intent of opening a week view (unlike the queue buttons). The other reason why I'd probably not pass this one is that it's hard to maintain - if we change the calendar to open the month by default, this tracking event wouldn't be automatically updated. What do you think @jezebelbaal ?

jezebelbaal: Hi Ana! From your ideas, I think that could be a tag such as default-week, default-month, for instance. So we know that the customer did not choose that option in an intentional way and we could separate them from intentional clicks. Does that make sense? What do you think, @jwinternheimer ?

anafilipadealmeida: I'd personally not overcomplicate it and just not pass it @jezebelbaal . We had asked Julian to make it not required already thinking about this scenario. We'll still be able to analyze all the data without it 😄

jezebelbaal: See your point Ana, just removed and we can assume it. Thanks for sharing : )

anafilipadealmeida (approved):

PR #1881, opened 2021-09-01
PUB-3779 | Refresh calendar page on create update pusher event

MayaUribe:
### Description
Refetch calendar data on create update pusher event.

### Context & Notes

[PUB-3779](https://buffer.atlassian.net/browse/PUB-3779): Refetch calendar data on create update pusher event.

### Screenshots
**Pusher event example**
Screen Recording 2021-09-08 at 01 34 53 PM

### Best Practices Checklist

### ⚠️ Pre-deploy Checklist
- [x] I have confirmed that the Cypress E2E test is passing (_Note: we have a known issue where it doesn't run on some PRs. Hopefully this will be fixed soon._)

image


packages/pusher-sync/middleware.js
```diff
@@ -156,6 +157,15 @@ const bindOrganizationEvents = (orgChannel, dispatch) => {
...
         post: postParser(data.update),
       });
+
+      const isCalendar = !!getMatch({
+        pathname: window.location.pathname,
+        route: calendarPage.route,
+      });
+
+      if (isCalendar) {
+        window.location.reload();
```
anafilipadealmeida, line 167: Hey @MayaUribe, I believe we were talking about refreshing the calendar queries and not the entire page (https://www.apollographql.com/docs/react/data/queries/#refetching). If we reload the page there are a number of side effects besides re-rendering things we might not need: poor UX experience (if a user has the calendar opened to compose a post and a teammate adds a post, they'd lose what they're doing), increased number of pageviews and other tracking events that come with it

MayaUribe: Thank you @anafilipadealmeida! Yes that makes a lot of sense, that's why I wanted to confirm if I understood correctly 😄

MayaUribe: Hey @anafilipadealmeida I made some changes here, but I wanted to confirm some things first since I haven't properly worked in the FE for a while :D

The way we handle the pusher events in the frontend is through redux, and in order to listen to the `POST_CREATED` action and refetch the calendar data, I need to create a reducer for the calendar, however, I know we've been moving away from redux, so I'm wondering if something like this would work, or if maybe there's a different approach I'm not seeing 😄

Thank you! 🙌

anafilipadealmeida: Great question @MayaUribe, I personally don't see a problem in us still using redux, as our app is built on that. I believe the goal was to reduce the usage of it, especially when it's not needed, which doesn't mean completely removing it. @msanroman, I'd love to know your thoughts on this particular case!

msanroman: The idea behind that [Reducing Redux thread from last year](https://threads.com/34386227995) was to indeed trigger a reflection on whether we need to use Redux for everything (as it's primarily the case for Publish and Analyze), and that many things, such as data, or component state, do not need of this complicated pattern to be managed now with hooks that allow us to colocate all this behavior within a given component (and prevent prop drilling such as the one we're doing here).

**It's fine to use Redux 🙂, and this PR looks good to me**. I'd say :shipit: !

One thing that catches my attention is that we're coupling three concepts here:

- Pusher events
- Queue events
- Whether the calendar should refetch its query

Pusher sends over an event, which gets translated into a queue event (`POST_CREATED`), and this is managed into another custom state in a custom reducer to signify whether a component deep down a tree needs to refetch a GraphQL query. In order to do any change, you need to have the context for everything.

My domain logic tells me I need to update the calendar information when a Pusher event comes in. Maybe a simpler way to manage it (I don't know if this is possible verbatim, or desirable given a broader context, but just sharing to where my mind goes) would be:

```jsx

const DaysOfMonth = ({
   currentDate,
   channelsSelected = [],
 }) => {

  useEffect(() => {
    organizationPusherChannel.bind('added_update', () => {
      refetch()
    })
    return () => {
      organizationPusherChannel.unbind('added_update')
    }
  }, [])


// ..etc
}
```

We can still centralize how we init Pusher channels, so that we can reuse things across the board, but then you make the calendar a decoupled piece that does not need to know about other things. It would be different if we had a explicit dependency on what was actually added to the queue, or we were duplicating code for managing this event elsewhere, but while it's the same event (added an update to a channel in this organization), the meaning is different: add this post in the queue (which is a push type of mutation) vs refetch whatever data you think you need to refetch (which is a pull type of mutation).

msanroman: (thanks for the ping @anafilipadealmeida, and great work, @MayaUribe ! dream team ✨ )

MayaUribe: Thank you so much for the thorough response @msanroman 🤩

I love this idea very much and is what I had in mind, however, I'm not sure how simple it'd be to add something like `organizationPusherChannel.bind` into the component, mainly because of the way we subscribe the events into pusher, which is done through the pusher middleware action:

https://github.com/bufferapp/buffer-publish/blob/a82bdee8bff2f67021d62327a996c408defe471d/packages/pusher-sync/middleware.js#L211-L217

I think it might be a bit tricky to be able to reuse the pusher channel to bind/unbind events in the component, which is why I went with this approach instead if there is a way to easily access the pusher outside the that you know of please let me know :D

packages/pusher-sync/middleware.js
```diff
@@ -208,13 +217,9 @@ export default ({ dispatch }) => {
...
-      case `getGlobalOrganizationId_${dataFetchActionTypes.FETCH_SUCCESS}`: {
-        const { globalOrganizationId } = action.result;
-        if (globalOrganizationId) {
-          const channelName = `private-updates-org-${globalOrganizationId}`;
-          const orgChannel = pusher.subscribe(channelName);
-          bindOrganizationEvents(orgChannel, dispatch);
-        }
+      case organizationsActionTypes.SET_CURRENT_ORGANIZATION: {
+        const { organizationId } = action;
+        subscribeOrganizationEvents(organizationId);
```
MayaUribe, line 222: ❓ @anafilipadealmeida @hamstu I noticed the pusher events weren't triggering consistently, and it looks like this only subscribes once, on page load, but it isn't subscribing on changing organization, so it wasn't really pushing the events.

I made this change here to subscribe to the organization events on `SET_CURRENT_ORGANIZATION` instead, but want to confirm if you think this is a good approach and that I'm not missing something 😄

anafilipadealmeida: Great catch @MayaUribe 🙌  Can't believe it took us so long to spot this, but I'm glad you did it! 😅

packages/pusher-sync/middleware.js
```diff
@@ -208,13 +217,9 @@ export default ({ dispatch }) => {
...
         setupProfilePusherEvents(action, pusher, dispatch);
         break;
-      case `getGlobalOrganizationId_${dataFetchActionTypes.FETCH_SUCCESS}`: {
-        const { globalOrganizationId } = action.result;
-        if (globalOrganizationId) {
-          const channelName = `private-updates-org-${globalOrganizationId}`;
-          const orgChannel = pusher.subscribe(channelName);
-          bindOrganizationEvents(orgChannel, dispatch);
-        }
+      case organizationsActionTypes.SET_CURRENT_ORGANIZATION: {
```
anafilipadealmeida, line 220: I'd use the `ORGANIZATION_SELECTED` instead 🙂

MayaUribe: Perfect! I'll adjust it, thank you @anafilipadealmeida 🙏

anafilipadealmeida (approved):

packages/pusher-sync/middleware.js
hamstu: Yikes - yeah also surprised we had missed this one! 🙈   Great catch @MayaUribe.

PR #1887, opened 2021-09-06
[PUB-3767]Add channel filter tracker

jezebelbaal:
### Description

This PR adds a tracker for both:

Channel Filtered: when user loads calendar and there are channels filtered
Channel Selected: when user intentionally selects channels at Channel Filter.

### Context & Notes

This is the related task: [https://buffer.atlassian.net/browse/PUB-3767](https://buffer.atlassian.net/browse/PUB-3767)

### Best Practices Checklist

### ⚠️ Pre-deploy Checklist

image


packages/calendar/components/ChannelFilter/index.jsx
```diff
@@ -38,6 +58,11 @@ const ChannelFilter = ({ channelFilter, channelSelected }) => {
     } else if (numberOfChannelsSelected === 0) {
       onSelectAllClick({ setChannels, channels });
     }
+    handleFilteredChannels(
+      channels.length,
+      numberOfChannelsSelected,
+      isViewOnly
```
anafilipadealmeida, line 64: Does the handleFilteredChannels function accept 3 arguments?

jezebelbaal: That was wrong, in fact 😅

anafilipadealmeida (commented):
@jezebelbaal , did you test this in staging with channels?
If I load the calendar, having 8 channels selected, I see 2 tracking events for channels filtered: one with both previousChannelCount and channelCount 0, and another one with previousChannelCount 8
Screenshot 2021-09-08 at 14 52 55

packages/calendar/components/ChannelFilter/index.jsx
```diff
@@ -25,6 +32,19 @@ const ChannelFilter = ({ channelFilter, channelSelected }) => {
   }`;
   const label = allChannelsAreSelected ? 'All channels' : labelForSoloChannels;
 
+  const handleFilteredChannels = (beforeFilterCount, afterFilterCount) => {
```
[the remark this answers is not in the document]
jezebelbaal: Hi, @anafilipadealmeida 

On this bug matter, I realized that when we have enter an empty channels state, the tracker would be called; we emulate a click using `onSelectAllClick` on load.

So I wonder if can we make assumptions:
I assume that when I enter an empty state channels page (reload, i.e), my previous count will be zero. Does that make sense?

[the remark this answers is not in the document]
jezebelbaal: Also, should I make the refactorings from Code Climate?

packages/calendar/components/ChannelFilter/index.jsx
```diff
@@ -45,6 +73,7 @@ const ChannelFilter = ({ channelFilter, channelSelected }) => {
     <Wrapper>
       <Select
         onSelectClick={selectedItem => {
+          handleSelectedChannels(channelsSelected.length, channels.length);
```
anafilipadealmeida, line 76: This line will also need to be adjusted to something like this:
```suggestion
          handleSelectedChannels({
            beforeFilterCount: channelsSelected.length,
            afterFilterCount: channels.length,
      });
```

packages/calendar/components/ChannelFilter/index.jsx
```diff
@@ -45,6 +73,10 @@ const ChannelFilter = ({ channelFilter, channelSelected }) => {
     <Wrapper>
       <Select
         onSelectClick={selectedItem => {
+          handleSelectedChannels({
+            beforeFilterCount: channelsSelected.length,
+            afterFilterCount: channels.length,
```
anafilipadealmeida, line 78: While tracking channels selected, the previousChannelCount is the number of channels selected before this click happened. The channelCount is how many channels are selected after this click happened. Summing up, one is the previous channelsSelected state, the other one is the current channelsSelected state.

anafilipadealmeida (changes requested):
Hey @jezebelbaal, I left you a couple of comments here 🙂

packages/calendar/components/ChannelFilter/index.jsx
```diff
@@ -25,14 +32,35 @@ const ChannelFilter = ({ channelFilter, channelSelected }) => {
...
+    trackChannelsSelected({ isWeekly, beforeFilterCount, afterFilterCount });
+  };
+
   // Select default filter values
   useEffect(() => {
     // Filter by one channel if coming from route with channel state (e.g. queue)
     if (channelSelected) {
       setChannels([channelSelected]);
+      handleFilteredChannels({
+        beforeFilterCount: channels.length,
```
anafilipadealmeida, line 54: Since when the page loads this is the first code to run, this can be replaced:
```suggestion
        beforeFilterCount: 0,
```

packages/calendar/components/ChannelFilter/index.jsx
```diff
@@ -25,14 +32,35 @@ const ChannelFilter = ({ channelFilter, channelSelected }) => {
...
+        afterFilterCount: 1,
+      });
       // Or select all channels
-    } else if (numberOfChannelsSelected === 0) {
+    } else if (numberOfChannelsSelected === 0 && !loading) {
       onSelectAllClick({ setChannels, channels });
+      handleFilteredChannels({
+        beforeFilterCount: numberOfChannelsSelected,
```
anafilipadealmeida, line 61: Since this only runs when there are no channelsSelected,
```suggestion
        beforeFilterCount: 0,
```

jezebelbaal: Makes sense, Ana! Thanks : )

packages/calendar/components/ChannelFilter/index.jsx
anafilipadealmeida: Just to make it clear @jezebelbaal , that's not what's happening here at the moment ☝️ By using the `channels.length`, you'll always get the same value

jezebelbaal: Hi, @anafilipadealmeida! Sorry it took me a while to figure this out and asked for your review before finishing it. Came up with a solution; maybe not the most elegant, though I tested across a couple more of scenarios and it seems to work. Thanks for your support and patience on this!

packages/calendar/components/ChannelFilter/generateFilterItems.jsx
```diff
@@ -10,14 +10,29 @@ export const onSelectAllClick = ({
   setChannels,
   channels,
   allChannelsAreSelected = false,
+  defaultSelected = true,
+  handleSelectedChannels,
 }) => {
+  const previousChannelCount = setChannels.length;
```
anafilipadealmeida, line 16: This setChannels is a function to set the state in React (the equivalent to useCount here: https://reactjs.org/docs/hooks-state.html). I assume what you'd want here is the length of the state array, which would be channelsSelected

packages/calendar/components/ChannelFilter/generateFilterItems.jsx
```diff
@@ -10,14 +10,29 @@ export const onSelectAllClick = ({
...
   // Unselect all channels
   if (allChannelsAreSelected) {
     setChannels([]);
+    if (!defaultSelected) {
```
anafilipadealmeida, line 20: I'm curious, what is the defaultSelected? I saw we are setting it to false for clicks on each channel and true for clicks to select all channels, but I'm wondering what's it for?

anafilipadealmeida (commented):
Hey @jezebelbaal, I'm attaching here what I'm currently seeing for the channelsSelected: tracking  https://share.getcloudapp.com/kpunR782. All 9 channels are selected. When I unselect 1, there are 8 now 👍 . However, as soon as I click on select all, instead of the tracking event to have an old state of 8 and a new of 9, it always returns 9 and 0.

My suggestion would be to remove the Channel Selected tracking event from this PR and deploy the Channel Filter, as it's ready, and pair with @MayaUribe on the other event 😊

packages/calendar/components/ChannelFilter/generateFilterItems.jsx
jezebelbaal: This is being used because selectAllClick is being used by when the page loads and when someone clicks the item. So I'm avoiding call the select tracking when it's being called by the filtering function.

MayaUribe: > My suggestion would be to remove the Channel Selected tracking event from this PR and deploy the Channel Filter, as it's ready, and pair with @MayaUribe on the other event 😊

Heya @jezebelbaal I agree with @anafilipadealmeida on splitting the work here 💯 would you be up for separating the `Calendar Channels Selected` into a different PR and we can pair on it together? 😊

packages/calendar/components/ChannelFilter/generateFilterItems.jsx
jezebelbaal: Hey Ana, that sounds better on. I faced this bug on selecting all, but weirdly I thought I could fix it. Will split them and probably it will be easier to review. Thanks for the inputs!

packages/calendar/components/ChannelFilter/generateFilterItems.jsx
jezebelbaal: Sure! 🏃‍♀️

packages/calendar/components/ChannelFilter/index.jsx
```diff
@@ -25,14 +31,31 @@ const ChannelFilter = ({ channelFilter, channelSelected }) => {
   }`;
   const label = allChannelsAreSelected ? 'All channels' : labelForSoloChannels;
 
+  const handleFilteredChannels = ({ beforeFilterCount, afterFilterCount }) => {
```
MayaUribe, line 34: A small one @jezebelbaal: I think we should rename this function so it's clear we are tracking, since `handle` can be very generic, it's very useful as we are reading the code in the future to easily detect that this piece is for tracking where we use it.

MayaUribe: We could either rename it or just use the `trackChannelsFiltered` directly :)

MayaUribe (approved):
I added a comment for readability @jezebelbaal 🙂

packages/calendar/components/ChannelFilter/index.jsx
jezebelbaal: that's right, @MayaUribe! in this case, all the props are properly available to this level, so no need to wrap this up. just called the straight tracker one instead :)

MayaUribe (approved):
Thank you @jezebelbaal 🚀

anafilipadealmeida (approved):

PR #1888, opened 2021-09-07
[PUB-3750] Task/click calendar card to edit

guldenmw:
### Description

- add the ability to edit a post by clicking on it in the calendar
- add tooltip with message "Click to edit" when hovering on post as visual indicator that it is editable.
- add a `getPost` rpc call to retrieve the post data in the correct format for the composer as the new GraphQL API doesn't parse the post data into the correct format yet
- add a callRpc helper function that uses the micro-rpc-client library to avoid using redux to store the post data from the old api

### Context & Notes

- calendar v2 feature flip will both disable the tooltip and enable the post preview once it's ready
- uses composerPopover state to tell the ComposerPopover what data to pass to the publish composer interface
- the `getPost` RPC call in PostItem can be replaced once we can either parse post data into the correct format for the composer or the composer is updated to no longer need the format returned by the legacy Publish API
- tasks with status `sent` are not editable and do not show the tooltip
- currently does not enforce view only permissions

edit-post

### Best Practices Checklist

### ⚠️ Pre-deploy Checklist

image


anafilipadealmeida (commented):
🤔 I'm not sure if it was in the requirements, but I'd not display the edit ability for past posts..

packages/calendar/components/PostItem/PostItem.jsx
```diff
@@ -0,0 +1,147 @@
...
+  Icon,
+  ErrorCard,
+  Error,
+} from './style';
+
+// Temporary wrapper component to conditionally add
+// tooltip component around PostItem based on whether
+// post has already been sent. To be deleted once PostPreview
+// component has been implemented.
+const PostItemWrapper = props => {
```
guldenmw, line 27: @anafilipadealmeida This part will exclude the edit button for sent items. The onClick also checks the post status. I figured posts in the past that haven't been sent should be editable so you can change the date?

anafilipadealmeida: Screenshot 2021-09-08 at 10 20 52

What got me a bit confusing was looking at a past post that didn't indicate any error and was still editable. Looking again this might not be a problem with the code, just the fact staging env post data doesn't seem to match prod 😅

anafilipadealmeida: I'll leave a note to myself to check prod after deploying this - to make sure if a post is past due and wasn't sent, it's marked as having an error.

anafilipadealmeida (approved):

packages/calendar/components/PostItem/PostItem.jsx
guldenmw: Also, it's worth noting that I haven't actually implemented view only permissions. 
Should I add that as part of this task?

anafilipadealmeida: Great question @guldenmw, I'd say there's no need because in edit mode the composer will only open for one single channel. However, thinking about this there's another detail we are missing with permissions: while we don't care if the user has mixed permissions in the channels filtered, to show the edit button the user needs to be a manager in that specific channel! We need to add that validation 🤔

guldenmw: I've added a commit that checks to see if the profile has isManager set to true. Mind checking it quick?

packages/calendar/components/PostItem/index.jsx
```diff
@@ -1,107 +1,25 @@
...
-PostItem.defaultProps = {
-  small: false,
-  post: {},
-};
-
-export default PostItem;
+export default connect(
+  (state, ownProps) => {
+    const { profiles } = state.profileSidebar;
+    const channelIsManager = profiles.some(
```
anafilipadealmeida, line 8: If I nit-pick, I'd probably adjust the name of this variable to either just `isManager`, as it's what we use across the repo, or to something like `isManagerInChannel`

guldenmw: I like `isManagerInChannel` 🙂 I updated it from isManager because a prop called `isManager` contextually on a post felt a bit odd.

PR #1891, opened 2021-09-13
PUB-3794 | Set even height for calendar rows in month view

MayaUribe:
### Description
Set even height for calendar rows in month view

### Context & Notes

[PUB-3794](https://buffer.atlassian.net/browse/PUB-3794): Calendar row heights vary depending on browser.

### Screenshots

**Before**

**After**
<div style="display:flex;">
  

  
</div>

### Best Practices Checklist

### ⚠️ Pre-deploy Checklist

image


anafilipadealmeida (commented):
Hey @MayaUribe , I believe there's a permanent vertical scroll in Firefox 🤔

MayaUribe:
> Hey @MayaUribe , I believe there's a permanent vertical scroll in Firefox 🤔

Yes good one @anafilipadealmeida I didn't notice it 🤔 it's been quite tricky to figure out for FF 🤔

federicoweber:
Hey @MayaUribe and @anafilipadealmeida I think I got a good fix for the scrollbar issue in Firefox.

I've also tried to enforce the header height to 32px as per specs, which is working well in Chrome but Firefox is not fully respecting the height limit for table elements, so it's growing the header just a little bit (see attached screen shots, left is Firefox right is Chrome). I think this is as good as we can get with a table structure, if we want to get more precise we should consider using a div structure with a flexbox grid.


image

packages/calendar/components/DaysOfMonth/index.jsx
```diff
@@ -19,7 +19,10 @@ import dayCells from '../Days';
 import getTimezoneFromBrowser from '../../utils/getTimezoneFromBrowser';
 
 const Row = styled.tr`
-  height: calc(100% / ${props => props.numberWeeks});
+  // We are adding 1 to the calculation to account for the header
+  // this theoretically will set the height to be smaller than needed but then
+  // the css will grow the elements proportionally to fill the space.
+  height: calc(100% / (${props => props.numberWeeks} + 1));
```
MayaUribe, line 25: 👀 Ah interesting @federicoweber!

MayaUribe:
> Hey @MayaUribe and @anafilipadealmeida I think I got a good fix for the scrollbar issue in Firefox.
> 
> I've also tried to enforce the header height to 32px as per specs, which is working well in Chrome but Firefox is not fully respecting the height limit for table elements, so it's growing the header just a little bit (see attached screen shots, left is Firefox right is Chrome). I think this is as good as we can get with a table structure, if we want to get more precise we should consider using a div structure with a flexbox grid.
> 
> image

Good one @federicoweber the reason I didn't set a fixed height for the header is that it might grow in the future when we introduce the campaign's feature (the horizontal bars in the header), but I honestly think this could work for now since styling the height for tables in Firefox has proven to be complex 😅
@anafilipadealmeida do you have any thoughts?

anafilipadealmeida (approved):
Great thoughts here @federicoweber and @MayaUribe! I'm good with shipping this as it is if you are! 😄 
Yes, when it comes to the calendar structure I'm afraid we'll always have to make trade-offs. We started with a CSS grid structure but then switched to a table because we couldn't implement advanced scroll there. We need to account for accessibility, expandable cell rows and header (to see more campaigns and more posts), overflows.. So we might need to adjust this along the way and perhaps diverge slightly from specs at times 😅

MayaUribe:
> Great thoughts here @federicoweber and @MayaUribe! I'm good with shipping this as it is if you are! 😄
> Yes, when it comes to the calendar structure I'm afraid we'll always have to make trade-offs. We started with a CSS grid structure but then switched to a table because we couldn't implement advanced scroll there. We need to account for accessibility, expandable cell rows and header (to see more campaigns and more posts), overflows.. So we might need to adjust this along the way and perhaps diverge slightly from specs at times 😅

Thank you for the context @anafilipadealmeida! That makes a lot of sense! I'm happy with this approach for now, and we'll adjust the header later when we need to include campaigns, at least it's close to the design 😅

PR #1901, opened 2021-09-15
[PUB-3806] Adds tracker to Channel Selection

jezebelbaal:
#Description

This PR adds tracker to channel selections on click.
Context & Notes

https://buffer.atlassian.net/browse/PUB-3806

### Best Practices Checklist

### ⚠️ Pre-deploy Checklist

image


packages/calendar/components/ChannelFilter/index.jsx
```diff
@@ -39,17 +42,17 @@ const ChannelFilter = ({
       trackChannelsFiltered({
         isWeekly,
         isViewOnly,
-        beforeFilterCount: 0,
-        afterFilterCount: 1,
+        previousChannelCount: 0,
+        channelCount: 1,
```
federicoweber, line 46: Why these two values are hardcoded?
Shouldn't be based on the actual count?

packages/calendar/components/ChannelFilter/index.jsx
```diff
@@ -39,17 +42,17 @@ const ChannelFilter = ({
...
       // Or select all channels
     } else if (numberOfChannelsSelected === 0 && !loading) {
       onSelectAllClick({ setChannels, channels });
       trackChannelsFiltered({
         isWeekly,
         isViewOnly,
-        beforeFilterCount: 0,
-        afterFilterCount: channels.length,
+        previousChannelCount: 0,
+        channelCount: channels.length,
```
federicoweber, line 55: Same in here 🤔.
Reading the comments I see those will depend on the navigation path, but disregarding the cause of the filter shouldn't the `channelsSelected` always be the source of truth?

Do you think we could consolidate all this into the `useTrackChannelChange` hook?

federicoweber: I was under the impression we wanted to count `channelsSelected`. What is the reason we are counting all channels in here?

packages/calendar/components/hooks/useTrackChannelChange.js
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState } from 'react';
+
+function useTrackChannelChange(
```
federicoweber, line 3: Thinking about this, probably `useTrackSelectedChannels` will be a better name 😉 .

packages/calendar/components/hooks/useTrackChannelChange.js
```diff
@@ -0,0 +1,28 @@
...
+    if (hasChannels) {
+      setPreviousState(channelsSelected.length);
+    }
+    if (previousState > -1) {
+      trackChannelsSelected({
+        previousChannelCount: previousState,
+        channelCount: channelsSelected.length,
+        isWeekly,
+      });
+    }
```
federicoweber, line 24: A small one, this logic is a bit hard to read maybe we can abstract the conditional into named variables?
Something like:
```
const isChannelsListsInitialized =  !hasChannels && channelsSelected.length > 0
const isPreviousChannelCountValid = previousState !== -1
```

packages/calendar/components/hooks/useTrackChannelChange.js
```diff
@@ -0,0 +1,28 @@
+import { useEffect, useState } from 'react';
+
+function useTrackChannelChange(
+  channelsSelected,
+  isWeekly,
+  trackChannelsSelected
+) {
+  const [previousState, setPreviousState] = useState(-1);
```
federicoweber, line 8: Let's rename `previousState` into `previousChannelsCount` it will make things a bit easier to read 😉.

packages/calendar/components/hooks/useTrackChannelChange.js
```diff
@@ -0,0 +1,28 @@
...
+    }
+    if (hasChannels) {
+      setPreviousState(channelsSelected.length);
+    }
+    if (previousState > -1) {
+      trackChannelsSelected({
+        previousChannelCount: previousState,
+        channelCount: channelsSelected.length,
+        isWeekly,
+      });
```
federicoweber, line 23: It could be worth adding a few unit tests to make sure we are calling the tracking correctly.
Let me know if you want to pair to set those up 🙂

federicoweber (changes requested):
Hey, @jezebelbaal thanks for working on this one.
I think the overall logic looks good, but I've left a few comments in the code on a couple of things I would love to address before shipping this one.

Especially I think it's important to add some unit tests to make sure we are properly tracking the events. Otherwise, it could be easy to accidentally introduce regressions that can go unnoticed for some time as the tracking has no visible effect :).

packages/calendar/components/ChannelFilter/index.jsx
jezebelbaal: Hi Fed! In this case, the setChannels right after will change the channelsSelected. In these case, we are setting up by default on the first load all available channels. This is why we don't get from channelsSelected. These are the exact scenarios we prevented from being tracked at our created hook, and will be tracked  differently.

packages/calendar/components/ChannelFilter/index.jsx
MayaUribe: Hey @federicoweber! I think the reason is that we are selecting these ones by default on the first load, and by default, it could be either all the channels or one specific channel.

packages/calendar/components/ChannelFilter/index.jsx
MayaUribe: > Do you think we could consolidate all this into the useTrackChannelChange hook?
I think it'd be great to consolidate things into the same hook, considering that we have two different ways to track this: the default channel selection (when we load the calendar page) and the manual channel selection (when the user actively filters the channels), at the moment the hook is only accounting for the later.

> I was under the impression we wanted to count channelsSelected. What is the reason we are counting all channels in here?
On this piece specifically is because we are auto-selecting all channels, so we want to track that action.

Let me know if that help @federicoweber 😊

federicoweber: Hey @MayaUribe talking with @jezebelbaal about this we decided not to refactor this part of the tracking for the moment.
The reason is that we are also using those conditionals to execute side effects (`onSelectAllClick`, and `setChannels`), which shouldn't be part of the tracking hook.

I think long term it will be beneficial to revisit this logic to properly separate the concerns, but given that we want to go live on Monday, and this is currently working fine, let's go roll with it😉.

packages/calendar/components/hooks/useTrackSelectedChannels.js
```diff
@@ -0,0 +1,32 @@
...
+    if (hasChannels) {
+      setPreviousState(channelsSelected.length);
+    }
+    if (isPreviousChannelCountValid) {
+      trackChannelsSelected({
+        previousChannelCount: previousChannelsCount,
+        channelCount: channelsSelected.length,
+        isWeekly,
+      });
+    }
```
federicoweber, line 28: So much easier to read, thanks @jezebelbaal 🙌

packages/calendar/components/ChannelFilter/index.jsx
federicoweber: @anafilipadealmeida, @mallen5311 I have a question not related to the PR. 
I would love to understand why we are tracking these two cases, given that they are default values, and not triggered by a user action.

Thanks!

federicoweber (approved):
Looks good to me @jezebelbaal thank you 🙌 :shipit:

federicoweber:
hey @jezebelbaal I see one of the tests is failing, can you please take a look at that before merging :).
Thanks!

packages/calendar/components/ChannelFilter/index.jsx
federicoweber: hey @anafilipadealmeida and @jwinternheimer, I might be wrong but following our slack [conversation](https://buffer.slack.com/archives/C151RRDL1/p1631817838405800) I think we are using these events the other way around.
If I understand this correctly:
-  `Calendar Channels Filtered` should the user action (filtering from the dropdown)
-  `Calendar Channels Selected` should the automatic action (on calendar load)

If that's the case we will need to flip the two events @jezebelbaal

packages/calendar/components/hooks/useTrackFilteredChannels.test.js
```diff
@@ -0,0 +1,59 @@
+import { renderHook, act } from '@testing-library/react-hooks';
+import useTrackFilteredChannels from './useTrackFilteredChannels';
```
federicoweber, line 2: hey @jezebelbaal I've moved this in here because it's only testing the actual hook :)

